PR #5, opened 2022-11-08
Feature/filter desirabilities

jduerholt:
This PR adds the discussed functionalibility to filter features also by their desirabilities. It is implemented in a generic fashion by an additional argument `by_attribute` in `get_features`. It can be used to apply `filter_by_class` to an attribute of a class. By this it can be used in the future also for other attributes despite desirabilities.

Tests for this new feature are already added.


bertiqwerty:
We implement alternative #7.

PR #6, opened 2022-11-08
Feature/to opti

jduerholt:
This PR adds the functionality to create a serialized opti/problem config from a domain and testing of the interconversion.

A few notes:
- One could write the mapper in a cleaner way. As it is in principle code for legacy compatibility, I did not put too much effort in it. @bertiqwerty fine for you?
- A few things are currently not possible to map from problem to domain and vice versa. From my point critical is the direction problem to domain and we have to discuss how to handle them:
    - In opti/problem an output feature has always also a lower and upper bound. In bofire/domain, we do not have these currently for an ContinuousOutputFeature. Do you need it and should we add it?
    - we do not have output_constraints in bofire/domain. What should we do with this? In our case it is somehow mixed with the desirabilities. Perhaps we need to discuss this in more detail. In this context out IdentityOutputFeatures (Max and Min) do not have the `target` attribute as you in `Maximize`/`Minimize`. Btw: we always assume maximization within Everest when setting up the desirabilities. Hope this is fine for you. 
- Currently we do not map problem.optima and problem.f, this will be going in another class on top of domain, which has to be still created.

### DIFF
--- a/bofire/mappers/opti.py
+++ b/bofire/mappers/opti.py
@@ -99,13 +99,18 @@ def feature2objective(feature: OutputFeature):
             "type": "maximize",
         }
     if isinstance(feature.desirability_function, CloseToTargetDesirabilityFunction):
-        return {
+        d = {
             "name": feature.key,
             "type": "close-to-target",
-            "exponent": feature.desirability_function.exponent,
+            # "exponent": feature.desirability_function.exponent,
             "target": feature.desirability_function.target_value,
-            "tolerance": feature.desirability_function.tolerance,
+            # "tolerance": feature.desirability_function.tolerance,
         }
+        if feature.desirability_function.exponent != 1:
+            d["exponent"] = feature.desirability_function.exponent
+        if feature.desirability_function.tolerance != 0:
+            d["tolerance"] = feature.desirability_function.tolerance
+        return d
     else:
         raise ValueError(f"Unsupported feature type {feature.__class__.__name__}.")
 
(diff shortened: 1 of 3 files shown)


jduerholt:
@bertiqwerty : I know what the problem with the failing pipeline is. In the latest mopti release the cbo benchmarks are not present. So we have two possibilities. Either pushing the latest mopi version to pypi or removing them from the tests. What do you think?

bertiqwerty:
I think we should push the final opti version.

jduerholt:
Thanks. Any comments to my questions from above?

bertiqwerty:
> Thanks. Any comments to my questions from above?

I asked some colleagues. Let's see what they say.

WaStCo:
Hi @jduerholt,
* prototype for the mapper is fine for me.
* opti/problem output lower / upper bound is not a strict necessity afaict (but see next point).
* output_constraints is required, but as in our call, this is rather to be mapped to the the desirabilities.
  * Regarding the "somehow mixed with the desirabilities" - do you think that we should define / disentangle these further?
  * Regarding target optimization - do you always define the optimization function so that the target is inplicit prior to passing it to the optimizer? This would ease adding additional methods as only maximization is required while a prior module takes care of creating minimization / target objectives in the form of a maximization task, correct?
* Thus, maximization is fine, may be parsed by the mapper aswell if not by another function.
* Agree that another mapper class on top of the domain is required (and belongs there).

bertiqwerty:
Hi @jduerholt, @WaStCo, can we merge this?

bertiqwerty:
This continues here. https://github.com/experimental-design/bofire-converters

PR #7, opened 2022-11-08
Feature/filter desirabilities alternative

jduerholt:
Hi Behrang,

I just readded the tests from my original implementation and changed a few things:
- I added to `filter_by_attribute` a first filter that checks which items in the list have the attribute for which the attribute_getter is looking for. By this one can also have objects in the data list that do not have the attribute for which the getter is looking for.
- I removed the sorting from `get_filter_by_attribute` and moved it to `Domain.get_outputs_by_desirability`. By this it is consistent with `filter_by_class` and can also applied to objects which does not allow any sorting. 

What do you think?


jduerholt:
By the way: why are you creating this `TDesirability`?

bertiqwerty:
> By the way: why are you creating this `TDesirability`?

To communicate that the expected type in `get_outputs_by_desirability` is a DesirabilityClass. Does this make sense?

bertiqwerty (approved):

PR #9, opened 2022-11-09
pytest catches warnings from unknown categories

R-M-Lee:
changed tests so that the warnings raised by CategoricalDescriptorEncoder with handle_unknown="ignore" in the presence of unknown categories are expected to be there.


jduerholt:
@jkleinekorte : could you review this PR regarding the transformer?

jkleinekorte (approved):

PR #10, opened 2022-11-10
Refactor/no desirability

jduerholt:
This PR implements the discussed changes regarding the `ContinuousOutputFeatureWoDesirabilityFunction`. It is removed and the functionality is now supported by the normal `ContinuousOutputFeature` with `desirability_function = None`

In addition I changed the `filter_by_class` and `filter_by_attribute` methods to also have the possibility to just exclude in the filter process. With this one can say:

`self.filter_outputs_by_desirability(exclude=DesirabilityFunction)` and one gets only those Output features which do not have a desirability function attached.

### DIFF
--- a/bofire/domain/util.py
+++ b/bofire/domain/util.py
@@ -45,7 +45,7 @@ def is_categorical(s: pd.Series, categories: List[str]):
 def filter_by_attribute(
     data: List,
     attribute_getter: Callable[[Type], Any],
-    includes: Union[Type, List[Type]],
+    includes: Union[Type, List[Type]] = None,
     excludes: Union[Type, List[Type]] = None,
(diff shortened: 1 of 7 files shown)


bofire/domain/util.py
```diff
@@ -57,6 +57,9 @@ def filter_by_attribute(
         except AttributeError:
             pass
 
+    print("includes", includes)
```
bertiqwerty, line 60: Debug prints?

bertiqwerty (approved):
Looks, I will remove the debug prints and merge.

bofire/domain/util.py
jduerholt: Yes, sorry!

PR #12, opened 2022-11-10
Feature/abstract class strategy

jduerholt:
Hi,

this is our proposal for the Strategy class definition. It is based on the old one from everest merged together with the one from mbo. 

Totally new is the concept of pending candidates here via having candidates in the domain definition. Maybe we have to further improve there and introduce in the candidates dataframe a column which specifies if it is still pending or not. 

It would be good to get your feedback on this defintion, so that we can finalize docstrings etc. In the next step we would then need the abstract defintions for the ModelPredictiveStrategy and the SamplingStrategy. If we have them, we can start to bring the strategies from mbo and everest back.

What do you think? 

ping @DavidWalz @bertiqwerty @R-M-Lee @jkleinekorte


DavidWalz:
Nice! Do we want to state a convention where the model (if there is one) is fitted?

bertiqwerty:
This looks very good to me. The would be fitted somewhere within `tell`, I guess.

jduerholt:
So you are fine with this? I would also create an abstract class ModelPredictiveStrategy with give a fit method and a predict method in case of a model predictive strategy? What do you think?

Based on the Strategy class we should be already able to bring back the `DOE` strategy? Would someone of you take care of this? 

In Everest we had a `RandomStrategy`, this one was a mix of botorch`s hit and run sampler, Sobol Sampler and uniform sampling. My gut feeling is to create three strategies out of this? What do you think? 

@DavidWalz you wrote also your own hit and run sampler in opti, do you want to keep this or switch to the one from botorch?

DavidWalz:
As BoTorch is already a dependency we should use from there.
I can take care of the D-optimal strategy but it will take some time as it would benefit from some major refactoring.
This brings me to the point of optional dependencies for strategies. I'll open an issue to outline my reasoning.

jduerholt:
I added now also the scaffold for the ModelPredictiveStrategy, what do you think?

ping @bertiqwerty @DavidWalz @R-M-Lee @jkleinekorte

bofire/strategies/strategy.py
```diff
@@ -0,0 +1,376 @@
...
+    @property
+    def pending_candidates(self) -> pd.DataFrame:
+        """Candidates considered as pending.
+
+        Returns:
+            pd.DataFrame: pending candidates.
+        """
+        return self.domain.candidates
+
+    def tell(
```
bertiqwerty, line 153: Why do algorithms without model such as random sampling or doe need a tell function?

bertiqwerty (approved):

bofire/strategies/strategy.py
DavidWalz: Even without model DoE-based methods depend on the location of points in input space.

jduerholt: I agree, we will also need it for closed loop experimentation for benchmark problems in python there the order is always:
- tell(experiments)
- ask() --> candidates
- f(candidates) --> experiments
- 
But I agree also with Behrang, that there are strategies where it is not needed, for this purpose I will create a further subclass in which I overwrite _tell() with pass so that this does not has to be done in every strategy again.

jduerholt:
@DavidWalz @bertiqwerty 

I added now also the abstract class `SamplingStrategy` which should be used for Strategies in which the proposed candidates does not depend on the past experiments. Fort this purpose, the method `_tell` is provided as just empty method so that the usual flow is preserved and it does not has to be overwritten in every sampling based strategy again.

Ofc. a strategy like DOE which depends on the past experiments would then not be a `SamplingStrategy` in this sense. One could also think of renaming `SamplingStrategy` to `ExperimentIndependetStrategy`. What do you think?

bertiqwerty:
> @DavidWalz @bertiqwerty
> 
> I added now also the abstract class `SamplingStrategy` which should be used for Strategies in which the proposed candidates does not depend on the past experiments. Fort this purpose, the method `_tell` is provided as just empty method so that the usual flow is preserved and it does not has to be overwritten in every sampling based strategy again.
> 
> Ofc. a strategy like DOE which depends on the past experiments would then not be a `SamplingStrategy` in this sense. One could also think of renaming `SamplingStrategy` to `ExperimentIndependetStrategy`. What do you think?

I would prefer to call everything a `SamplingStrategy` instead another name separation or a more complicated name. I would prefer even more to not have a derived `SamplingStrategy` at all but to let DoE and random sampling and what not derive from `Strategy`. And everything that can predict y.

jduerholt:
I do not fully understand what you mean. I fully agree, that everything that predicts something is a ModelPredictiveStrategy. 
DoE would be just based on Strategy,  as no predict and fit function is needed. 

Or do you mean to introduce a new kind of class for sampling based approaches, like DoE and random sampling? A consequence would be that the additional functionalities of ModelPredictiveStrategy would go into Strategy.

Perhaps, you can elaborate this a bit more ;)

bertiqwerty:
> I do not fully understand what you mean. I fully agree, that everything that predicts something is a ModelPredictiveStrategy. DoE would be just based on Strategy, as no predict and fit function is needed.
> 
> Or do you mean to introduce a new kind of class for sampling based approaches, like DoE and random sampling? A consequence would be that the additional functionalities of ModelPredictiveStrategy would go into Strategy.
> 
> Perhaps, you can elaborate this a bit more ;)

I would prefer not to have a class `SamplingStrategy` or similar. Just `Strategy` and `PredictiveStratgey` as base classes. I currently see two alternatives.
## A
```mermaid
 graph TD;
      Strategy-->PredictiveStrategy;
      Strategy-->DoE;
      Strategy-->Random;
      Strategy-->MoreSamplingMethods...;
      PredictiveStrategy-->ParEgo;
      PredictiveStrategy-->Entmoot;
      PredictiveStrategy-->MorePredictiveMethods...;
```
## B
```mermaid
 graph TD;
      Strategy-->PredictiveStrategy;
      Strategy-->SamplingStrategy;
      SamplingStrategy-->DoE;
      SamplingStrategy-->Random;
      SamplingStrategy-->MoreSamplingMethods...;
      PredictiveStrategy-->ParEgo;
      PredictiveStrategy-->Entmoot;
      PredictiveStrategy-->MorePredictiveMethods...;
```

I prefer A. I think it is sufficient and simpler. But if you all want B that would also be ok. I would not prefer something more complicated than B.

DavidWalz:
Btw, for optimal design the proposals do depend on an assumed model which needs to be defined. It's just that for models that are linear in the Ansatz-functions, the proposals do not depend on the observed y (only on x which allows to simultaneously propose all experiments). Still you would have a model specified that allows for (typically non-probabilistic) predictions.  
So DOptimalStrategy could be either `SamplingStrategy` or `PredictiveStrategy` depending on wether `predict` is implemented.

bertiqwerty:
> Are you in general fine with how the `PredictiveStrategy` looks now? Especially with having the `calc_acquisition` method in? One could also get this method in via another inheritance layer in case one has a `PredictiveStrategy` with is not able to calculate an aquisition function value.

That is a good point. I have missed that. We already have an ask method. What is the purpose of `calc_acquisition`? Maybe, the following is related to `calc_acquisition`. I am not completely sure.

Further, we also might want to include models that do not have an uncertainty build in and just use their predict. For this case I see currently 3 options.
1. We could build some kind of simple default `ask` that uses a distance-to-data based uncertainty estimation and just does optimization of the acquisition function via random sampling as fallback.
1. We do not integrate those methods here, because one might argue that they are not suitable for BO without uncertainty measurement.
1. We give them a separate base class to make them distinguishable from those that can be asked.

jduerholt:
The point of the `calc_acquisition` in `Everest` was to use in pool based benchmark studies. In a pool based benchmark the strategy has to choose based on a predefined set of candidates which experiment to conduct next. In the pool benchmark we were then calculating the acquisition value for every candidate of interest and choosing the one with the highest acquisition value as next experiment to conduct. 

For random strategies this was not implemented. We could make this whole thing more generic and provide `ask` with an additional argument `candidates` and if this argument is provided we switch to a method `_choose` which then implements how to choose from a predefined set. This would then also work for the random strategy, we would then just implement uniform sampling from the candidates set there.  In case of a BO strategy, the method would implement choosing by the highest acquistion function value.

Is this also answering your question @bertiqwerty? For me we can definitely have strategies with models that do not provide uncertainty. How the actual ask or choose then work has to be implemented in the strategy. One option would be to choose just on highest objective value. Other options would be a distance metric etc.

bertiqwerty:
> The point of the `calc_acquisition` in `Everest` was to use in pool based benchmark studies. In a pool based benchmark the strategy has to choose based on a predefined set of candidates which experiment to conduct next. In the pool benchmark we were then calculating the acquisition value for every candidate of interest and choosing the one with the highest acquisition value as next experiment to conduct.
> 
> For random strategies this was not implemented. We could make this whole thing more generic and provide `ask` with an additional argument `candidates` and if this argument is provided we switch to a method `_choose` which then implements how to choose from a predefined set. This would then also work for the random strategy, we would then just implement uniform sampling from the candidates set there. In case of a BO strategy, the method would implement choosing by the highest acquistion function value.

Ok, enabling every algorithm to do candidate selection makes sense to me. It should be possible to select more than one candidate, right? For instance, the top-5 in terms of acquisition-function values.

> Is this also answering your question @bertiqwerty? For me we can definitely have strategies with models that do not provide uncertainty. How the actual ask or choose then work has to be implemented in the strategy. One option would be to choose just on highest objective value. Other options would be a distance metric etc.

What about a default-`ask`-fallback? Doesn't need to be implemented in the base class. Could also be a helper method that contributors could call inside their implementation.

jduerholt:
> Ok, enabling every algorithm to do candidate selection makes sense to me. It should be possible to select more than one candidate, right? For instance, the top-5 in terms of acquisition-function values.

Yes, this should be possible, I will then add this to the PR this evening but also remove the `calc_acqisition` method and call it more generic.

> What about a default-ask-fallback? Doesn't need to be implemented in the base class. Could also be a helper method that contributors could call inside their implementation.

Good idea, I would then do this outside of the base class, for the choosing from predifinded candidates the most basic fallback would be just uniform sampling of the candidates, but I would not implement this as default in the base class because people get lazy and would then just use it also if their are better methods.

I will then update the PR later so that we can merge it then finally and then go on to bringing strategies back as proposed here https://github.com/experimental-design/bofire/issues/19

jduerholt:
@bertiqwerty @DavidWalz the abstract classes should now be finished. Just have a look, than we can merge it and bring back strategies ;)

PR #14, opened 2022-11-11
simplified filter by attribute

bertiqwerty:
When writing docstrings to test mkdocs, I realized that the filter by attribute function was too complicated.


jduerholt (approved):
Thanks!

PR #21, opened 2022-11-15
Refactor/containers

jduerholt:
As already mentioned is some email, I was looking in more detail into opti's parameter and constraints classes and I liked it. So I integrated it also into boFire. 

domain.constraints is now not anymore a list but a `Constraints` object which behaves like a list but implements additional features which are only relevant for constraints. The same for input and output features. All methods which needs access to more than one object stay in the `Domain` class.

@bertiqwerty @DavidWalz @R-M-Lee @jkleinekorte What do you think? Do you like this, or do you want to have it as before? If you like it, I will add more tests to it. 

Looking forward to your feedback!

Best,

Johannes

### DIFF
--- a/bofire/domain/constraints.py
+++ b/bofire/domain/constraints.py
@@ -353,8 +353,12 @@ class Constraints(BaseModel):
     def __getitem__(self, i):
         return self.constraints[i]
 
-    def evaluate_constraints(self, experiments: pd.DataFrame) -> pd.DataFrame:
+    def __call__(self, experiments: pd.DataFrame) -> pd.DataFrame:
         return pd.concat([c(experiments) for c in self.constraints], axis=1)
+
+    def add(self, constraint: Constraint):
+        assert isinstance(constraint, Constraint)
+        self.constraints.append(constraint)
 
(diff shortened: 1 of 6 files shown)


R-M-Lee:
I think this makes a lot of sense and I like it better than the previous version.

bertiqwerty:
Looks good to me. We should merge this rather sooner than later to minimize merge conflicts.

jduerholt:
> Looks good to me. We should merge this rather sooner than later to minimize merge conflicts.

True, I will add some more tests today and then we can merge this evening ;)

jduerholt:
@bertiqwerty : So I think it is done and we can merge this beast.  

A few notes:
- I added a lot of tests also for features which were migrated already before from `opti` and have not tested yet. So I found bugs that I fixed.
- Some of the tests referring in the containers are still in the `test_domain.py`, but this should not matter. We can tidy it up at some point
- There could be still things which I overlooked that could be ported from the `Domain` class into the specific container classes. But we will figure this out.

bertiqwerty (approved):

PR #24, opened 2022-11-16
Type hints and Pyright

bertiqwerty:


jduerholt:
This fixes all the problems which you had with pyright or? In your vscode, do you have the type checking defined as `strict`? I would propose to first merge https://github.com/experimental-design/bofire/pull/25 and then resolve the already existing conflicts before adding new stuff? What do you think?

bertiqwerty:
Yes. Sometimes I used `# type: ignore`. And pyright is now part of the testing pipeline. Actually, I used the default setting. Not sure if this is strict or not.

jduerholt:
Are you fixing the PR or should I do it?

bertiqwerty:
I can do it.

jduerholt:
I am also already in the process of doing it and almost done. One question:

You changed "copy_on_model_validation" to "none". Why? I need to have `False` there. Are we running on different pydantic versions? Which are you using?

jduerholt:
I am still on 1.9.0, for this I need to have `False`. Perhaps I have to update ..., and then we also have to pin it in the requirements, else the tests will fail.

bertiqwerty:
> I am also already in the process of doing it and almost done. One question:
> 
> You changed "copy_on_model_validation" to "none". Why? I need to have `False` there. Are we running on different pydantic versions? Which are you using?

Ok, then commit your version when you are.

 I got thousands of warnings with False. In the pipeline we also always have the newest version currently. I have `pydantic==1.10.2`.

jduerholt:
Yep, I fixed it now also. I will push this afternoon ;-)

jduerholt:
I was still on 1.9.0 ...

jduerholt (approved):

jduerholt:
I just merged it, it is now ready for merging it into main.

bertiqwerty:
Thanks.

PR #25, opened 2022-11-17
Feature/sampling

jduerholt:
This PR brings (back) the following features:
- `to_unit_range` and `from_unit_range` as methods of the `NumericalInputFeature`.
- `get_fixed` and `get_free` as methods of `InputFeatures` to filter quickly over fixed/non-fixed features
- uniform sampling as a method of `InputFeatures`, I really like how you did it in opti 
- sobol sampling also as method of `InputFeatures`. In opti you had it as part of the sampling module, in everest we had it as part of the `RandomStrategy`. But I think it makes most sense to have it as a method of `InputFeatures` as uniform sampling as it only affects the input features.

### DIFF
--- a/bofire/domain/features.py
+++ b/bofire/domain/features.py
@@ -1001,7 +1001,9 @@ class InputFeatures(Features):
         Returns:
             pd.DataFrame: Dataframe containing the samples.
         """
-        return pd.concat([feat.sample(n) for feat in self.get(InputFeature)], axis=1)
+        return self.validate_inputs(
+            pd.concat([feat.sample(n) for feat in self.get(InputFeature)], axis=1)
+        )
 
     def sample_sobol(self, n: int) -> pd.DataFrame:
         """Draw uniformly random samples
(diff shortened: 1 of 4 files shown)


bofire/domain/features.py
```diff
@@ -170,7 +214,8 @@ def validate_experimental(self, values: pd.Series, strict=False) -> pd.Series:
 
         Args:
             values (pd.Series): A dataFrame with experiments
-            strict (bool, optional): Boolean to distinguish if the occurence of fixed features in the dataset should be considered or not. Defaults to False.
```
bertiqwerty, line 173: I would prefer using type hints over of types in the documentation

bertiqwerty (approved):

bofire/domain/features.py
jduerholt: Hmm, I do not understand what you mean ...

bertiqwerty: There is no type hint for the variable fixed, but there is a comment saying it should be bool and that it is optional. Btw., here, optional means that it has a default parameter. But if you use `Optional[bool]` it means it can be `None` independent of the existence of a default parameter.
image

PR #29, opened 2022-11-23
Feature/sampling2

jduerholt:
This PR implements the features:
- refactoring of sampling methods in InputFeatures --> one method that can be called with an EnumArgument to choose the actual sampling method.
- In addition to UNIFORM and SOBOL, also LHS (latin hypercube sampling) is now supported.
- For constrained sampling a new base class, called `Sampler` is implemented. we can discuss if this is necesseray or if we implement it just as a Strategy. Difference is that samplers do not take care for past experiments and not at all for output features. But we can discuss this.
- I implemented then the `RejectionSampler` and the `PolytopeSampler`. So far NChooseK is not yet supported as the definition of NChooseK is wider in bofire than in opti. But we can for the start easily adopt the strategy for dealing with them as in opti. This can crash if you have on a feature bot a linear constraint and an nchoosek constraint.
- In addtion a module torch_tools.py is introduced. There I implemented a method to create torch linear (in)equality constraints out of our constraint definitions. This method was also available in mbo, but I ported the one from everest as this is also able to remove fixed features from linear constraints. Unit scaling is also supported.
- `PolytopeSampler` is implemented by using fatures from botorch. For this reason, botorch is now a new dependency.

Any comments @DavidWalz @jkleinekorte @R-M-Lee @bertiqwerty ?

### DIFF
--- a/bofire/domain/domain.py
+++ b/bofire/domain/domain.py
@@ -117,6 +117,39 @@ class Domain(BaseModel):
         return v
 
     @validator("constraints", always=True)
+    def validate_linear_constraints(cls, v, values):
+        """Validate if all features included in linear constraints are continuous ones.
+
+        Args:
+            v (List[Constraint]): List of constraints or empty if no constraints are defined
+            values (List[InputFeature]): List of input features of the domain
+
+        Raises:
+            ValueError: _description_
+
+
+        Returns:
+           List[Constraint]: List of constraints defined for the domain
+        """
+        if "input_features" not in values:
+            return v
+
+        # gather continuous input_features in dictionary
+        continuous_input_features_dict = {}
+        for f in values["input_features"]:
+            if type(f) is ContinuousInput:
+                continuous_input_features_dict[f.key] = f
+
+        # check if non continuous input features appear in linear constraints
+        for c in v:
+            if isinstance(c, LinearConstraint):
+                for f in c.features:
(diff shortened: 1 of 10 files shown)


DavidWalz:
Great work. Looks good to me.

Two minor points
- From the interface I would have expected the SamplingMethodEnum in bofire.samplers and not in bofire.domain.features. 
- I took me a moment to find out that the way to sample from samplers is via `sampler(n_samples)`. Maybe it would be one less thing to remember if it were a Strategy, thus offering an `ask` method.

jduerholt:
Concerning the enum, you are correct, but moving it to samplers would create a circular import. We could move it into helper module. What do you think? Note that you do not need to import the Enum, as I use the `validate_arguments` decorator. Thanks to pydantic you can always use also the string representation ;)

In samplers, I renamed the `__call__` method to `ask`. Concerning making it a strategy: In strategies we always enforce that we have also an output feature in the domain, but we could also relax this a bit and make it non-mandatory for the standard Strategy. Or we build the hierarchy in a way that Strategy inherits from Sampler. 

Would you be fine with merging it as is, then also the constraint handling for botorch is back and then we create a seperate Issue to discuss what exactly should be a strategy and what is a sampler and if we should have a distinct line between them.

bertiqwerty (approved):

Issue #19, opened 2022-11-14
Strategy Implementations


jduerholt:
As the abstract strategy templates are almost finished (https://github.com/experimental-design/bofire/pull/12), we should to discuss the how and who regarding bringing back of the strategies:

- Random Strategy:
    - In our implementation we had a flage `use_sobol` to switch to Sobol sampling (using the botorch implementation) if allowed by the constraints of the domain, else we were switching to the HitAndRunSampler of boTorch. NChooseK constraint and non-linear constraints were not implemented in our RandomStrategy. When calling Sobol and having categorical features we were using uniform sampling for the categorical ones and Sobol for the continuous. 

    - You implemented different Samplers (Hit and Run, Sobol, Uniform, Rejection) in `opti.sampling` and just called it in theRandomStrategy. Based on the domain, you were choosing which sampler to apply.

    - How should we do it in BoFire? Having one RandomStrategy which does the majic under the hood or having specific strategies, one for Sobol, one for Regejction? I am completely open here ...

- Entmoot: 
    - Should be a child of `PredictiveStrategy`
    - Is somebody from the BASF side takes care of adding it again?

- RandomForest:
    - Should be also a `PredicitiveStrategy`. I see that you are using there the sampling methods of opti sampling. Concerning point one (RandomStrategy), this could be an argument for doing the sampling from a conceptual point of view exaclty as you did it in opti. So having a sampling module and wrap it into the RandomStrategy. 
    - Is again somebody from BASF volunteering of adding this strategy into bofire?
    - Later one could also think about adding here also predefined objective not based randomly sampled weights. 

- DoE:
    - Would be for me also a child of `Strategy`. 
    - We already agreed that @DavidWalz is taking care of adding it.

- TSEMO:
    - Would be for me a child of `PredictiveStrategy`.
    - Can somebody of the BASF side take care of porting it?

Now the botorch complex. We have currently the following botorch based strategies:
- SOBO, AdditiveSOBO and Multiplicative SOBO. They differ just by the employed `botorch` `Objective` class. SOBO only allows for a single OutputFeature. In AdditiveSobo the objectives defined per OutputFeature are combined in additive way, whereas the scalarization in MultiplicativeSOBO is multiplicative. 
- The SOBO strategies support mixed domains, linear (in)equality and NChooseK constraints (bug making handling of the latter ones very slow has to be fixed).
- In addition, we support the following model specific things (per output model one GP is build, currently always SingleTaskGP):
[...]

PR #30, opened 2022-11-28
Refactor/objective constraints

jduerholt:
This PR resolves the issue regarding output constraint objectives discussed in https://github.com/experimental-design/bofire/issues/19 by introducing a new class called `BotorchOutputObjective`. This class also implements a `get_callables` method to use the objectives directly in constrained ooptimizations in botorch for example with `qNoisyExpectedHypervolumeImprovement` as shown here qNoisyExpectedHypervolumeImprovement. So far these constraints are not supported in everest/bofire for qehvi and qnehvi and parego.

The only thing what botorch is currently not supporting is a different `steepness` or `eta` in their nomenclature for having several constraints with different steepness values. I have PR up in botorch to solve this: https://github.com/pytorch/botorch/pull/1526

### DIFF
--- a/bofire/domain/objectives.py
+++ b/bofire/domain/objectives.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from abc import abstractmethod
-from typing import Dict
+from typing import Dict, Union
 
 import numpy as np
 import pandas as pd
@@ -22,7 +22,7 @@ class Objective(BaseModel):
     """The base class for all objectives"""
 
     @abstractmethod
-    def __call__(self, x: pd.DataFrame) -> pd.DataFrame:
+    def __call__(self, x: Union[pd.Series, np.ndarray]) -> Union[pd.Series, np.ndarray]:
         """Abstract method to define the call function for the class Objective
 
         Args:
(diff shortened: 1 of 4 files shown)


jduerholt:
@bertiqwerty: any ideas regarding this?

bofire/domain/objectives.py
```diff
@@ -18,11 +19,27 @@
 TWeight = confloat(gt=0, le=1)
 
 
+class BotorchConstrainedObjective:
+    """This abstract class offers a convenience routine for transforming sigmoid based objectives to botorch output constraints."""
+
+    @abstractmethod
+    def get_callables(self, idx: int) -> List[Callable[[Tensor], Tensor]]:
```
bertiqwerty, line 26: maybe rename to `get_constrained_objectives`

bertiqwerty (approved):

bertiqwerty:
> @bertiqwerty: any ideas regarding this?

Looks good :)

bofire/domain/objectives.py
jduerholt: What do you think about `get_constraint_callables`? I want to make clear that this method transforms the objective to a botorch output constraint that is expressed via a callable.

bertiqwerty: The fact that it returns callables is already documented by the return type.

jduerholt: This a very valid argument :D, then just `get_constraints`?

bertiqwerty: Or maybe `to_constraints` since this is more a conversion of the object than a property getter?

jduerholt: This is perfect. I will change it accordingly ;)

jduerholt: I renamed it.

PR #31, opened 2022-11-30
Feature/strategies

jkleinekorte:
transfers Everest strategies to bofire:
- botorch base strategy as parental class
- single-objective BO
- multi-objective: qEHVI, qNEHVI, qParEGO


bertiqwerty (approved):
As discussed per mail, let's merge.

PR #36, opened 2022-12-06
Refactor/refactor sobo

jkleinekorte:
replaces the if else statements to initialize various acquisition function types by get_acquisition_function from botorch


jduerholt:
In addition, I implemented a way to add strategy specific hyperparams.

bertiqwerty (approved):

PR #40, opened 2022-12-10
Remove add features

bertiqwerty:
I think, we have too many add features method that is part of our code-base and need to maintained by us. We could construct according lists and instantiate domains from those list instead of writing `domain.add_feature(...)`. In this PR I have looked at how regarding changes could be implemented for input and output features.

What do you think @jduerholt @R-M-Lee and all?

### DIFF
--- a/bofire/domain/domain.py
+++ b/bofire/domain/domain.py
@@ -827,7 +827,7 @@ class Domain(BaseModel):
                 (self.experiments, experiments), ignore_index=True
             )
 
-    def set_constraints_unvalidated(
+    def _set_constraints_unvalidated(
         self, constraints: Union[Sequence[Constraint], Constraints]
     ):
         """Hack for reduce_domain"""
--- a/bofire/domain/domain.py
+++ b/bofire/domain/domain.py
@@ -1,7 +1,7 @@
 import itertools
 import typing
 from copy import deepcopy
-from typing import Any, Dict, List, Optional, Tuple, Type, Union
+from typing import Any, Dict, List, Optional, Tuple, Type, Union, cast
 
 import numpy as np
 import pandas as pd
@@ -314,11 +314,11 @@ class Domain(BaseModel):
         """
         self.constraints.add(constraint)
 
-    def add_feature(self, feature: Feature) -> None:
-        """add a feature to list domain.features
+    def add_features(self, features: Features) -> None:
+        """add a features to list domain.features
 
         Args:
-            feature (Feature): object of class Feature, which is added to the list
+            features: to be added to the input- or output-feature0-list depending on the type
 
         Raises:
             ValueError: if the feature key is already in the domain
@@ -328,14 +328,15 @@ class Domain(BaseModel):
             raise ValueError(
                 "Feature cannot be added as experiments/candidates are already set."
             )
-        if feature.key in self.get_feature_keys():
-            raise ValueError(f"Feature with key {feature.key} already in domain.")
-        if isinstance(feature, InputFeature):
-            self.input_features.add(feature)
-        elif isinstance(feature, OutputFeature):
-            self.output_features.add(feature)
+        for feature in features:
+            if feature.key in self.get_feature_keys():
+                raise ValueError(f"Feature with key {feature.key} already in domain.")
+        if isinstance(features, InputFeatures):
+            self.input_features = cast(InputFeatures, self.input_features + features)
+        elif isinstance(features, OutputFeature):
+            self.output_features = cast(OutputFeatures, self.output_features + features)
         else:
-            raise TypeError(f"Cannot add feature of type {type(feature)}")
+            raise TypeError(f"Cannot add features of type {type(features)}")
 
     def remove_feature_by_key(self, key):
         """removes a feature from domain indicated by its key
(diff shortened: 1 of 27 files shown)


bertiqwerty:
See #41.

jduerholt:
@bertiqwerty as discussed: let us get rid of the add features methods. Do you update the PR so that we can merge it in? 

cc: @evo-bschiller

bertiqwerty:
as mentioned in #41 i will remove add features within this pr.

jduerholt:
Perfect, thanks!

evo-bschiller:
@bertiqwerty : when do you plan to finish this PR?
I would like to start working on #43 but wait for your changes so that we do not end up in merge hell :-P

bertiqwerty:
I am currently working on it.

bertiqwerty:
I committed the out comes of the compromise discussed in #41. This contains the removal of the `add_feature` and `add_constraint` methods. Further, I removed according `remove...` methods. One ugly change I had to make due to the inability of Pyright to analyze the outcome of the Pydantic validators was the addition of 
```
inputs() -> InputFeatures
outputs() -> OutputFeatures
cnstrs() -> Constraints
```
getters to `Domain`. An alternative could be to set the type of `input_features` to `InputFeatures` and have an explicit `__init__` in Domain that takes an `input_features` argument of type `Union[Sequence[InputFeature], InputFeatures]` (similar for outputs and constraints). Not sure what is the best way to combine Pydantic classes with `__init__`s. Please have a look.

bofire/domain/domain.py
```diff
@@ -42,9 +54,20 @@ class Domain(BaseModel):
         constraints (List[Constraint], optional): List of constraints. Defaults to [].
     """
 
+    def outputs(self) -> OutputFeatures:
```
jduerholt, line 57: This change is fine for me, but I would prefer to set it up as property using the @property decorator so that it behaves like an attribute. What do you think?

bofire/domain/domain.py
```diff
@@ -266,9 +292,9 @@ def get_features(
         Returns:
             List[Feature]: List of features in the domain fitting to the passed requirements.
         """
-        return (self.input_features + self.output_features).get(
-            includes, excludes, exact
-        )
+        assert isinstance(self.input_features, InputFeatures)
```
jduerholt, line 295: Why do we need this assert?

bofire/domain/domain.py
```diff
@@ -304,69 +330,9 @@ def get_feature(self, key: str):
         Returns:
             Feature: The feature with the passed key
         """
+        assert isinstance(self.input_features, InputFeatures)
```
jduerholt, line 333: Same here, why this assert? It is always of type InputFeatures or?

jduerholt:
Looks good to me. I just would prefer to change the freshly introduced inputs/outputs etc getters to properties. 

One other question: What is the difference between Sequence and List?

bertiqwerty:
> Looks good to me. I just would prefer to change the freshly introduced inputs/outputs etc getters to properties.
> 
> One other question: What is the difference between Sequence and List?

1. You mean properties with the names of the fields? That would not work because then the field names need to change. Or do you mean properties with the current names of the methods so that we would get rid of `()`? That we could do.
2. You can pass a list as sequence or e.g. a tuple. A sequence does not support `append`.

jduerholt:
> 1. You mean properties with the names of the fields? That would not work because then the field names need to change. Or do you mean properties with the current names of the methods so that we would get rid of `()`? That we could do.

I mean the second, getting rid of the `()`.

bofire/domain/domain.py
bertiqwerty: Because the type annotation of `self.input_features` is `Union[Sequence[InputFeature], InputFeatures]` and not `InputFeatures`. If it was only `InputFeatures` Pyright would complain when you, e.g., construct via `Domain(input_features=[f1, f2], ...)`. The reason is that the static type checkers cannot keep track of the conversion of the Pydantic validators for `Sequence[InputFeature]` to `InputFeatures`.

bofire/domain/domain.py
bertiqwerty: Yes, we can do that.

bofire/domain/domain.py
jduerholt: ah and the type checker gets it via the assert?

bofire/domain/domain.py
jduerholt: Do you do it?

bofire/domain/domain.py
bertiqwerty: As mentioned above, yes, it is. But Pyright cannot analyze the Pydantic conversion if someone passes a list during construction of `Domain`.

bofire/domain/domain.py
bertiqwerty: Exactly.

bofire/domain/domain.py
bertiqwerty: Yes. I shouldn't have used the confusing "we". :)

bertiqwerty:
Btw. did you notice
https://github.com/experimental-design/bofire/blob/06873189c4c5a007eb433daf2fc7e7d029d95f26/bofire/domain/domain.py#L827-L835

jduerholt:
> Btw. did you notice the new `Domain`-method

Yep, you need it for reduce, or? It is fine for me. Perhaps we should name the method `_set_constraitns_unvalidated` to make sure that it is an "expert" method.

jduerholt (approved):
Looks good to me.

PR #46, opened 2022-12-28
Feature/RandomStrategy

DavidWalz:
- adds RandomStrategy (exposes polytope or rejection sampling)
- fixes some bugs
- make Strategy._tell and _init_domain non-abstract so that they are optional for derived classes (up for debate)

### DIFF
--- a/bofire/strategies/random.py
+++ b/bofire/strategies/random.py
@@ -42,4 +42,4 @@ class RandomStrategy(Strategy):
         return candidate_pool.sample(n=candidate_count)
 
     def _ask(self, candidate_count: PositiveInt) -> pd.DataFrame:
-        return self.sampler.ask(candidate_count)
+        return self.sampler.ask(candidate_count)  # type: ignore
--- a/bofire/strategies/random.py
+++ b/bofire/strategies/random.py
@@ -1,0 +1,36 @@
+from typing import Optional
+
+import pandas as pd
+from pydantic.types import PositiveInt
+
+from bofire.samplers import RejectionSampler, Sampler
+from bofire.strategies.strategy import Strategy
+
+
+class RandomStrategy(Strategy):
+    sampler: Optional[Sampler] = None
+
+    def _init_domain(self) -> None:
+        self.sampler = RejectionSampler(domain=self.domain)
+
+    def is_constraint_implemented(self) -> bool:
+        return True
+
+    def is_feature_implemented(self) -> bool:
+        return True
+
+    def is_objective_implemented(self) -> bool:
+        return True
+
+    def has_sufficient_experiments(self) -> bool:
+        return True
+
+    def _choose_from_pool(
+        self,
+        candidate_pool: pd.DataFrame,
+        candidate_count: Optional[PositiveInt] = None,
+    ) -> pd.DataFrame:
+        return candidate_pool.sample(n=candidate_count)
+
+    def _ask(self, candidate_count: PositiveInt) -> pd.DataFrame:
+        return self.sampler.ask(candidate_count)
--- a/bofire/strategies/random.py
+++ b/bofire/strategies/random.py
@@ -1,9 +1,10 @@
 from typing import Optional
 
 import pandas as pd
+from pydantic.error_wrappers import ValidationError
 from pydantic.types import PositiveInt
 
-from bofire.samplers import RejectionSampler, Sampler
+from bofire.samplers import PolytopeSampler, RejectionSampler, Sampler
 from bofire.strategies.strategy import Strategy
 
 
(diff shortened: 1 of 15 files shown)


DavidWalz:
Tests to be added

DavidWalz:
Should be ready for review.

bofire/strategies/strategy.py
```diff
@@ -242,7 +232,7 @@ def ask(
                 ), "Number of requested candidates is larger than the pool from which they should be chosen."
             candidates = self._choose_from_pool(candidate_pool, candidate_count)
 
-        self.domain.validate_candidates(candidates=candidates)
+        self.domain.validate_candidates(candidates=candidates, only_inputs=True)
```
jduerholt, line 235: I would like to have it here always with `only_inputs = False`, as this is our general definition of candidates, that ones always also has the predicted value, uncertainty and objective value. In case of the random strategy this would be `nan`.

tests/bofire/strategies/botorch/test_base.py
```diff
@@ -727,110 +727,43 @@ def test_base_get_true_categorical_features():
...
 
     myStrategy.tell(data)
     predictions = myStrategy.predict(data)
 
-    assert len(predictions.columns.tolist()) == 2 * len(
-        domain.get_feature_keys(OutputFeature)
-    )
+    assert len(predictions.columns) == 2 * len(domain.get_feature_keys(OutputFeature))
     assert data.index[-1] == predictions.index[-1]
     domain.experiments = None
```
jduerholt, line 769: Why do you changed the tests regarding base?

jduerholt (commented):
Hi @DavidWalz,

a happy new year to you!

Thanks for the PR and sorry for the late review. Looks overall good to me, I just let some comments.

Best,

Johannes

bofire/strategies/strategy.py
DavidWalz: We have `Strategy` and `PredictiveStrategy`. I would argue that only predicitve strategies need to have predictions, whereas non-predictive strategies such as random sampling and certain classic experimental designs don't. 

Can you give me a point where we would run into problems with candidates without output predictions?

tests/bofire/strategies/botorch/test_base.py
DavidWalz: No particular reason. I was at it because I had some failing test. You are right that the MR should be focused. I'll revert.

bofire/strategies/random.py
```diff
@@ -0,0 +1,57 @@
...
+            raise Exception(errors)
+
+    @classmethod
+    def is_constraint_implemented(cls, my_type: Type[Constraint]) -> bool:
+        if my_type in [NChooseKConstraint, NonlinearEqualityConstraint]:
+            return False
+        return True
+
+    @classmethod
+    def is_feature_implemented(cls, my_type: Type[Feature]) -> bool:
```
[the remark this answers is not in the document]
DavidWalz: The tests including discrete inputs worked, but let me check what level of magic you implemented in `RejectionSampler` and `PolytopeSampler`.

bofire/strategies/strategy.py
jduerholt: You are right. Only point from my side would be that our backend expects it in this way, which is I tink not a valid point ;) 

If we do it as you propose (which is fine for me), then we have to change the `ask` for the predictive strategy in a way that this ones calls it with the additional checks, to make sure that the output of the predictive one has also the predictions.

bofire/strategies/random.py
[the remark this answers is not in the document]
jduerholt: Not sure, that it is majic :D

tests/bofire/strategies/botorch/test_base.py
jduerholt: No need for it, I was just wondering.

bofire/strategies/strategy.py
DavidWalz: We could implement `PredictiveStrategy.ask` 
```
def ask(...):
    super().ask()
    self.domain.validate_candidates(candidates=candidates)
```

bofire/strategies/random.py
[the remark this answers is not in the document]
DavidWalz: Both samplers claim to support discrete inputs (`is_feature_implemented`). For rejection sampling this is natural. For polytope sampling you implemented this a random sampling step after botorch's hit-and-run sampler which operates only on continuous inputs.

DavidWalz:
Happy new year!

One more thing, in https://github.com/experimental-design/bofire/issues/19#issue-1448109859 you considered exposing the base sampling method (uniform, Sobol, LHS). This would only affect the continuals in `RejectionSampler`, right?

bofire/strategies/random.py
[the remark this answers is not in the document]
jduerholt: Ok, this is fine for now!

bofire/strategies/strategy.py
jduerholt: Yep, this would be what I prefer. Can you add it to the PR?

jduerholt:
Yep, this comment is partially obsolete. Concerning the input features we have the sampling methods directly in `InputFeatures` and those are now exposed also to the rejection sampler as implemented here: https://github.com/experimental-design/bofire/blob/2c654108f8199de01d4d25827edbe395f3c01fc1/bofire/samplers.py#L213

jduerholt (approved):
Looks good to me.

PR #47, opened 2023-01-04
model serialization using pydantic and type unions

### LINKED ISSUE #43: model serialization (w/o to_config / from_config)
Currently, some models provide `to_config` and `from_config` methods to (de)serialize the data held by the respective child classes, e.g., Feature and InputFeature.
If I got it right, `to_config` adds a type field in addition to the rest of the data (while the rest is "basically" `.dict()`).
This type field is then used by `from_config` to determine the actual model class that we want to initialize.

AFAIK, such a type is required in the serialized data as there might be no other way to determine the correct subclass (e.g., to distinguish two different subclasses with the same fields).

Instead of creating these two methods for all models, I propose to use Literal type fields as discrimators.
While we still need a `from_spec` method (similar to but simpler than `from_config`), we can use the built-in `dict` method instead of creating `to_config`.

The following example uses this approach to model three feature types:

```python
from pydantic import BaseModel
from typing import Union, Literal
import abc

class Feature(BaseModel, abc.ABC):
    type: str
    
    @staticmethod
    def from_spec(spec: dict):
        return {
            "NumericalInput": NumericalInputFeature,
            "CategoricalInput": CategoricalInputFeature,
            "CategoricalOutput": CategoricalOutputFeature,
        }[spec["type"]](**spec)

class InputFeature(Feature, abc.ABC):
    pass

class OutputFeature(Feature, abc.ABC):
    pass

class NumericalInputFeature(InputFeature):
    type: Literal["NumericalInput"] = "NumericalInput"
    min: float
    max: float

class CategoricalInputFeature(InputFeature):
    type: Literal["CategoricalInput"] = "CategoricalInput"
    categories: list[str]

class CategoricalOutputFeature(InputFeature):
    type: Literal["CategoricalOutput"] = "CategoricalOutput"
    categories: list[str]
```

Note that any additional model XFeature must be added to the `from_spec` method (or rather its mapping).

The (de)serialization can then be done as follows:

```python
f1 = NumericalInputFeature(min=1, max=2)
f2 = CategoricalInputFeature(categories=["i", "n"])
f3 = CategoricalOutputFeature(categories=["o", "u", "t"])

print("objects:", f1, f2, f3)

print("serialized:", f1.dict(), f2.dict(), f3.dict())

print(
    "deserialized:",
    Feature.from_spec(f1.dict()),
    Feature.from_spec(f2.dict()),
    Feature.from_spec(f3.dict()),
)
```

```
objects: type='NumericalInput' min=1.0 max=2.0 type='CategoricalInput' categories=['i', 'n'] type='CategoricalOutput' categories=['o', 'u', 't']
serialized: {'type': 'NumericalInput', 'min': 1.0, 'max': 2.0} {'type': 'CategoricalInput', 'categories': ['i', 'n']} {'type': 'CategoricalOutput', 'categories': ['o', 'u', 't']}
deserialized: type='NumericalInput' min=1.0 max=2.0 type='CategoricalInput' categories=['i', 'n'] type='CategoricalOutput' categories=['o', 'u', 't']
```

Unfortunately, models with a Feature field are unable to resolve this correctly, e.g.:

```python
class BadModel(BaseModel):
    feature: Feature

bad = BadModel(feature=f1.dict())
print(bad)
print(bad.feature.__class__.__name__)
```

```
feature=Feature(type='NumericalInput')
Feature
```

Instead, we can specify AnyFeature as a Union of all desired subclasses and use it in models that references them.
Here, the type field is used as discriminator to determine the correct model, e.g.:

```python
AnyFeature = Union[
    NumericalInputFeature,
    CategoricalInputFeature,
    CategoricalOutputFeature,
]

class GoodModel(BaseModel):
    feature: AnyFeature

good = GoodModel(feature=f2.dict())
print(good)
print(good.feature.__class__.__name__)
```

```
feature=CategoricalInputFeature(type='CategoricalInput', categories=['i', 'n'])
CategoricalInputFeature
```

In this approach, a new subclass must be added to AnyFeature as well.
This also works for composite types, e.g.:

```python
class Features(BaseModel):
    values: list[AnyFeature]

features = Features(values=[
    f1.dict(),
    f2.dict(),
    f3.dict(),
])
print(features)
print([v.__class__.__name__ for v in features.values])
```

```
values=[NumericalInputFeature(type='NumericalInput', min=1.0, max=2.0), CategoricalInputFeature(type='CategoricalInput', categories=['i', 'n']), CategoricalOutputFeature(type='CategoricalOutput', categories=['o', 'u', 't'])]
['NumericalInputFeature', 'CategoricalInputFeature', 'CategoricalOutputFeature']
```

This approach harnesses the general power of pydantic to (de)serialize data and removes the necessity for `to_config` and `from_config`.

The similar `from_spec`method is required solely to deserialize an instance directly but not used by referencing models as they rely on AnyFeature and the discriminating type field.

What do you think, does it make sense to change this (de)serialization?

evo-bschiller:
Here, we remove the existing (de)serialization via the `to_config` and `from_config` methods and add serialization via the use of `Union` types and `type` properties in the pydantic models.
Now, serialization can be done by the `.dict()` of each pydantic model object.
Deserialization is then achieved by passing this dict to the appropriate constructor, e.g., `domain_ = Domain(**domain.dict())` as long as all references to other model types use the `Union` type `AnyX`, e.g., `AnyFeature`.
This does not work to deserialize inheriting objects, i.e., `Feature(**feature_of_type_categorical_input.dict())` cannot create an instance of feature of type `CategoricalInput` because `Feature` does not know about it.
Instead, you can use `pydantic.parse_obj_as` to create an instance of the correct type, e.g., `pydantic.parse_obj_as(AnyFeature, f.dict())`. Note that the correct model class is picked from the `Union` type `AnyFeature` using the descriminating `type` property of each `Feature`.
For simplicity, we also added a static function to such base classes, e.g., `Feature.from_dict(f.dict())`, which calls `pydantic.parse_obj_as`.
closes #43

### DIFF
--- a/bofire/domain/constraints.py
+++ b/bofire/domain/constraints.py
@@ -1,7 +1,7 @@
 import collections.abc
 from abc import abstractmethod
 from itertools import chain
-from typing import Dict, List, Sequence, Tuple, Type, TypeVar, Union
+from typing import Dict, List, Sequence, Tuple, Type, TypeVar, Union, Literal
 
 import numpy as np
 import pandas as pd
@@ -14,6 +14,8 @@ from bofire.domain.util import BaseModel, filter_by_class
 
 class Constraint(BaseModel):
     """Abstract base class to define constraints on the optimization space."""
+    
+    type: str
 
     @abstractmethod
     def is_fulfilled(self, experiments: pd.DataFrame) -> pd.Series:
@@ -82,6 +84,8 @@ class LinearConstraint(Constraint):
         coefficients (list): list of coefficients (float) of the constraint.
         rhs (float): Right-hand side of the constraint
     """
+    
+    type: Literal["LinearConstraint"] = "LinearConstraint"
 
     features: TFeatureKeys
     coefficients: TCoefficients
@@ -140,6 +144,8 @@ class LinearEqualityConstraint(LinearConstraint):
         coefficients (list): list of coefficients (float) of the constraint.
         rhs (float): Right-hand side of the constraint
     """
+    
+    type: Literal["LinearEqualityConstraint"] = "LinearEqualityConstraint"
 
     # def is_fulfilled(self, experiments: pd.DataFrame, complete: bool) -> bool:
     #     """Check if the linear equality constraint is fulfilled for all the rows of the provided dataframe.
@@ -179,6 +185,8 @@ class LinearInequalityConstraint(LinearConstraint):
         coefficients (list): list of coefficients (float) of the constraint.
         rhs (float): Right-hand side of the constraint
     """
+    
+    type: Literal["LinearInequalityConstraint"] = "LinearInequalityConstraint"
 
     # def is_fulfilled(self, df_data: pd.DataFrame) -> bool:
     #     """Check if the linear inequality constraint is fulfilled in each row of the provided dataframe.
@@ -257,6 +265,9 @@ class LinearInequalityConstraint(LinearConstraint):
 
 
 class NonlinearConstraint(Constraint):
+    # TODO: add docstring to NonLinearConstraint
+    
+    type: Literal["NonlinearConstraint"] = "NonlinearConstraint"
     expression: str
 
     def __call__(self, experiments: pd.DataFrame) -> pd.Series:
@@ -264,6 +275,9 @@ class NonlinearConstraint(Constraint):
 
 
 class NonlinearEqualityConstraint(NonlinearConstraint):
+    # TODO: add docstring to NonlinearEqualityConstraint
+    type: Literal["NonlinearEqualityConstraint"] = "NonlinearEqualityConstraint"
+    
     def is_fulfilled(self, experiments: pd.DataFrame) -> pd.Series:
         return pd.Series(np.isclose(self(experiments), 0), index=experiments.index)
 
@@ -272,6 +286,9 @@ class NonlinearEqualityConstraint(NonlinearConstraint):
 
 
 class NonlinearInqualityConstraint(NonlinearConstraint):
+    # TODO: add docstring to NonlinearInqualityConstraint
+    type: Literal["NonlinearInqualityConstraint"] = "NonlinearInqualityConstraint"
+    
     def is_fulfilled(self, experiments: pd.DataFrame) -> pd.Series:
         return self(experiments) <= 0
 
@@ -290,6 +307,7 @@ class NChooseKConstraint(Constraint):
             this flag decides if zero active features are also allowed.
     """
 
+    type: Literal["NChooseKConstraint"] = "NChooseKConstraint"
     features: TFeatureKeys
     min_count: int
     max_count: int
@@ -460,3 +478,6 @@ class Constraints(BaseModel):
                 exact=exact,
             )
         )
+
+# TODO: check list of all constraints, possibly remove "intermediate" classes
+AnyConstraint = Union[LinearConstraint, LinearEqualityConstraint, LinearInequalityConstraint, NonlinearConstraint, NonlinearEqualityConstraint, NonlinearInqualityConstraint, NChooseKConstraint, Constraints]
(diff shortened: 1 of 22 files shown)


bofire/domain/constraints.py
```diff
@@ -213,9 +195,14 @@ def as_greater_equal(self) -> Tuple[List[str], List[float], float]:
...
 
+    # TODO: from_greater_equal should take the object as input
     @classmethod
     def from_greater_equal(
-        cls, features: List[str], coefficients: List[float], rhs: float
+        cls,
+        features: List[str],
+        coefficients: List[float],
+        rhs: float,
+        **kwargs,
```
jduerholt, line 205: why do we need the kwargs here?

bofire/domain/constraints.py
```diff
@@ -230,9 +217,14 @@ def from_greater_equal(
...
 
+    # TODO: from_smaller_equal should take the object as input
     @classmethod
     def from_smaller_equal(
-        cls, features: List[str], coefficients: List[float], rhs: float
+        cls,
+        features: List[str],
+        coefficients: List[float],
+        rhs: float,
+        **kwargs,
```
jduerholt, line 227: Same question here regarding the kwargs ;)

bofire/domain/constraints.py
evo-bschiller: When calling this method, we pass **constraint.dict() as arguments.
In addition to the fields desired here, the newly introduced `type` property is also passed along.
Of course, this leads to errors.

This was the easier solution than removing `type` from all inputs.
IMO this was a quick fix, because I would rather go for:
- pass the actual models as arguments instead of their properties, or
- change the calls to pass `x=model.x` instead of `**`

bofire/domain/constraints.py
evo-bschiller: see above :-P

tests/bofire/domain/test_domain.py
```diff
@@ -393,8 +393,11 @@ def test_categorical_combinations_of_domain_filtered(domain, data, include, excl
 
 @pytest.mark.parametrize("domain", [domain, domain2])
 def test_domain_serialie(domain):
```
jduerholt, line 395: typo here, should be serialize

jduerholt:
For me this looks in principle good. @bertiqwerty: can you also have a look?

bertiqwerty:
> Deserialization is then achieved by passing this dict to the appropriate constructor, e.g., `domain_ = Domain(**domain.dict())` as long as all references to other model types use the `Union` type `AnyX`, e.g., `AnyFeature`. This does not work to deserialize inheriting objects, i.e., `Feature(**feature_of_type_categorical_input.dict())` cannot create an instance of feature of type `CategoricalInput` because `Feature` does not know about it. Instead, you can use `pydantic.parse_obj_as` to create an instance of the correct type, e.g., `pydantic.parse_obj_as(AnyFeature, f.dict())`. Note that the correct model class is picked from the `Union` type `AnyFeature` using the descriminating `type` property of each `Feature`. For simplicity, we also added a static function to such base classes, e.g., `Feature.from_dict(f.dict())`, which calls `pydantic.parse_obj_as`.

So in the latter case we have a static method `from_config` and in the case of `Domain` we do not? I propose to always add a static `from_config`, i.e., also for `Domain`, due to consistency.

bertiqwerty:
Looks good! I left some minor comments.

bofire/domain/constraints.py
evo-bschiller: i will change it to the second option

tests/bofire/domain/test_domain.py
evo-bschiller: will do

evo-bschiller:
Hi @jduerholt @bertiqwerty,
I adressed all your comments, committed the corresponding changes, and merged with main

jduerholt (approved):
THX!

PR #55, opened 2023-01-12
deletes study from tests

jkleinekorte:
removes study.py, multiobjective.py and singleobjective.py as well as all dependencies in the tests on this


bofire/benchmarks/benchmark.py
```diff
@@ -38,8 +38,11 @@ def best_multiplicative(domain: Domain):
 
 
 class Benchmark:
-    def f(self, candidates: pd.DataFrame) -> pd.DataFrame:
+    def f(self, candidates: pd.DataFrame, return_complete=False) -> pd.DataFrame:
```
jduerholt, line 41:
```suggestion
    def f(self, candidates: pd.DataFrame, return_complete: bool=False) -> pd.DataFrame:
```

jduerholt: Could you also include the flag in the tests of the benchmark functions?

jduerholt (changes requested):
Only two minor things, else it looks great.

jduerholt (approved):
Looks good to me. Thanks!

PR #64, opened 2023-01-20
Hotfix/linear kernel

jduerholt:
This PR adds:
- Bugfix regarding the active dimensions in the linear kernel. This keyword was previously ignored which could lead to unintended behavior with the Mixed GP.
- In addition, priors are added, so far they are only usable for the lenghtscale of RBF/Matern. This has to be extended and a bit refactored in the hackathon.

### DIFF
--- a/bofire/models/priors.py
+++ b/bofire/models/priors.py
@@ -1,0 +1,71 @@
+from abc import abstractmethod
+from typing import Literal, Union
+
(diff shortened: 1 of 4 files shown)


jduerholt:
@bertiqwerty: can you have a look at this PR?

bertiqwerty (approved):

PR #65, opened 2023-01-23
Aspen_benchmark

evo-jdridder:
- Aspen_benchmark class added, that creates a connection to a Aspen Plus simulation as a benchmark problem.
- Autosafe functionality integrated into benchmark.single_run() that saves benchmark function results including candidates and hypervolume in a .json file every 10 iterations.


jduerholt:
Branches mixed up, closed and new one is created.

PR #113, opened 2023-02-14
68 port doe from basfmbo

dlinzner-bcs:
- ported doe to bofire
- drop categorical support for first release


.github/workflows/lint.yaml
```diff
@@ -58,5 +58,6 @@ jobs:
       - name: pyright
         run: |
           pip install .[testing]
+          conda install -c conda-forge cyipopt
```
DavidWalz, line 61: do we need this for linting?

.github/workflows/test.yaml
```diff
@@ -18,7 +18,9 @@ jobs:
       - name: Install Bofire
         run: pip install .[testing]
       - name: Run tests
-        run: pytest -ra --cov=bofire --cov-report term-missing tests
+        run: |
+        conda install -c conda-forge cyipopt
```
DavidWalz, line 22: I think so far we only use linux-based runners. There you can pip install cyipopt and install ipopt via the corresponding package manager (apt).

requirements.txt
```diff
@@ -7,4 +7,5 @@ pytest
 torch
 botorch>=0.8.1
 multiprocess
-plotly
\ No newline at end of file
+plotly
+formulaic==0.3.4
```
DavidWalz, line 11: add cyipopt

DavidWalz (commented):
On dependencies:
I would state cyipopt as pip dependency and require the user to make sure that ipopt is installed. If the user is working on windows they can conda install cyipopt on top get the precompiled ipopt lib.

Also the setup.py needs to be extended to include formulaic & cyipopt as optional dependencies for [doe].

tests/bofire/strategies/doe/test_utils.py
```diff
@@ -0,0 +1,664 @@
...
+        input_features=[
+            ContinuousInput(key=f"x{i+1}", lower_bound=0.95, upper_bound=1.0)
+            for i in range(4)
+        ],
+        output_features=[ContinuousOutput(key="y")],
+    )
+    assert np.allclose(g_efficiency(X, domain), 0.333, atol=5e-3)
+
+    # define domain: sampling not implemented
+    domain = Domain(
```
dlinzner-bcs, line 445: @jduerholt this would be an example where I do not know hat to do.

original question to Johannes: 
I am currently reimplementing tests from DoE to bofire. I noticed that in bofire NChooseKConstraints require to have inputs with lower_bounds==0. Additionally, LinearInequalityConstraints require two have at least two features – so there is no workaround. This blocks me from having NChooseKConstraints in all cases where inputs are not lower bounded by 0 - which seems restrictive to me...

dlinzner-bcs: I understand you solution would be to add 
LinearEqualityConstraint(
                features=["x1", "x2", "x3", "x4"], coefficients=[-1, -1, -1, -1], rhs=-0.95
            ),
which would enforce the desired behavior as max_count=1 implies that only one variable can be active.

jduerholt: Yep, it is implicit. But what does it mean for an NChooseK constraint if one of its substituents has a lower_bound>0? This is for me a bit ill defined, or not?

jduerholt:
@dlinzner-bcs : this could be interesting also for this PR: https://github.com/experimental-design/bofire/pull/114. The Polytope sampler can now also handle NChooseK constraints.

tests/bofire/strategies/doe/test_utils.py
dlinzner-bcs: Yes, I also discussed with @R-M-Lee  now and I see now what I want to do makes no sense. Still, I am pretty sure I have seen such ill defined problems in the old DoE package (e.g. I want chemical A or B, but if I have A i want to have at list X amount of it..). I think this makes sense if you add binary variables to the picture a_is_active, b_is_active and put NChooseK on those. But I don't think this needs to be solved by release. For now I suggest we drop support for such problems until we know how to solve then properly.

jduerholt: Yep, good idea. I think this could be also a nice discussion for our next Hackathon ;)

jduerholt:
@dlinzner-bcs : Is this now ready for review for me, or should I wait a bit more?

bofire/strategies/doe/design.py
```diff
@@ -0,0 +1,255 @@
...
+from bofire.strategies.doe.jacobian import JacobianForLogdet
+from bofire.strategies.doe.utils import (
+    constraints_as_scipy_constraints,
+    get_formula_from_string,
+    metrics,
+    nchoosek_constraints_as_bounds,
+)
+
+
+def _domain_for_sampling(domain: Domain) -> Domain:
```
jduerholt, line 22: I think we can remove this now, as the NChooseK sampler is now able to sample also from domains with NChooseK constraints.

bofire/strategies/doe/design.py
```diff
@@ -0,0 +1,255 @@
...
+def find_local_max_ipopt(
+    domain: Domain,
+    model_type: Union[str, Formula],
+    n_experiments: Optional[int] = None,
+    tol: float = 0,
+    delta: float = 1e-7,
+    ipopt_options: Dict = {},
+    jacobian_building_block: Optional[Callable] = None,
+    sampling: Optional[np.ndarray] = None,
+    fixed_experiments: Optional[np.ndarray] = None,
```
jduerholt, line 94: What do you think about modelling this as pandas dataframe, then we can also validate that it is formatted correctly by using `domain.validate_candidates(fixed_experiments, only_inputs=True)`. Furthermore, it fits more into the bofire way of using dataframes on all user exposed interfaces.

bofire/strategies/doe/design.py
```diff
@@ -0,0 +1,255 @@
...
+
+def find_local_max_ipopt(
+    domain: Domain,
+    model_type: Union[str, Formula],
+    n_experiments: Optional[int] = None,
+    tol: float = 0,
+    delta: float = 1e-7,
+    ipopt_options: Dict = {},
+    jacobian_building_block: Optional[Callable] = None,
+    sampling: Optional[np.ndarray] = None,
```
jduerholt, line 93: I would also change this to a pandas dataframe as described below.

bofire/strategies/doe/design.py
```diff
@@ -0,0 +1,255 @@
...
+        return obj
+
+    return objective
+
+
+def find_local_max_ipopt(
+    domain: Domain,
+    model_type: Union[str, Formula],
+    n_experiments: Optional[int] = None,
+    tol: float = 0,
```
jduerholt, line 89:
```suggestion
    tol: float = 0.0,
```

bofire/strategies/doe/design.py
```diff
@@ -0,0 +1,255 @@
...
+    elif n_experiments < n_experiments_min:
+        warnings.warn(
+            f"The minimum number of experiments is {n_experiments_min}, but the current setting is n_experiments={n_experiments}."
+        )
+
+    # initital values
+    if sampling is not None:
+        x0 = sampling
+    else:
+        domain_for_sampling = _domain_for_sampling(domain=domain)
```
jduerholt, line 158: As described above: I think this is not anymore necessary if you merge the current main in, and use the Polytopesampler as is.

tutorials/models.ipynb
```diff
@@ -1686,7 +1686,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3.8.12 ('everest_env')",
+   "display_name": "bofire",
    "language": "python",
    "name": "python3"
   },
```
jduerholt, line 1692: I think the changes on this file can be removed from the PR, or?

.github/workflows/test.yaml
```diff
@@ -19,6 +19,10 @@ jobs:
         run: pip install .[testing]
       - name: Run tests
         run: pytest -ra --cov=bofire --cov-report term-missing tests
+      - name: Install optional dependencies
```
jduerholt, line 22: Why not install it directly together with the whole framework above and then invoke pytest in a way that it runs also the test involving cyipopt?

bofire/strategies/doe/design.py
```diff
@@ -0,0 +1,255 @@
...
+
+    # check if all points respect the domain and the constraint
+    domain.validate_candidates(
+        candidates=design.apply(lambda x: np.round(x, 8)), only_inputs=True
+    )
+
+    return design
+
+
+def check_fixed_experiments(
```
jduerholt, line 235: Let's also add a test for this method ;)

jduerholt:
I just had a look why the tests are failing.

Instead of `from bofire.domain import Domain` it is now `from bofire.domain.domain import Domain`.

This is the true path, the old one was masked, so internally we should use them, for users we will introduce an `api` import path. But this will not affect this PR.

tests/bofire/strategies/doe/test_design.py
```diff
@@ -288,13 +285,13 @@ def test_find_local_max_ipopt_fixed_experiments():
         ],
     )
 
-    with pytest.warns(ValueError):
```
dlinzner-bcs: @jduerholt here the sampler complains that there is nothing to sample. I think the sampler is correct - it can just sample [[1,0,0],[0,1,0],..]. How should this be handled? Should the sampler refuse to provide a sample if only a single allowed one exist?

jduerholt: Fixed it in main.

jduerholt (approved):
So for me it is fine, and I think we should merge it now. Next step would then be to wrap a strategy around it. 

Also fine for @bertiqwerty  and @DavidWalz  to merge now as is?

PR #116, opened 2023-02-16
Make Pyright happy

bertiqwerty:
… pydantic

This PR should fix the Pyright problems in the pipeline. Basically, we replace wlog
```
TSomething = conlist(item_type=str, min_items=2)
```
with
```
TSomething = Annotated[List[str], Field(min_items=2)]
```
This way, Pyright does not complain about the type while still type checking against `List[str]` and Pydantic still validates.

See `typing.Annotated` under
https://docs.pydantic.dev/usage/types/#standard-library-types


jduerholt:
@evo-bschiller : Do you have a look at the pydantic changes

@bertiqwerty : You also changed the pipelines or and united the jobs for flake8, black and isort again, or?

bertiqwerty:
> @evo-bschiller : Do you have a look at the pydantic changes
> 
> @bertiqwerty : You also changed the pipelines or and united the jobs for flake8, black and isort again, or?

Yes. I thought 3 setups instead of one are in this case an additional waste of compute resources.

evo-bschiller:
Looks great, I love it when you find ways to actually make something work in multiple frameworks at the same time :-D

If I remember correctly, the current goal for bofire is to support python version 3.7 and above.
The import `from typing import Annotated` is only available starting at 3.8.
For backwards compatibility, we must use the following import:
```
from typing_extensions import Annotated
```
To allow this, we must add `typing-extensions` as dependency:
https://pypi.org/project/typing-extensions/

An example of this in the pydantic doc:
https://docs.pydantic.dev/usage/schema/#typingannotated-fields

evo-bschiller:
FYI: EOL for 3.7 is in 4 months (https://endoflife.date/python).
IMHO this is a good argument to remove support for 3.7 and start with 3.8 instead

evo-bschiller:
just realized that we are using 3.8+, forget everything I said :-D

evo-bschiller (approved):

PR #117, opened 2023-02-16
Feature/serial

evo-bschiller:
Here, we extend and improve the serialization and deserialization of pydantic models.
This includes domain, models, and strategies.

Helpers for (de)serialization can be found in bofire.serial.

To make this possible, we cleaned up the AnyX definitions (now found in bofire.any) and split some existing modules, e.g., bofire.domain.features.
Furthermore, we added capabilities to (de)serialize pandas DataFrame and Series.

Note that the necessary restructuring of some modules can lead to broken imports in code using the previous version.


jduerholt:
@bertiqwerty : can you have a look at this PR from Benni which is adding the serialization to bofire and tidies up the code a bit. 

For me it is fine, and we should merge it is in as soon as possible as this will also cause some restructuring in other PRs.

jduerholt (approved):

PR #119, opened 2023-02-16
Feature/random forest

jduerholt:
This PR implements the sklearn `RandomForestRegressor` into bofire and shows how to use it in a multiobjective optimization. 

It is not yet tested, for this reason this PR is WIP. 

Might be interesting for @rosonaeldred, @R-M-Lee  and @DavidWalz.


R-M-Lee:
How does this relate to what we have been doing here:
https://github.com/experimental-design/bofire/tree/69-starting-with-random-forest-a-general-uncertaintyoptimization-wrapper-for-any-model
There we use RandomForestRegressor from sklearn and random sampling to generate a candidate pool.

jduerholt:
Yes, I would recommend to use this wrapped RF in the new strategy, what I show here is how to make RF compatible with the existing strategies, but we are here dependent on the most recent botorch branch. I will ask there, when they are planning a release.

If it is too far in the future, we should increase the recent version into out pipeline.

jduerholt:
In addition, this PR also bumps the botorch version to 0.8.2 as this is necessary to use the RF model.

jkleinekorte (approved):

PR #129, opened 2023-02-27
Update setup.py

jduerholt:
This PR bumps the python version to 3.9 as the `annotated` type was introduced in 3.9.


bertiqwerty:
Good catch!

bertiqwerty (approved):

PR #134, opened 2023-03-07
removed import statements for black, isort, flake since they might ha…

bertiqwerty:
…ve different versions compared to pre-commit or pipeline


WaStCo (approved):
Only documentation updated, thank you.

PR #135, opened 2023-03-09
add optional "unit" field to NumericalInput and ContinuousOutput

jkeupp:
We've always had an optional "unit" field in our domain definition, at least for numerical inputs and outputs. Would it be okay for you to add it ?


jduerholt:
@evo-bschiller @bertiqwerty 

I am not sure here, as this is information which is not relevant for optimizer at all, I am unsure what it should help. @jkeupp : do you want it for your backend? We could then think about adding it to the new data models but not to the functional models.

jduerholt:
@jkeupp : another point, we are currently refactoring a bit, so we should defintiely wait until the end of the refactoring.

jkeupp:
Yes, we used to store a "campaign", which has a domain, and inside it, there used to be optional units for each in- and output feature. If we're lacking this, there'd be several options which all seem ugly to me. We need the unit info! ... 

- we'd need additional data structure in the campaign, with repeated feature keys, where we can then store the unit
- we add the units info into the feature key itself, perhaps using some special delimiter, 
- well, how are you doing it. I'm absolutely fine with going for an alternative solution, but so far, I don't see it.

PS: sry for not having checked in with the tests, I'll not forget that later on :P

jduerholt:
Hi @jkeupp, we just discussed it. I just discussed it with @evo-bschiller, it is fine for us.

I would recommend to add a field `unit: Optional[str] = None` to the abstract `NumericalInputFeature`. Note that we merged yesterday a major refactoring in, so you have to update to this new commit.

Best,

Johannes

PR #136, opened 2023-03-13
refactor everything

evo-bschiller:
In this single commit, we refactor the complete repo (bofire and tests) to achieve the following:

- all data models (formerly partly in domain) are now moved to data_models, i.e., everything that can be (de)serialized
- models are renamed to surrogates
- surrogates and strategies are split in data model and function part (in surrogates and strategies)


jduerholt (commented):
Looks good to me.

jduerholt (approved):

PR #137, opened 2023-03-14
Doe add support for nonlinear constraints

Osburg:
This PR adds
- support for nonlinear constraints for doe
- optional names and jacobian attributes to constraint classes
- an optional tolerance parameter for constraint violation (to facilitate checks with `is_satisfied()`)
The changes where proposed and discussed in [this](https://github.com/basf/mopti/issues/39) and [this](https://github.com/basf/doe/issues/14) issue opened in mopti, doe respectively.
- more precise jacobian computation method as suggested in [this PR](https://github.com/basf/doe/pull/16) in basf/doe


jduerholt:
Hi @Osburg, 

thanks for your contribution. 

We merged this morning a major refactoring in. Basic essence of this is that we created the new subpackage `data_models` in which the actual pydantic data structures are stored. 

Strategies and models (now surrogates) are splitted into a data_model which is validated and then the functional part which is then execcuting the actual tasks. For your PR this means that you have to adjust to the move of `domain.py` etc. into the new structure. 

Could you do this change? I will also add a few comments to the PR.

Best,

Johannes

bofire/domain/constraint.py
```diff
@@ -38,6 +40,18 @@ def __call__(self, experiments: pd.DataFrame) -> pd.Series:
         """
         pass
 
+    @abstractmethod
```
jduerholt, line 43: Do we want to enforce jacobian implementation for all constraints? This means for example also a Jacobian for NChooseK which is not really possible as it is in principal non-differentiable in its general form. Of course there are continuous relaxations for it, I would just implement it for those constraints where it is possible and not add it as abstract method for the base class.

bofire/domain/constraint.py
```diff
@@ -231,17 +264,40 @@ class NonlinearConstraint(Constraint):
 
     type: Literal["NonlinearConstraint"] = "NonlinearConstraint"
     expression: str
+    jacobian_expression: str = None
```
jduerholt, line 267:
```suggestion
    jacobian_expression: Optional[str] = None
```

bofire/domain/constraint.py
```diff
@@ -231,17 +264,40 @@ class NonlinearConstraint(Constraint):
...
+    features: FeatureKeys = None
```
jduerholt, line 268: @bertiqwerty, we could add the features then also as required attribute for all constraints in the `Constraint` base class. I would prefer when it is always available.

jduerholt (commented):
I already made some comments, in genereal we should also provide tests for the jacobian implementations for example.

bofire/domain/constraint.py
Osburg: I agree, we do not need to have a jacobian implementation for every type of constraint. Removed the decorator.

tests/bofire/data_models/test_constraints.py
```diff
@@ -120,7 +124,8 @@ def test_constraints_get(constraints, ConstraintType, exact, expected):
 
 
 def test_constraints_plus():
-    returned = constraints + constraints2
+    returned = constraints + constraints4 + constraints2
+    print(returned.constraints)
```
jduerholt, line 128:
```suggestion

```

jduerholt (changes requested):
Looks good for me, I just had some minor comments inline. One other thing: can you move the notebook to `tutorials/doe`? I already placed there one last week. And can you resolve the merge conflict? Thanks for your contribution!

bofire/strategies/doe/design.py
```diff
@@ -104,44 +104,56 @@ def find_local_max_ipopt(
...
 
-    # initital values
     if sampling is not None:
         domain.validate_candidates(sampling, only_inputs=True)
         x0 = sampling.values
     else:
-        sampler = PolytopeSampler(data_model=PolytopeSamplerDataModel(domain=domain))
-        x0 = sampler.ask(n_experiments, return_all=False).to_numpy().flatten()
+        try:
+            sampler = PolytopeSampler(domain=domain)
```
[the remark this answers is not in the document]
Osburg: Yes, nonlinear constraints are not supported for the polytope sampler (at least according to its method `is_constraint_implemented()`). try except is now replaced by checking if there are nonlinear constraints.

jduerholt:
Pyright complains because of features as optional field for the non linear constraints. What do you think to make it there also required? So that features has to provided always for every constraint?

jduerholt:
Pyright complains because of features as optional field for the non linear constraints. What do you think to make it there also required? So that features has to provided always for every constraint? An alternative would be to set it to type: ignore

Osburg:
I think there are cases where having to add the feature keys would be annoying, for example if you are only interested in evaluating the constraint violation but no gradient. So I think it would be good to leave it as an optional attribute, what do you say?
And I think the problem is not that features is optional for `NonlinearConstraint` objects but that in general a `Constraint` object does not have a features attribute, right?
So what I did is change the type of the `constraint` attribute of `ConstraintWrapper` to `NonlinearConstraint`, since the constraint wrapper is in fact only needed for nonlinear constraints. Then I also do not get complaints from pyright anymore. 

P.S.: Some stuff here was new for me, for example I never used pyright before. This is why I did not notice the pyright errors before. Sry for that! Future contributions will be less messy, promised! :D

bofire/strategies/doe/design.py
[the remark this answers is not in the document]
jduerholt: Great, thanks. Yes, the polytopesamler can only handle linear and NchooseK constraints.

jduerholt (approved):
Looks good to me! Thank you very much for this contribution!

PR #141, opened 2023-03-17
MakeDocsHappy

evo-jdridder:
Update for mkdocs so the package's documentation can be generated again after refactoring.


jduerholt (approved):
lgtm!

PR #148, opened 2023-03-22
Refactor/bounds

evo-jdridder:
The ContinuousInputfeature object and its subclasses now have their bounds saved into a Tuple[float, float] object instead of two separate float objects.


evo-jdridder:
For some reason github visually includes all commits that I have made on other branches eventhough I checked out a fresh branch from `upstream main` before starting working. `Changed files` only show changes made in the working branch `refactor/bounds`.

PR #171, opened 2023-04-24
allow n-choose-k when variable bounds don't contain 0

### LINKED ISSUE #118: Discuss domains with NChooseKContraint and ContinuousInputs with non-zero bounds
Yep, good idea. I think this could be also a nice discussion for our next Hackathon ;)

_Originally posted by @jduerholt in https://github.com/experimental-design/bofire/pull/113#discussion_r1108096449_

DavidWalz:
As discussed we want to allow n-choose-k over variables whose bounds don't contain 0.

Example:

* x1 in [0.1, 1]
* x2 in [0.2, 1]
* x3 in [0.3, 1]
* n-choose-k([x1, x,2, x3], min_count=1, max_count=2)

means that 1-2 out 3 variables should be active, and when they are active their bounds apply.

DOE does not require this. BoTorch apparently does, so a check needs to be placed in the right place.

Closes #118

### DIFF
--- a/bofire/data_models/domain/domain.py
+++ b/bofire/data_models/domain/domain.py
@@ -198,35 +198,6 @@ class Domain(BaseModel):
                     ), f"{f} must be continuous."
         return v
 
-    @validator("constraints", always=True)
-    def validate_lower_bounds_in_nchoosek_constraints(cls, v, values):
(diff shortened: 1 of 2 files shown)


jduerholt (approved):
LGTM

PR #185, opened 2023-04-25
Refactor/remove matplotlib from domain

jduerholt:
- remove dependency to matplotlib
- move plotting functionality based on plotly to central point
- update botorch version to latest release
@bertiqwerty, @DavidWalz : why do we need sklearn in the minimal installation?


bertiqwerty (approved):

bertiqwerty:
> @bertiqwerty, @DavidWalz : why do we need sklearn in the minimal installation?

I don't see the reason. David?

DavidWalz:
> > @bertiqwerty, @DavidWalz : why do we need sklearn in the minimal installation?
> 
> I don't see the reason. David?

Without sklearn you can't import `feartures` as this GAUCHE feature thingy needs CountVectorizer from sklearn. Not nice.

Issue #186, opened 2023-04-26
Move scikit-learn dependency


bertiqwerty:
See https://github.com/experimental-design/bofire/pull/185#issuecomment-1522106260

PR #187, opened 2023-04-26
moved sklearn dependency out of install_requires

bertiqwerty:
#186

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -22,7 +22,7 @@ jobs:
           pytest tests/bofire/data_models/test_domain.py \
             tests/bofire/data_models/test_base.py \
             tests/bofire/data_models/test_constraint_fulfillment.py \
-            tests/bofire/data_models/test_constrints.py \
+            tests/bofire/data_models/test_constraints.py \
             tests/bofire/data_models/test_domain_validators.py \
             tests/bofire/data_models/test_features.py \
             tests/bofire/data_models/test_filters.py \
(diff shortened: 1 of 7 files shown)


bertiqwerty:
Does anyone know, why the test-pipeline is not triggered?

bertiqwerty:
> Does anyone know, why the test-pipeline is not triggered?

Never mind, typo in the yaml.

jduerholt:
I think our tests are not yet modular enough to run them independet for the different installation modi ...

jduerholt:
And perhaps we should have the different tests configs in different jobs in the pipeline so that one sees directly in the PR where it crashes.

bertiqwerty:
I have safeguarded the hell out of the tests. Can you have another look, @jduerholt .

jduerholt:
Looks good to me, only one thing which came to my mind:

In the minimal test plan we should also run the following other test files:

- data_models/test_base.py
- data_models/test_constraint_fulfillment.py
- data_models/test_constrints.py
- data_models/domain_validators.py
- data_models/test_features.py
- data_models/test_filters.py
- data_models/test_nchoosek_combinatorics.py
- data_models/test_numeric.py
- data_models/test_unions.py
- data_models/test_util.py

bertiqwerty:
Ok, @jduerholt, I have added more tests.

jduerholt:
Thanks, thats looks perfect. One last thing which came to my mind: we should also add the `cloudpickle` dependency to `optimization`. We use it to store/save empirical surrogates which cannot be pickled. Currently, the tests are not executed as these tests are safeguarded. I will approve now, than you can merge it in without the need to wait for me.

jduerholt (approved):

PR #189, opened 2023-05-01
Fix DoE optimality criteria tutorial (new objective enum)

DavidWalz:


DavidWalz:
Misread the git message. direct push to main was possible.

PR #194, opened 2023-05-08
Molecular Inputs and models

simonsung06:
Pull request for molecular inputs and models that are usable with `optimize_acqf_mixed`. Lets discuss the ideas implemented here. Suggested changes and ideas are very welcome!

The following are new additions and bug fixes/changes:

**data_models sub-folder**:
- `MolecularInput` class is implemented and is able to generate and use any type of molecular feature
- The `MolecularInput` class inherits from `CategoricalInput` and shares some similarities to `CategoricalDescriptorInput`; the key difference is that `self.values` is generated
- Molecular features are defined in molfeatures as their own classes. This structure enables separate arguments for molecular features to be specified using `MolecularEncodingEnum` when `MolecularInput` is instantiated. Molecular features implemented so far are fingerprints, fragments, fingerprints_fragments, and mordred descriptor features. They are generated and stored as `self.values` during the fit function
- Tanimoto kernel from the Guache package has been copied into the `kernels/fingerprint_kernels` folder.
- `HammondDistanceKernel` is presumably incorrectly spelled and so is renamed to `HammingDistanceKernel`. `HammingDistanceKernel` is just using the `CategoricalKernel `from BoTorch
- Additions to `data_models/domain/features.py` for the new `MolecularInput `and molecular feature types (`molfeatures`)
- `enum.py` updated to include new `MolecularEncodingEnum` for different types of molecular features
- Suggested change for data_models/features/categorical.py in the `from_onehot_encoding` and `from_dummy_encoding` functions that allow for multiple `_` characters in user-specified feature names. The original code split the descriptor name string at every `_` character including in the user-specified feature names and then always selects the second split string. The fix will enable it to always correctly select the last split string.
- New data_model surrogates for `TanimotoGP` and `MixedTanimotoGP` created. Details of how they function are below

**surrogates sub-folder**:
- `TanimotoGP` is designed to be used exclusively with any number of `MolecularInput` with fingerprints, fragments, and fingerprints_fragments features. `MixedTanimotoGP` is designed to be used with these particular molecular input feature types but in combination with at least one other continuous, categorical, or `MolecularInput` with mordred descriptors features. `MolecularInput` with fingerprints, fragments, and fingerprints_fragments features are not scalarised and therefore are treated differently to `MolecularInput` with mordred descriptors features. As a result, `MolecularInput` with mordred descriptors features can be used with SingleTaskGP whereas the rest cannot.
- `TanimotoGP` will exclusively use the `TanimotoKernel` kernel
- `MixedTanimotoGP` will use any specified continuous, categorical, or molecular kernel, defaults to `Matern`, `HammingDistanceKernel`, `TanimotoKernel`. `active_dims` is used to select the features for each kernel type during the fit function. Kernels for continuous (inc. `MolecularInput` with mordred descriptors features) and categorical are only used if the respective input type is used. The kernels are combined in a similar manner to `MixedSingleTaskGP` where the final `covar_module = sum of kernels + product of kernels.` This is just the preliminary `covar_module` method and is open to manual tweaking and tuning for model performance.
- `TanimotoGP` and `MixedTanimotoGP` are usable with `optimize_acqf_mixed` to generate candidates. Only tried `EXHAUSTIVE` method so far.
- `get_scalar` function in `single_task_gp.py` has been slightly changed to account for `MolecularEncodingEnum.MOL_DESCRIPTOR`
- Additions to `mapper.py` for new surrogates

**strategies sub-folder**:
- `get_categorical_combinations` modified to account for `MolecularEncodingEnum`

**utils sub-folder**:
- Additions to `cheminformatics.py` for mordred descriptors

**Other notes**:
- Bag of characters is currently not enabled because the number of generated features can vary, i.e. test data containing unseen molecules can have a different number of features to the training set. Therefore, predictions cannot be made on the test data. To my knowledge, I do not know a way around this issue at the moment so I am open to ideas/corrections.
- I had some issues using one-hot encoding for `CategoricalInput` when with various other Input types due to the `CategoricalInput` not being the right-most feature in the array. I left a commented out crude workaround in the get function of `data_models/domain/features.py`. Any advice on how to properly do this without this crude solution is welcome.
- Please ignore all docstrings for new class/functions. They are not accurate
- Tests not written yet due to potential changes based on this discussion
- Constraints not tested yet with `MolecularInput`


jduerholt:
Hi Simon,

thanks for this great PR. A few comments from my side (based on them we should then discuss in detail).

Let's start with the `MolecularInput` feature:

- You inherited now directly from the `CategoricalInput`, which has a few consequences:
    - We are not able to setup a prediction model for molecules without defining categories in advance
    - We have to provide always both categories and smiles which I see as double work. From my understanding, smiles are just categories in this context.
- I would propose a different inheritance structure in this context: Namely introducing, a `MolecularCategoricalInput` which inherits from both `MolecularInput` and `CategoricalInput`. This would give us the possibility to reuse a lot of the functionalities and are more fine grained. Together with this, I would then also tidy up a bit the whole inheritance structure of the features.
- I am also not sure that we really need the descriptor functionalities here. We have to discuss this.

Regarding kernels. You have now put the actual implementations also into the `data_models` subdirectory, instead we should create a kernels directory in the main structure and create a mapping for this. For implementing custom kernels, the current data structure is not well suited. I have to think about it and come up with a better idea, more in the direction of a `map` method as for `strategies` and `surrogates`. This would also allow for installing the types API without the botorch/gpytorch/pytorch dependencies. 

Regarding the GPs: 

- Why do you create a new type TanimotoGP? Could we not just use the existing SingleTaskGP and provide it with the TanimotoKernel? In addition we introduce there an identity scaler which applies no scaling to the inputs. 
- Somehow the samme could also work with the original MixedGP.
- In our strategies, we should include botorchs `optimize_acqf_discrete` method to optimize over purely categorical search spaces.

Misc:

- Thanks for catching the bug with the `_` in the descriptors. Can you create a PR just for this, so that we get it directly in without the dependency on this large feature PR?
- I know the bug with the categoricals sometimes not being the outmost indices. I have to think there about a proper solution.

What do you think about splitting this PR up in specific parts? I would see the following things:

- Implementation of the MolecularInput feature
- Implementation of the surrogates and kernels
- MISC stuff

jduerholt:
This has been implemented in seperate PRs, so it is obsolete.

PR #196, opened 2023-05-09
Feature/forbid extra

jduerholt:
This PR introduces the feature that one cannot anymore initialize the data models with arguments that are not part of the data model. For this purpose the `Extra.forbid` option of pydantic is used.


bofire/data_models/base.py
```diff
@@ -12,6 +13,8 @@ class Config:
         validate_assignment = True
         arbitrary_types_allowed = False
         copy_on_model_validation = "none"
+        extra = Extra.forbid
```
evo-bschiller, line 16: please remove the lines above that contain the TODO to actually do this :-D

evo-bschiller (changes requested):
you should remove the todo of stuff you have done :-P

bofire/data_models/base.py
jduerholt: Thanks for the hint. Done!

evo-bschiller (approved):

PR #197, opened 2023-05-10
replaced flake8 and isort with ruff

bertiqwerty:
I replaced flake with ruff and ignored some issues ruff had additionally. However, many of those issues are very valid. Hence, I propose to first introduce the tool and then step-by-step remove items from the ignore list in the `pyproject.toml` . I have also fixed a few minor issues identified by ruff.


DavidWalz:
Looks good! 

Rule exceptions have a tendendency to stay for good, so I went ahead with enforcing rules from E and B, which again uncovered some potential bugs and typos.

From the remaining exceptions we probably don't want to enforce E501, B017 and B027 (up for debate).

I don't understand B023.

B028, B904, B905 are nice to have but IMO not that important.

DavidWalz (approved):
From my side good to go. Thanks!

jduerholt:
Thanks! Great work. Should we just create an issue to discuss which rules to enforce and which to ignore? Of course, after merging this in.

bertiqwerty:
Thanks, David.

bertiqwerty:
> I don't understand B023.

https://docs.python-guide.org/writing/gotchas/#late-binding-closures

PR #201, opened 2023-05-11
Add option not to raise validation error in strategy.ask()

evo-nlueck:
Sometimes `strategy.ask()` raises errors during the validation of the generated candidates in `self.domain.validate_candidates(...)` with the explanation that some constraint is violated. In some cases it might still be useful to return the candidates, though. Therefore I added a boolean argument `raise_validation_error` to `strategy.ask()`. If it is `True` an error will still be raised (default behaviour). If it is `False` only a warning will be displayed and the candidates will be returned regardless of the constraint violations.

To achieve this, I created a new class of `ConstraintNotFulfilledError` to differentiate it from other errors that can be raised in `validate_candidates`. Other `ConstraintErrors` might be worth a thought as well.  I could think of a `ConstraintInfeasibleError` for cases where multiple constraints contradict each other (e.g. $x > 1$ and $x < 0$ where no value for $x$ is possible).

### DIFF
--- a/bofire/data_models/constraints/api.py
+++ b/bofire/data_models/constraints/api.py
@@ -1,6 +1,6 @@
 from typing import Union
 
-from bofire.data_models.constraints.constraint import Constraint
+from bofire.data_models.constraints.constraint import Constraint, ConstraintNotFulfilledError
 from bofire.data_models.constraints.linear import (
     LinearConstraint,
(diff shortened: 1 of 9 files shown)


jduerholt (approved):
Looks good to me. Thank you very much!

PR #202, opened 2023-05-12
Feature/trim pipeline

jduerholt:
In our pipelines, we are currently always installing torch including nvidia drivers and GPU support, which is not needed. For this reason, I changed it to install only the cpu version.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -32,6 +32,7 @@ jobs:
           python-version: 3.9
       - name: pyright
         run: |
-          pip install ".[optimization,cheminfo,tests]"
+          pip install torch --index-url https://download.pytorch.org/whl/cpu
+          pip install ".[optimization,tests,cheminfo]"
           pyright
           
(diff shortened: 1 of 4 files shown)


bertiqwerty:
I stumbled across the same problem. Why did you close this?

WaStCo:
In a discussion with @bertiqwerty , stumbled upon the same thing.

I think the test was dropped because we used:

```console
pip install torch --index-url https://download.pytorch.org/whl/cpu
```

But I think we only need to add an extra index to pip.

```console
pip install torch --no-cache-dir --extra-index-url https://download.pytorch.org/whl/cpu
```

Where I am unsure about, torch will also be found in the original index, since we do not further specify the version to a point where it is unique (name:version unique to the CPU only index). Thus, this might still fail.

jduerholt:
> I stumbled across the same problem. Why did you close this?

I do not know anymore. I think the speedup in the tests was kind of neglectable ...

bertiqwerty:
The main advantage would not be the pipeline but that users do not have to install as much in the default setting.

Issue #199, opened 2023-05-11
Ruff rules to enforce


bertiqwerty:
As follow up of #197, we discuss here which of the open ruff rules to enforce.
|rule|David|Behrang|
|-|-|-|
|E501|don't enforce->black|don't care -> black|
|B017|don't enforce|enforce*|
|B023|enforce|enforce|
|B027|don't enforce|don't enforce|
|B028|not important but nice|not important but nice|
|B904|not important but nice|not important but nice|
|B905|not important but nice|don't enforce|

*especially in tests one could check for the expected exception

- [E501 - line too long](https://beta.ruff.rs/docs/rules/line-too-long/)
- [B017 - x raises exception](https://beta.ruff.rs/docs/rules/assert-raises-exception/)
- [B023 - late binding closures](https://docs.python-guide.org/writing/gotchas/#late-binding-closures)
- [B027 - empty method wo @abstract](https://beta.ruff.rs/docs/rules/#flake8-bugbear-b)
- [B028 - enforce stacklevel parameter](https://beta.ruff.rs/docs/rules/no-explicit-stacklevel/)
- [B904 - raise form inside except](https://beta.ruff.rs/docs/rules/#flake8-bugbear-b)
- [B905 - zip without strict](https://beta.ruff.rs/docs/rules/#flake8-bugbear-b)

@jduerholt, here is the discussion issue you proposed. I haved added links to the rules, @DavidWalz's opinion from the PR, and mine to get the discussion started.

bertiqwerty:
> I find B017 relevant, but it requires us (to my understanding) to change all base Exceptions that we raise in bofire to a suitable derived Exception and adjust the tests accordingly.

Exactly. There, some work needs to be done.

jduerholt:
Here also my two cents:

-  we should enforce B023, as this can really lead to unintended behavior.
- enforcing B905, would need to set python 3.10 as minimal requirement as the `strict` keyword was introduced in 3.10
- regarding B017, I will get a student in June who should work presumably on outlier detection via robust GP regression in Bofire (https://github.com/pytorch/botorch/discussions/1311), but his first task will be to tidy-up the tests to reflect the new data structure, I will tell him then also to look at this.

The rest, I would ignore.

bertiqwerty:
All right. Since B023 is merged, let's track B017 and close the rest.

PR #203, opened 2023-05-12
Ruff/b023

jduerholt:
This PR resolves the issue with B023 and enforces from now on this ruff rule. See also the discussions in https://github.com/experimental-design/bofire/issues/199.


bertiqwerty (approved):
Looks good. I just simplified the code change a little.

PR #204, opened 2023-05-13
use sympy to obtain nonlinear constraint jacobians

### LINKED ISSUE #198: Symbolic differentiation for nonlinear constraints
Since sympy is already implicitly used (by using formulaic's differentiation): should we use symbolic differentiation to generate the derivatives of the `expression` for nonlinear constraints? Then the user would not have to compute it manually anymore...
If you are okay with this I would create a PR for this somewhen the next days.

Osburg:
This small PR adds the possibility to use sympy to obtain `jacobian_expression` in `NonlinearConstraint` automatically which is needed to compute the jacobian of nonlinear constraints.
Closes #198 .


jduerholt (approved):
This is really neat. Thx!

PR #206, opened 2023-05-13
New input scaler, ScalerEnum.NONE

simonsung06:
A new type of input scaler defined that is for applying no scalarization. Works primarily by modifying `get_scaler` to return `None`. Changes downstream from there follow due to setting `scaler` to `None`

### DIFF
--- a/bofire/data_models/surrogates/scaler.py
+++ b/bofire/data_models/surrogates/scaler.py
@@ -8,4 +8,4 @@ class ScalerEnum(Enum):
 
     NORMALIZE = "NORMALIZE"
     STANDARDIZE = "STANDARDIZE"
-    NONE = "NONE"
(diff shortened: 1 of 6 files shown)


simonsung06:
Thanks @jduerholt. 

I forgot that we had decided on the name being `IDENTITY` so I have changed it to that.
I have improved the tests slightly but I am unsure on the best practices so any advice on this would be appreciated.

Cheers

jduerholt:
Hi Simon, looks good, also testing wise. Regarding testing guidelines and best practices, we have to definitely improve and create some. But first we have to tidy them up, which should happen in June by an intern. I restarted the test of the tutorials. Let's see if it run through, if yes, I will approve, else we have to see what is going on there.

jduerholt (approved):

PR #207, opened 2023-05-16
remove version reading

jduerholt:
In `setup.py` we were reading the version erroneously from `__init__.py`. This is obsolete as we are using `pyproject.toml` and `setuptools_scm`.

Maybe we could also get rid of the complete `setup.py`. In this project, it looks for me that they also have optional dependecies in the `pyproject.toml` (https://github.com/datamol-io/molfeat/blob/main/pyproject.toml) Maybe we have to revise ours to. @bertiqwerty : what to you think?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,9 +2,6 @@
 requires = ["setuptools", "setuptools_scm[toml]", "wheel"]
 build-backend = "setuptools.build_meta"
 
-[project]
-dynamic = ["version"]
-
 [tool.setuptools_scm]
 local_scheme = "node-and-date"
 write_to = "./bofire/version.py"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,6 +2,9 @@
 requires = ["setuptools", "setuptools_scm[toml]", "wheel"]
 build-backend = "setuptools.build_meta"
 
+[project]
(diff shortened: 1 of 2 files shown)


jduerholt:
Also looking into the actual pipeline, it is stating that it installs bofire_0.1dev. Something is still wrong here ...

jduerholt:
> Also looking into the actual pipeline, it is stating that it installs bofire_0.1dev. Something is still wrong here ...

Hmm, but this happens also in botorch, when I am looking at their pipeline.

jduerholt:
> Hmm, but this happens also in botorch, when I am looking at their pipeline.

I think this happens as the pipeline is just getting the files from the master branch visout any further versioning info like tags. If one would pull the whole repo, than it should look different, I would expect.

bertiqwerty:
> Maybe we could also get rid of the complete setup.py. In this project, it looks for me that they also have optional dependecies in the pyproject.toml (https://github.com/datamol-io/molfeat/blob/main/pyproject.toml) Maybe we have to revise ours to. @bertiqwerty : what to you think?

I think it is a good idea to get rid of the `setup.py` and only rely on the `pyproject.toml` if possible.

bertiqwerty (approved):

PR #213, opened 2023-05-25
211 tag for v003

bertiqwerty:


jduerholt:
In principle, we should also run the real tests against the other versions or?

bertiqwerty:
> Have you also tried to run the other tests with 3.8? I think, it should not be a problem to add also testing of 3.8 and 3.10 as defaults. I checked in the last weeks how much compute we have so far consumed, and I had the impression that they are not counting this kind pipeline jobs, as it was all zero ...

Ok, I've added full tests for Python 3.8 and 3.10.

jduerholt:
Thanks, but I have honestly no clue why 3.8 is now failing ...

jduerholt:
I have an idea :D

bertiqwerty:
> Thanks, but I have honestly no clue why 3.8 is now failing ...

Me neither 😁. Xgboost says 3.8 is sufficient. I will try to reproduce this locally.

jduerholt:
I think it is fixed now, the problem came from merging the xgboost thing in, I was using there not the `Annotated` that you introduced later.

jduerholt (approved):
Looks good to me. Thanks Behrang!

PR #214, opened 2023-05-25
simplify annotated by using `typing_extensions` directly

bertiqwerty:
`typing_extensions` already checks if `Annotated` is part of `typing`. No need to re-implement the logic, since we now have `typing_extensions` as dependency anyway.


jduerholt (approved):
looks good to me.

PR #216, opened 2023-05-30
Feature/zero also valid

jduerholt:
This PR adds the possibility to define that for a continuous input feature zero is also a valid option. For example if you have a feature which should normally be resided in the range 3 to 5 but zero would be also valid. 

Another option would be to model it via an nchoosek constraint with only one variable inside, which is a bit awkward, I a open to discussions how to handel it.

cc: @DavidWalz  @Osburg  @dlinzner-bcs  @R-M-Lee @bertiqwerty

### DIFF
--- a/bofire/data_models/constraints/nchoosek.py
+++ b/bofire/data_models/constraints/nchoosek.py
@@ -2,9 +2,10 @@ from typing import List, Literal
 
 import numpy as np
 import pandas as pd
-from pydantic import root_validator, validator
+from pydantic import Field, root_validator, validator
+from typing_extensions import Annotated
 
-from bofire.data_models.constraints.constraint import Constraint, FeatureKeys
+from bofire.data_models.constraints.constraint import Constraint
 
 
 def narrow_gaussian(x, ell=1e-3):
@@ -23,7 +24,7 @@ class NChooseKConstraint(Constraint):
     """
 
     type: Literal["NChooseKConstraint"] = "NChooseKConstraint"
-    features: FeatureKeys
+    features: Annotated[List[str], Field(min_items=1)]
     min_count: int
     max_count: int
     none_also_valid: bool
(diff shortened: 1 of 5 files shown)


bertiqwerty:
> This PR adds the possibility to define that for a continuous input feature zero is also a valid option. For example if you have a feature which should normally be resided in the range 3 to 5 but zero would be also valid.
> 
> Another option would be to model it via an nchoosek constraint with only one variable inside, which is a bit awkward, I a open to discussions how to handel it.
> 
> cc: @DavidWalz @Osburg @dlinzner-bcs @R-M-Lee @bertiqwerty

I don't fully understand the use-case. If a variable is 0, does it mean, it should be ignored?

bertiqwerty:
> The use case would be a formulation optimization setting in which you have for example a continuous input which has a lower bound of 3 and an upper bound of 5 if it is used. But there is also the option to not use it at all in the formulation which would result to an amount of 0, which would violate the specified bounds. For this reason, I introduce the `zero_also_valid` attribute here ...

Yeah. Makes sense. Maybe, we should call the property `optional` instead of `zero_also_valid`?

bertiqwerty:
Thinking about this a little more, maybe we should not add additional properties to features that make them implicitly add constraints. I think it would be more transparent and explicit to have a small builder-method to create an n-choose-k-constraint from a feature. E.g., 
```python
constr = nchoosek_from_feature("x1")
```
instead of
```python
constr = NChooseKConstraint(features=["x1"], min_count=0, max_count=1, none_also_valid=True)
```

jduerholt:
Neat idea, but I think we still need the flag in the feature as the candidate validators do the validation based on the value and will complain if the value is out of range.

bertiqwerty:
> Neat idea, but I think we still need the flag in the feature as the candidate validators do the validation based on the value and will complain if the value is out of range.

Don't we have the same problem currently for all other n-choose-k constraints then?

jduerholt:
Die Zeit ist über es hinweggegangen.

PR #223, opened 2023-06-23
clarify that lack of support for non-continuous only applies to doe

sgbaird:


bertiqwerty:
Thanks for the clarification @sgbaird. I would just rephrase to "While some of the optimization algorithms...", because not all can handle discrete inputs.

bertiqwerty (approved):

PR #231, opened 2023-06-29
add stepwise

jduerholt:
This PR adds the possibility to compose different strategies into one. For example define a strategy in which for the first `10` experiments the random strategy is used and after this a predictive BO strategy. 

This is especially useful for closed loop stuff and the way our backend is build, as we can then just put a Stepwise strategy in, which switches automatically to BO when enough data is there. 

A few questions:

-  What do you think about the name `StepwiseStrategy`? Fine for you?
- Should the variable `Step.data_model` be renamed to `Step.strategy`?


bertiqwerty:
> This PR adds the possibility to compose different strategies into one. For example define a strategy in which for the first `10` experiments the random strategy is used and after this a predictive BO strategy.
> 
> This is especially useful for closed loop stuff and the way our backend is build, as we can then just put a Stepwise strategy in, which switches automatically to BO when enough data is there.
> 
> A few questions:
> 
>     * What do you think about the name `StepwiseStrategy`? Fine for you?
> 
>     * Should the variable `Step.data_model` be renamed to `Step.strategy`?

Thanks Johannes. Sounds interesting! I'll have a look.

bertiqwerty (commented):
Looks good to me! I left some comments. About the name one might think about `PiecewiseStrategy`. But I would also accept `StepwiseStrategy`.

bofire/data_models/strategies/stepwise.py
```diff
@@ -0,0 +1,71 @@
...
+    PolytopeSampler,
+    RejectionSampler,
+    RandomStrategy,
+    DoEStrategy,
+]
+
+
+class Step(BaseModel):
+    type: Literal["Step"] = "Step"
+    data_model: AnyStrategy
```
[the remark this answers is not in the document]
jduerholt: changed to `strategy_data`

bofire/data_models/strategies/stepwise.py
```diff
@@ -0,0 +1,71 @@
...
+    num_required_experiments: Annotated[int, Field(ge=0)]
```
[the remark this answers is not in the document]
jduerholt: You are right, but we are currently using the `num...` at a lot of different occasions already, also in the base class of `Strategy` as `num_experiments` and `num_candidates`. I would advocate for keeping it here, and make a seperate PR where we purge it out of the complete repo, fine for you?

What do you mean with different stopping criteria?

bofire/data_models/strategies/stepwise.py
```diff
@@ -0,0 +1,71 @@
...
+        # check also for domain equality
+        for i, step in enumerate(v):
+            if step.data_model.domain != values["domain"]:
+                raise ValueError(
+                    f"Domain of step {i} is incompatible to domain of StepwiseStrategy."
+                )
+        return v
+
+    @classmethod
+    def is_feature_implemented(cls, my_type: Type[Feature]) -> bool:
```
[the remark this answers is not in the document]
jduerholt: I do not think so, every step holds its own strategy_data, and when setting up the strategy_data object of a step, the strategy_data is validated and the `is_feature_implemented` method is called.

bofire/data_models/strategies/stepwise.py
```diff
@@ -0,0 +1,71 @@
...
+        return True
+
+    @classmethod
+    def is_constraint_implemented(cls, my_type: Type[Constraint]) -> bool:
+        return True
```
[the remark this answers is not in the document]
jduerholt: See comment from above.

bofire/strategies/stepwise.py
```diff
@@ -0,0 +1,78 @@
...
+class StepwiseStrategy(Strategy):
+    def __init__(self, data_model: data_model, **kwargs):
+        super().__init__(data_model, **kwargs)
+        self.steps = data_model.steps
+
+    def has_sufficient_experiments(self) -> bool:
+        return True
+
+    def _get_step(self) -> Tuple[int, Step]:  # type: ignore
+        if self.num_experiments == 0:
```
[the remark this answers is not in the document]
jduerholt: See comment above.

bofire/strategies/stepwise.py
```diff
@@ -0,0 +1,78 @@
...
+            raise ValueError(
+                f"Maximum number of candidates for step {istep} is {step.max_parallelism}."
+            )
+        # map it
+        strategy = map(step.data_model)
+        # tell the experiments
+        if self.num_experiments > 0:
+            strategy.tell(experiments=self.experiments)
+        # tell pending
+        if self.num_candidates > 0:
```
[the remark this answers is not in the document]
jduerholt: See comment above.

bofire/data_models/strategies/stepwise.py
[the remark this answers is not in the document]
bertiqwerty: I would change `num_candidates` and co. already to reduce the number of conflicts and changes when also we change this at the other places. We already have a mix anyway. But I can also live with keeping it this way and changing all later.

What I mean with different stopping criteria is that the number of experiments is not always the limiting factor, especially if you have an automated system that runs without anyone watching. Basically, the `Step`-class would need a callable that evaluates to true or false based on whatever criterion one could think of. This would be used instead of checking the number of experiments which could still be the default. This makes the implementation more flexible but is not super complicated.
For instance, the first prio ist to get an accurate Pareto-front or as high of a yield as possible, and
the second prio is to know more about the whole space if you have some experiments left or to re-do promising experiments to reduce or estimate the noise. 
You could say the first step continues until it is only exploiting, e.g., since your yield is already super high, or something more sophisticated like this https://proceedings.mlr.press/v206/ishibashi23a.html. Afterwards, the second step continues with something more explorative to increase the quality of the surrogate model everywhere.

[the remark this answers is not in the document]
jduerholt: Very smart idea, how would you mix it with the `n_experiments`, so what has to be fulfilled? `n_experiments`? The callable or both? Or do we need an extra flag for it?

The question would be how to make this callable serializable? Via individual modules that can be parameterized or via  cloupickle serialized callables?

Maybe both? And having the num_experiments just as one module. I think, I have an idea. Let me give it a try.

jduerholt:
Hi Behrang, I added more sphisticated conditional logic. The current implementation is based on a chain of conditions and it starts to check from the back. The step with the condition that matches first, is chosen. One could further extend this to include a full decision tree, but I do not think, that this is currently needed.

What do you think?

bofire/strategies/stepwise/conditions.py
```diff
@@ -0,0 +1,53 @@
...
+        return False
+
+
+CONDITION_MAP = {
+    data_models.CombiCondition: CombiCondition,
+    data_models.RequiredExperimentsCondition: RequiredExperimentsCondition,
+}
+
+
+def map(
```
bertiqwerty, line 48: Unrelated to this PR, map is actually a builtin function in Python. Maybe we should rename this everywhere.

bofire/strategies/stepwise/conditions.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class CombiCondition(Condition):
+    def __init__(self, data_model: data_models.CombiCondition) -> None:
+        self.conditions = [map(c) for c in data_model.conditions]  # type: ignore
+        self.n_required_conditions = data_model.n_required_conditions
+
+    def evaluate(self, domain: Domain, experiments: pd.DataFrame) -> bool:
+        n_matched_conditions = 0
+        for c in self.conditions:
```
bertiqwerty, line 34: I'd expect a combination of conditions to combine them rather with `or`  or `and`. Not sure what the use case of counting the number `True`s could be.

bofire/strategies/stepwise/stepwise.py
```diff
@@ -0,0 +1,75 @@
...
+class StepwiseStrategy(Strategy):
+    def __init__(self, data_model: data_model, **kwargs):
+        super().__init__(data_model, **kwargs)
+        self.steps = data_model.steps
+
+    def has_sufficient_experiments(self) -> bool:
+        return True
+
+    def _get_step(self) -> Tuple[int, Step]:  # type: ignore
+        for i in range(len(self.steps) - 1, -1, -1):
```
bertiqwerty, line 52: Why do you start with the last step instead of the first step?

bertiqwerty (commented):
Hi Johannes, I agree that we do not need a decision tree.

bofire/strategies/stepwise/conditions.py
jduerholt: Yep, but I was lazy and wanted to have it in one class. You can have `and` via a list of two conditions and `n_required_conditions=2`  and `or` via also two conditons and `n_required_conditions=1`. 

Of course, we can also replace the `n_required_conditions` field by an `operator` field which takes a literal of being either `"and"` or `"or"`. But this would be less expressive as my solution.

What do you think?

bofire/strategies/stepwise/conditions.py
jduerholt: Yep, we should do this in a seperate PR for the whole code base.

bofire/strategies/stepwise/conditions.py
bertiqwerty: Ah. Ok. Got it.

bofire/strategies/stepwise/stepwise.py
jduerholt: For me it is easier like this:  the step that should be applied is always the step which comes before the first step with an unfullfilled condition. 

For example, if you have three steps and the conditions looks like this `[True, True, False]`, then we would proceed with the step with index 10`. For me it is then easier to loop from the back and stop with the first condition where one gets a fulfilled condition instead starting from the front and looking for the first `False` and then returning the step before the `False`.

Do you agree?

bertiqwerty: I would start with the first step that evaluates to `True` and care about the later steps when it is there turn. This seems simpler to me. When the condition of the first step changes to `False`  I would go to the next step. Basically, as in your initial version with the only difference that numbers of experiments is not the only possible criterion.

jduerholt: Hmm, but this would change the whole logic of the implemented conditions. For example the `RequiredExperimentsCondition` with `n_required_experiments` will neve turn to `False` when it is once turned `True`. For this reason, I start from the back to always use the latest applicable step.

bertiqwerty: Ah, yes. That's right. I vote for changing the inequality-sign in `RequiredExperimentsCondition` and I would rename it to `NumberOfExperimentsCondition`. The reason why I think we should start with the first step is that it is what people are used to. Starting with the first step. I think it will be confusing to start with the last step.

jduerholt: Ok, I will change it.

jduerholt: I implemented it according to your suggestion, should we then for the `NumberOfExperimentsCondition` implement the possibility of infinity experiments, so that it holds true forever, so that the user can decide if he wants to end an campaign with a step or not ...

one could encode it via `n_experiments = -1`

What do you think?

bertiqwerty: Ah. You mean the last step could be forever? Yes, we can do that. For this an always `True` condition might be more explicit. But your suggestion is also fine.

jduerholt: Yes, last step forever ;) I like the idea of an `AlwaysTrueCondition`. I will add it, and validation that this can only be used in the last step.

jduerholt: @bertiqwerty: Done, now it should be finally ready. Can you have a look again?

bertiqwerty (approved):

PR #239, opened 2023-07-10
Hyperparameter Optimization for Trainable Surrogates

jduerholt:
This PR adds the possibility to optimize hyperparameters of bofire surrogates in an easy manner with bofire. For this purpose, trainable surrogates can have a configuration object attached in which the hyperparameters and its ranges etc. are defined, together with a method how to update them. 

Changing this configuration object which is part of the surrogate's data model allows to tweak the applied hyperparameter optimization. 

Furthermore, this PR adds:
- `FactorialStrategy` for purely discrete/categorical domains
- LinearSurrogate to easily setup linear models

In a next step, one could enable automatic hyperparameter optimization also in strategies.

### DIFF
--- a/bofire/data_models/surrogates/single_task_gp.py
+++ b/bofire/data_models/surrogates/single_task_gp.py
@@ -1,19 +1,98 @@
-from typing import Literal
+from typing import Literal, Optional
 
+import pandas as pd
 from pydantic import Field
 
-from bofire.data_models.kernels.api import AnyKernel, MaternKernel, ScaleKernel
+from bofire.data_models.domain.api import Inputs
+from bofire.data_models.enum import RegressionMetricsEnum
+from bofire.data_models.features.api import CategoricalInput
+from bofire.data_models.kernels.api import (
+    AnyKernel,
+    MaternKernel,
+    RBFKernel,
+    ScaleKernel,
+)
 from bofire.data_models.priors.api import (
     BOTORCH_LENGTHCALE_PRIOR,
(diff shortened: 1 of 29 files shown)


bofire/data_models/surrogates/single_task_gp.py
```diff
@@ -1,19 +1,98 @@
...
+    hyperstrategy: Literal[
+        "FactorialStrategy", "SoboStrategy", "RandomStrategy"
+    ] = "FactorialStrategy"
+
+    @staticmethod
+    def _update_hyperparameters(
+        surrogate_data: "SingleTaskGPSurrogate", hyperparameters: pd.Series
+    ):
+        def matern_25(ard: bool, lenghtscale_prior: AnyPrior) -> MaternKernel:
+            return MaternKernel(nu=2.5, lengthscale_prior=lenghtscale_prior, ard=ard)
```
R-M-Lee, line 52: @jduerholt typo here: lenghtscale_prior --> lengthscale_prior

jduerholt: Thanks for spotting, I just fixed it.

R-M-Lee:
Looks great by the way. What do you think about allowing GroupKFold in addition to KFold, or even allowing the user to specify their own cross validation iterator?

bertiqwerty:
If Robert says it looks great, we can approve it I guess :)

bertiqwerty (approved):

PR #240, opened 2023-07-12
add random state to hyperfit

jduerholt:


jduerholt:
add random state to hyperfit

PR #242, opened 2023-07-17
Allowing int for category with str(int) e.g 1 and '1' are same category but not 'one'

swagataroy123:
@jduerholt , please look at this


jduerholt (approved):
Looks good to me. Thanks!

PR #248, opened 2023-08-01
Give information about HSL linear solvers in documentation

### LINKED ISSUE #244: IPOPT: check for availability of HSL-Coin
Recently I noticed in a different context that IPOPT can be much faster (up to 10x speedup) using the linear solvers from [HSL-Coin](https://licences.stfc.ac.uk/product/coin-hsl) compared to MUMPS. 
It could be worthful to check if this is also the case in case of our applications (are there any except for doe?). If yes we should consider checking for availability of HSL solvers on the users system and make use of them if they are present.

Osburg:
Closes #244


jduerholt (approved):
Thanks a lot.

PR #259, opened 2023-08-11
Doe categorical second attempt

ufukguenes:
Reopens #243


ufukguenes:
@jduerholt regarding pyright.
As far as I can tell pyright has two problems with the code. First, when using the get_features method the returned type is of type Features which does not guarantee that I can use them like a RelaxableBinaryInput, however I can be sure that they are only RelaxableBinaryInputs as that's what I am asking for. Maybe I should add a type ignore to these instances. A cleaner option would be to make the return type of get_features generic. 
The second problem is, that the NChooseKConstraint is allowed to have Inputs which are not ContinousInputs or DiscreteInputs which might lead to a problem, but this this also not taken in to account with the previous implementation. So here as well adding type ignore could be a solution

jduerholt:
Ok, then just `type: ignore` them.

bofire/strategies/doe_strategy.py
```diff
@@ -20,22 +37,139 @@ def __init__(
...
+                n_experiments=_candidate_count,
+                fixed_experiments=self.candidates,
+                verbose=self.data_model.verbose,
+                partially_fixed_experiments=self._partially_fixed_experiments_for_next_design,
+                categorical_groups=all_new_categories,
+            )
+        elif self.data_model.optimization_strategy in [
+            "branch-and-bound",
+            "default",
+            "partially-random",
```
dlinzner-bcs, line 153: why do we have again all these strategies that map to the same optimizer?

ufukguenes: depending on which Inputs are in the domain, need to use the BaB regardless of which strategy is provided. For example, if there are CategoricalInputs and NChooseKConstraints and the strategy is set to partially-random. We need solve the NChosseKConstraints by using the already existing strategy where we randomly sample solution, but the CategoricalInputs have to be solved with the BaB.

dlinzner-bcs: Ok.  Can you please add comments that explain these cases and what the strategies mean? And maybe "partially-random" is not very idiomatic. It is referring to  the current NChooseKConstraint implementation, correct? Maybe call it "random-nchoosek-realization"?

bofire/strategies/doe/utils_features.py
```diff
@@ -0,0 +1,147 @@
...
+from bofire.data_models.features.continuous import ContinuousInput
+
+
+class RelaxableBinaryInput(ContinuousInput):
+    """Base class for all binary input features. It behaves like a continuous inputs to allow for easy relaxation.
+    It is not a true binary variable as it does not allow the variable to be either 0 or 1 while prohibiting all values
+    in between.
+
+    """
+
```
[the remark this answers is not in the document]
jduerholt: bounds: Tuple(float, float) = (0,1)

bofire/strategies/doe/utils_features.py
```diff
@@ -0,0 +1,147 @@
...
+    """Feature with discrete ordinal values allowed in the optimization.
+
+    Attributes:
+        key(str): key of the feature.
+        values(List[float]): the allowed discrete values during the optimization.
+    """
+
+    values: List[float]
+
+    def __init__(self, **kwargs):
```
jduerholt, line 69: not needed, do it in the above.

bofire/strategies/doe/utils_features.py
```diff
@@ -0,0 +1,147 @@
...
+from typing import List, Tuple
+
+import numpy as np
+import pandas as pd
+from pydantic import root_validator
+
+from bofire.data_models.features.continuous import ContinuousInput
+
+
+class RelaxableBinaryInput(ContinuousInput):
```
jduerholt, line 12:
```suggestion
class _RelaxableBinaryInput(ContinuousInput):
```

ufukguenes:
@jduerholt @dlinzner-bcs as discussed, I removed the two features I added and use a Dict/ List to keep track of the relaxed variables

jduerholt:
> @jduerholt @dlinzner-bcs as discussed, I removed the two features I added and use a Dict/ List to keep track of the relaxed variables

Great. Thank you very much!

bofire/strategies/doe_strategy.py
```diff
@@ -20,22 +32,138 @@ def __init__(
     ):
         super().__init__(data_model=data_model, **kwargs)
         self.formula = data_model.formula
+        self.data_model = data_model
+        self._partially_fixed_experiments_for_next_design = None
+
+    def tell(
```
jduerholt, line 38: Hmm, for what is this `tell` needed?

tests/bofire/data_models/specs/strategies.py
```diff
@@ -125,6 +125,8 @@
     lambda: {
         "domain": domain.valid().obj().dict(),
         "formula": "linear",
+        "optimization_strategy": "default",
+        "verbose": False,
```
jduerholt, line 129: What is this ´verbose´ flag doing?

tutorials/basic_examples/Reaction_Optimization_Example.ipynb
jduerholt, line 1: Is this indended that so much is deleted from all these notebooks? No, or? It looks like, that all the output is gone. Have you modified them? This relates to all the notebooks ...

bofire/strategies/doe_strategy.py
jduerholt: Especially as you overwrite `tell` and not just `_tell`. That's looks a bit dangerous ...

bofire/strategies/doe_strategy.py
```diff
@@ -20,22 +32,138 @@ def __init__(
...
+                raise ValueError(
+                    f"No variance in experiments for fixed feature {feature.key}"
+                )
+        self._tell()
+
+    def _tell(self) -> None:
+        self.set_candidates(
+            self.experiments[self.domain.get_feature_keys(includes=Input)]
+        )
+        raise NotImplementedError(
```
jduerholt, line 78: Also this is a bit cumbersome, the idea is that one has a generic `tell` method in `Strategy`, which is then calling a strategy dependent `_tell`. In principal, I do not see why a tell is needed at all, we have already the pending candidates to take previous experiments into account ...

jduerholt (changes requested):
Thank you very much. Looks much better know. I let some small comments. We should especially have a closer look what happens to the tutorial notebooks.

bofire/strategies/doe_strategy.py
dlinzner-bcs: Whoops. I told Ufuk to implement tell. My bad. @ufukguenes let us use set_candidates() then instead.

tests/bofire/data_models/specs/strategies.py
ufukguenes: If true, the method prints out information about the optimization process. So for BaB, the number of branches explored and the current lower bound.

ufukguenes:
@jduerholt I converted the tell method to set_candidates. I had to over wright the method in the child class as we want to support partially defined experiments that can be filled up by the optimizer. We think that is a valuable feature by it self. The parent set_candidates checked for None values which we use to define the partially defined experiments.

jduerholt:
> @jduerholt I converted the tell method to set_candidates. I had to over wright the method in the child class as we want to support partially defined experiments that can be filled up by the optimizer. We think that is a valuable feature by it self. The parent set_candidates checked for None values which we use to define the partially defined experiments.

This is a very smart idea. In a seperate PR, we should update the base class to make it a general feature.

jduerholt (approved):
@ufukguenes: thanks for all your efforts!

@ufukguenes  @dlinzner-bcs : Feel free to merge it ;)

PR #260, opened 2023-08-14
Feature/categorical molecular2

jduerholt:
This PR adds the CategoricalMolecularInput feature.

### DIFF
--- a/bofire/data_models/domain/features.py
+++ b/bofire/data_models/domain/features.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import itertools
 import warnings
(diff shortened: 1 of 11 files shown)


simonsung06:
@jduerholt 
Changes and additions look good to me.

One comment about model building when we have the different input types. As it is now, we would stick to using MixedSingleTaskGP, for example, when we have CategoricalMolecularInput with them treated as categoricals? i.e. Not using an additional hypothetical "molecular_kernel" argument.

simonsung06 (approved):
All good!

PR #262, opened 2023-08-15
Added outlier detection tutorial

swagataroy123:
@jduerholt ,
Added outlier detection tutorials


jduerholt (changes requested):
Hi Swagata, 

regarding the `SMOKE_TEST`: 

It is a flag used by the testing pipelines, you have now used it in a way that parts of the notebooks are not tested at all, it should be more used in the way that every part is also run through in the tests but with a speedup, that means if `SMOKE_TEST` is used, we use much less data, much less iterations and so fourth, of course the results will then not be significant but we can test the functionality of the notebook. 

Please change it accordingly. The displayed output of the notebooks in the Repo should of course be from a complete run without smoke test, as it is also currently.

Best,

Johannes

PR #264, opened 2023-08-18
Dominik iterative branch and bound

dlinzner-bcs:
- add a branch-and-bound strategy that iteratively adds experiments 
- useful for problems with many categoricals/nchoosek constraints to still get a result

### DIFF
--- a/bofire/strategies/doe_strategy.py
+++ b/bofire/strategies/doe_strategy.py
@@ -160,11 +160,15 @@ class DoEStrategy(Strategy):
                 )
                 _adapted_partially_fixed_candidates = pd.concat(
                     [
-                        adapted_partially_fixed_candidates,
+                        _adapted_partially_fixed_candidates,
                         _design.round(6).tail(1),
                     ],
                     axis=0,
                     ignore_index=True,
+                )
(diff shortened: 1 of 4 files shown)


bofire/strategies/doe_strategy.py
```diff
@@ -1,189 +1,227 @@
...
+                domain=new_domain,
+                model_type=self.formula,
+                n_experiments=_candidate_count,
+                fixed_experiments=None,
+                verbose=self.data_model.verbose,
+                partially_fixed_experiments=adapted_partially_fixed_candidates,
+                categorical_groups=all_new_categories,
+                discrete_variables=new_discretes,
+            )
+        elif self.data_model.optimization_strategy in [
```
dlinzner-bcs, line 136: This is the strategy implementation.

jduerholt:
@dlinzner-bcs : I fixed the bug, which caused the crash in the testpipeline, just merge main into this feature branch, then it should work again ;)

jduerholt (approved):
Fine for me, but do you know why the diff is looking so strange? Why are all these files shown which have not been touched in the PR?

PR #266, opened 2023-08-22
fix number of experiments condition for zero experiments

jduerholt:
This PR fixes a bug in the `NumberOfExperimentsCondition` when the `StepwiseStrategy` has no experiments ...


bofire/strategies/stepwise/conditions.py
simonsung06, line 1: Minor thing. Type hint for experiments argument does not suggest None is allowed.

jduerholt: Good catch!

simonsung06 (approved):
All good!

PR #268, opened 2023-08-23
Bayesian optimization on molecules test

swagataroy123:
@jduerholt ,
Replicated https://github.com/leojklarner/gauche/blob/main/notebooks/Bayesian%20Optimisation%20Over%20Molecules.ipynb


jduerholt (changes requested):
Hi Swagata,

thank you very much. I just let some comments regardin the benchmark stuff. Start with fixing them and I will later add comments regarding the rest.

Best,

Johannes

bofire/benchmarks/LookUpTableBenchmark.py
```diff
@@ -0,0 +1,58 @@
...
+    It takes in a file with inputs and outputs and returns output for matching inputs from that file.
+
+    Args:
+        Benchmark: Subclass of the Benchmark function class.
+    """
+
+    def __init__(
+        self,
+        domain: Domain,
+        LookUpTable: pd.DataFrame,
```
jduerholt, line 19:
```suggestion
        lookup_table: pd.DataFrame,
```

just call it `lookup_table`, arguments are always snake case. You also have to adjust below.

bofire/benchmarks/LookUpTableBenchmark.py
```diff
@@ -0,0 +1,58 @@
+import numpy as np
+import pandas as pd
+
+from bofire.benchmarks.benchmark import Benchmark
+from bofire.data_models.domain.api import Domain
+
+
+class LookUpTableBenchmark(Benchmark):
```
jduerholt, line 8:
```suggestion
class LookupTableBenchmark(Benchmark):
```

It is one word "lookup".

bofire/benchmarks/LookUpTableBenchmark.py
```diff
@@ -0,0 +1,58 @@
...
+        Args:
+            X (pd.DataFrame): Input values with input columns only.
+
+        Returns:
+            pd.DataFrame: output values from the LookUpTable. Columns are ouput keys and valid_output keys.
+        """
+        location = []
+        for i in X.index:
+            condition = np.ones(len(self.LookUpTable), dtype=bool)
+            for k in self.domain.inputs.get_keys():
```
jduerholt, line 46: This looping over the single entries is kind of cumbersome. Try something like this:

```
import pandas as pd

# create a sample dataframe
df = pd.DataFrame({'A': [1, 2, 3], 'B': ['a', 'b', 'c'], 'C': [True, False, True]})

# create a sample series
s = pd.Series([2, 'b', False])

# check if the values of the series show up in a row of the dataframe
mask = df.isin(s)

# get the index of the rows that match the values in the series
index = mask.index[mask.all(axis=1)]

print(index)
```

If it matches all values, you can then use the index to return via loc the correct output values.

bofire/benchmarks/LookUpTableBenchmark.py
```diff
@@ -0,0 +1,58 @@
...
+                condition = condition & (self.LookUpTable[k] == X[k][i]).to_numpy()
+            if np.where(condition)[0].size != 0:
+                location.append(np.where(condition)[0][0])
+            else:
+                raise ValueError(f"Input combination {i} not found in Look up table")
+        X_temp = self.LookUpTable.loc[location]
+        X_temp.index = pd.RangeIndex(len(X_temp))
+        Y = pd.DataFrame()
+        for k in self.domain.outputs.get_keys():
```
jduerholt, line 55: The dataframe which is read in in the first place should already have the valid keywords. When you then have the complete set of matching indices, you can just loc the columns and rows that you need and return them with resetted index.

bofire/benchmarks/LookUpTableBenchmark.py
swagataroy123: isin() does not work with input keys having similar categories. I used pd.merge which is also a nice substitute for for loops

bofire/benchmarks/LookupTableBenchmark.py
```diff
@@ -0,0 +1,60 @@
...
+        df = pd.merge(X, X_temp, how="left", indicator=True)
+        if (
+            df.loc[df._merge == "left_only", df.columns != "_merge"].proxy_index.size
+            != 0
+        ):
+            indices = df.loc[
+                df._merge == "left_only", df.columns != "_merge"
+            ].proxy_index.to_list()
+            raise ValueError(f"Input combination {indices} not found in Look up table")
+        Y = X_temp[self.domain.outputs.get_keys()]
```
jduerholt, line 56: Do it in one line ;)

jduerholt (approved):
Looks good. Thanks!

PR #269, opened 2023-08-24
add generic benchmarl module

jduerholt:
This PR adds a generic benchmark class which takes a `Domain` and a `Callable` as input.


simonsung06 (approved):
A really useful added function :)

PR #271, opened 2023-08-25
Feature/new acqfs

jduerholt:
This PR introduces the new acqfs from botorch qLogEI and qLogNEI which seems to perform better than the orignal ones. In addition it is now also possible to select an acqf for qParEgo.


simonsung06 (approved):
Cheers Johannes for implementing

PR #272, opened 2023-08-25
add aggregate by mean

jduerholt:
This PR adds the possibility to aggregate experiments also by median and not only by mean.


simonsung06 (approved):
Looks good to me. Thanks.

PR #273, opened 2023-08-28
Fix import of `get_acquisition_function`

jduerholt:
This PR fixes the issue with the new import structure for `get_acquisition_function` in the current botorch devel version.


simonsung06 (approved):
Good to go

PR #274, opened 2023-08-30
Fix SOBO with one constrained objective

jduerholt:
This PR fixes the problem when having only a constrained objective in the SOBO strategy. This will be shortly revised again, when Gabe has finished implementing the new handling of output constraints in botorch.


jduerholt:
> This PR fixes the problem when having only a constrained objective in the SOBO strategy.

PR #278, opened 2023-09-06
Implement ENTMOOT in Bofire

TobyBoyne:
Initial progress on converting a BoFire problem to an ENTMOOT optimisation problem.


TobyBoyne:
As shown in `strategies/entmoot/entmoot_example.py`, the `ask()` method now matches the syntax for other strategies.

Is there any functionality that is missing? If not, I will add tests (especially for the converting between features/constraints), and better documentation.

bofire/strategies/entmoot/entmoot_example.ipynb
jduerholt, line 1: This file should go under the `tutorials` section where also the other notebooks are.

bofire/strategies/entmoot/entmoot_benchmarks.py
jduerholt, line 1: This should go either to the tests or the benchmarks or in the notebook.

bofire/data_models/strategies/enting.py
```diff
@@ -0,0 +1,41 @@
...
+    DiscreteInput,
+    Feature,
+)
+from bofire.data_models.objectives.api import MinimizeObjective, Objective
+from bofire.data_models.strategies.strategy import Strategy
+
+
+class EntingStrategy(Strategy):
+    type: Literal["EntingStrategy"] = "EntingStrategy"
+    enting_params: Dict[str, Any] = {}
```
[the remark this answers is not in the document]
jduerholt: As we have also the XGBSurrogate in BoFire, I am wondering, if one could export the hyperparams of the XGB surrogates that are used to the user as we are also doint it for the botorch based strategies. You can have a look here: https://github.com/experimental-design/bofire/blob/a4ca3eb2ecf8348e5c387cbef273191c1ba54d45/bofire/data_models/strategies/predictives/botorch.py#L35 and here: https://github.com/experimental-design/bofire/blob/main/bofire/data_models/surrogates/xgb.py

jduerholt (changes requested):
Hi @TobyBoyne,

thank you very much for your PR and contributing to BoFire. I let some high level comments. Feel free to question and discuss them. I am looking forward to be able to use ENTMOOT in BoFire.

Best,

Johannes

bofire/data_models/strategies/enting.py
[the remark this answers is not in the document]
TobyBoyne: I agree for the `enting_params`, and I will create an EntingSurrogate to include the parameters (these are unique to ENTMOOT and I'll have to manually collate them).

The `solver_params` attribute is a little trickier - this is passed directly to the solver, and so is solver agnostic. Since we don't have a list of every solver, and every associated set of parameters, I think this might be trickier to type hint. A solution is to just not allow passing anything to the solver, or use a default set of options, but you then lose some freedom. This mostly comes down to ENTMOOT using Pyomo where BoFire uses ipopt: losing freedom to use arbitrary solvers might not be such a bad thing.

bofire/data_models/strategies/enting.py
```diff
@@ -0,0 +1,41 @@
...
+    @classmethod
+    def is_feature_implemented(cls, my_type: Type[Feature]) -> bool:
+        return my_type in [
+            CategoricalInput,
+            DiscreteInput,
+            CategoricalDescriptorInput,
+            ContinuousInput,
+            ContinuousOutput,
+        ]
+
```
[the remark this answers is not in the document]
TobyBoyne: Yes, ENTMOOT does support multiobjective optimization. An example with the DLTZ2 benchmark has been added to the notebook.

bofire/data_models/strategies/enting.py
```diff
@@ -0,0 +1,41 @@
...
+    @classmethod
+    def is_objective_implemented(cls, my_type: Type[Objective]) -> bool:
```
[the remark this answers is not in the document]
TobyBoyne: ENTMOOT doesn't natively support maximization, but I have multiplied the values in the `fit` and `predict` function by `-1`. I've used the Ackley benchmark to demonstrate this (although I'm not sure why this uses `MaximizeObjective`, since Ackley looks to be a minimization problem as far as I can tell).

TobyBoyne:
Hi @jduerholt,

Thank you for the review - lots of helpful feedback!

I still haven't addressed the issue of hint typing parameters - I started looking into the surrogates, but I'm not sure I understand what function they have? I can implement the types as a simple class that inherits from BaseModel, but I'd rather use surrogates if that is the proper way to do things. As mentioned in the comments, I can't use any of the provided surrogates since they don't have the same attributes (ENTMOOT uses `lbgm`, and `xgboost` is not yet supported.

Kind regards,
Toby

TobyBoyne:
For reference, here is a mock-up of what the `enting_params` looks like. Note that the parameters in TrainParams are defined by `lgbm`, and I can't find a comprehensive Type of all possible options.

```python

from typing import Literal

class EntingParams:
    unc_params: 'UncParams'
    tree_train_parameters: 'TreeTrainParams'

class UncParams:
    beta: float = 1.96 #>0
    acq_sense: Literal["exploration", "penalty"] = "exploration"
    dist_trafo: Literal["normal", "standard"] = "normal"
    dist_metric: Literal["euclidean_squared", "l1", "l2"] = "euclidean_squared"
    cat_metric: Literal["overlap", "of", "goodall4"] = "overlap"


class TreeTrainParams:
    train_lib: Literal["lgbm"] = "lgbm"
    train_params: 'TrainParams'

class TrainParams:
    # this is determined by lightbgm
    # the typing in the package uses Dict[str, Any]
    # the defaults used by ENTMOOT are shown below
    objective: str = "regression"
    metric: str = "rmse"
    boosting: str = "gbdt"
    num_boost_round: int = 100
    max_depth: int = 3
    min_data_in_leaf: int = 1
    min_data_per_group: int = 1
    verbose: int = -1
```

jduerholt:
Thanks @TobyBoyne, I was not at home over the weekend. I will have a detailed look tmr. Best, Johannes

jduerholt:
Here is the first version of the mentioned PR: https://github.com/experimental-design/bofire/pull/283

TobyBoyne:
Hi @jduerholt,

Yes, I think that all makes sense - thank you for the explanation. So to confirm, as an example, the `_fit()` logic should be moved to `LGBMSurrogate`, but the `_ask()` method should remain in the strategy? I'm looking through `BotorchStrategy._ask()` and can't find any references to the surrogate - is the surrogate not used in this method?

Also, in terms of moving forward - do I have write permissions for the `feature/surrogates` branch/PR? If so, should I implement the `EntingSurrogates` there, and if not, should I wait for that pull request to be accepted before creating the EntingSurrogates?

Kind regards,
Toby

bofire/data_models/strategies/enting.py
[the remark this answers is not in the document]
R-M-Lee: > I am not a big fan of having dictionary fields that can have anything inside, as it is somehow contraty of using a validation framework like pydantic. Can you provide the possible parameters and its defaults and types and validation, that can be used as fields of the data model?

This change should be made directly in entmoot (@spiralulam wanted to do it); Johannes's argument is in general a good one and not just valid here

spiralulam:
Good point! Will work on it for Entmoot.

Am Di., 12. Sept. 2023 um 13:23 Uhr schrieb Robert Lee <
***@***.***>:

jduerholt:
> I'm looking through `BotorchStrategy._ask()` and can't find any references to the surrogate - is the surrogate not used in this method?

In `botorch`, the `model` is attached to the acquistion function object, for this reason we first get an acquistion function object in `_ask` and then use this object which has the model inside for the optimization: https://github.com/experimental-design/bofire/blob/a4ca3eb2ecf8348e5c387cbef273191c1ba54d45/bofire/strategies/predictives/botorch.py#L256 and here: https://github.com/experimental-design/bofire/blob/a4ca3eb2ecf8348e5c387cbef273191c1ba54d45/bofire/strategies/predictives/sobo.py#L60.

The `self.model` object is created here on the basis of the surrogates: https://github.com/experimental-design/bofire/blob/a4ca3eb2ecf8348e5c387cbef273191c1ba54d45/bofire/strategies/predictives/botorch.py#L113

In the call to `compatibilize`, the individual surrogates are compile in one botorch `ModelList` object.

>  So to confirm, as an example, the _fit() logic should be moved to LGBMSurrogate, but the _ask() method should remain in the strategy? 

You would have it in both the surrgate and the strategy, but you can call the in the strategies `_fit` the `_fit` of the surrogate, but I am not sure if this is currently needed, as Entmoot is currently not supporting different surrogate specs for different outputs. So you could do it independly from each other for now.

> Also, in terms of moving forward - do I have write permissions for the feature/surrogates branch/PR? If so, should I implement the EntingSurrogates there, and if not, should I wait for that pull request to be accepted before creating the EntingSurrogates?

I try to get the PR trough the latest tmr, but in terms of moving forward, I would recommend that you just create a new branch/PR in which you first just implement the `LGBMSurrogate` (including tests) as this is the first thing that is needed. In the meantime, I will then finalize my PR and you can then implement the `EntingSurrogates` and the `EntingStrategy` on top of both.

For the `LGBMSurrogate`, you can draw inspiration from the `XGBSurrogate`.

What do you think?

jduerholt:
When we have the `EntingStrategy` included we will also refactor the directory strategy for the botorch based strategies into `strategies/predictives/botorch/`.

TobyBoyne:
> I would recommend that you just create a new branch/PR in which you first just implement the `LGBMSurrogate` (including tests) as this is the first thing that is needed

Happy to work on that! Thanks for all the reviewing and help, very much appreciated!

TobyBoyne:
I've made an initial commit for the surrogate, however in order to implement the regression in #285, I need the utils in this PR - any thoughts on how I can overcome that?

TobyBoyne:
I've been working on implementing changes in the Entmoot codebase. Since this was last discussed, there are a few things that have changed:

1. We have discussed that it is not necessary to split this into a surrogate. This PR contains a working strategy implementation, and splitting this into a strategy is potential future work.
2. The typing issue has been partially resolved with the addition of the `EntingParams`. This has been implemented in entmoot - see that repo for more information. The solver still accepts Dict[any], but that's somewhat out of our hands in that Pyomo doesn't have an extensive list of arguments.
3. Other changes include matching entmoot's Constraints to be similar to bofire's, and creating an interface for maximisation problems.

TobyBoyne (commented):
The commit history and conversation on the PR should address all requested changes.

R-M-Lee (approved):

jduerholt:
I will also review it ;) Totally overlooked this :D

bofire/data_models/strategies/predictives/enting.py
```diff
@@ -0,0 +1,54 @@
...
+    MaximizeObjective,
+    MinimizeObjective,
+    Objective,
+)
+from bofire.data_models.strategies.predictives.predictive import PredictiveStrategy
+
+
+class EntingStrategy(PredictiveStrategy):
+    type: Literal["EntingStrategy"] = "EntingStrategy"
+    enting_params: EntingParams = Field(default_factory=EntingParams)
```
jduerholt, line 30:
```suggestion
    enting_params: EntingParams = Field(default_factory=lambda: EntingParams())
```

Is this `EntingParams` object something which can be serialized/desirialized by pydantic?

bofire/strategies/predictives/enting.py
```diff
@@ -0,0 +1,128 @@
...
+        self._solver_params = data_model.solver_params
+        self._learn_from_candidates_coeff = data_model.learn_from_candidates_coeff
+
+    def _init_problem_config(self) -> None:
+        cfg = domain_to_problem_config(self.domain)
+        self._problem_config: ProblemConfig = cfg[0]
+        self._model_pyo: pyo.ConcreteModel = cfg[1]
+
+    @property
+    def input_preprocessing_specs(self) -> TInputTransformSpecs:
```
jduerholt, line 38: If not needed, can you just remove this method, or is it enforced by the base class?

bofire/strategies/predictives/enting.py
```diff
@@ -0,0 +1,128 @@
...
+from pydantic import PositiveInt
+
+import bofire.data_models.strategies.api as data_models
+from bofire.data_models.features.api import TInputTransformSpecs
+from bofire.data_models.objectives.api import MaximizeObjective
+from bofire.strategies.predictives.predictive import PredictiveStrategy
+from bofire.utils.entmoot import domain_to_problem_config
+
+
+class EntingStrategy(PredictiveStrategy):
```
jduerholt, line 18: Could you also add tests for the methods in this class?

bofire/data_models/strategies/predictives/enting.py
jduerholt: Furthermore, we are introducing with this a dependency in the data model to an external library. So far, we tried to avoid this, as we want to be able to perform a minimal installation of the data models without the need for external optimizer libs. Can you wrap this EntingParams thingy for examples as a pydantic class and map it in the functional strategy to the actual object from the entmoot lib? @bertiqwerty what do you think?

jduerholt (changes requested):
Hi Toby,

looks overall great. I just let a few comments. 

One other thing is still missing: We need to test, that the data model is serializeable and deserializeable to deploy it behind a fastAPI. For this reason: Could you add an example config of an `EntingStrategy` here? As soon it is added there, it will be tested for serialization and deserialization ;)

Best,

Johannes

bofire/data_models/strategies/predictives/enting.py
bertiqwerty: Agreed. We should keep the data models as self-contained as possible.

TobyBoyne: Hi @jduerholt, could you please clarify this? The entmoot package current has some object that lists all the parameters that it takes. Would you like me to duplicate this list in bofire? This feels like duplication of code which may result in issues (if entmoot gets updated, we would need to update this list in bofire too).

Note the EntingParams is a `dataclass`, so we can just use `pydantic.dataclass`. This means going to/from pydantic will be very easy.

bertiqwerty: I fully agree with @jduerholt. Code duplication sounds evil due to obvious reasons. On the other hand however, code duplication means less dependencies and less coupling which can be more important. As in this case. As Johannes said, since ENTMOOT is not part of BoFire we will need to adapt to changes in the ENTMOOT repo anyway. Our tests should cover all possible parameters so that problems are detected (might be already the case, haven't checked).

bofire/strategies/predictives/enting.py
TobyBoyne: Frustratingly, it's it is needed. The parent method is an `abstractmethod`. Also, if the function returns `None`, then the line below throws an error, since you're trying `None.items()`.

https://github.com/experimental-design/bofire/blob/4790d637016f4e482ae8bafa6cb711b66e26b99e/bofire/data_models/domain/features.py#L502

jduerholt:
Hi Toby, thank you very much. I will have a look the latest tmr. Best, Johannes

bofire/strategies/predictives/enting.py
```diff
@@ -0,0 +1,131 @@
...
+            if candidate_count > 1:
+                self._add_candidate_as_experiment(candidate)
+
+        return pd.concat(candidate_lst)
+
+    def _fit(self, experiments: pd.DataFrame):
+        input_keys = self.domain.inputs.get_keys()
+        output_keys = self.domain.outputs.get_keys()
+
+        X = experiments[input_keys].to_numpy()
```
jduerholt, line 106: How is `_enting` handling missing values in `y`? The experiments data frame can contain `nan` for the outputs. You can postprocess them in different ways via ´domain.outputs.preprocess_one_valid_output´, `domain.outputs.preprocess_all_valid_outputs`, and `preprocess_any_valid_output`. Here you should at least use `preprocess_any_valid_output`, to only get the valid measurements and do not have a completely empty row in `y`.

bofire/utils/entmoot.py
```diff
@@ -0,0 +1,151 @@
...
+    Args:
+        domain (Domain): the definition of the optimization problem.
+
+    Returns:
+        A tuple (problem_config, model_pyo), where problem_config is the problem definition
+        in an ENTMOOT-friendly format, and model_pyo is the Pyomo model containing constraints.
+    """
+    problem_config = ProblemConfig()
+
+    for input_feature in domain.inputs:
```
jduerholt, line 40:
```suggestion
    for input_feature in domain.inputs.get():
```

bofire/utils/entmoot.py
```diff
@@ -0,0 +1,151 @@
...
+        _bofire_feat_to_entmoot(problem_config, input_feature)
+
+    for output_feature in domain.outputs:
```
jduerholt, line 43:
```suggestion
    for output_feature in domain.outputs.get():
```

bofire/utils/entmoot.py
```diff
@@ -0,0 +1,151 @@
...
+        _bofire_output_to_entmoot(problem_config, output_feature)
+
+    constraints = []
+    for constraint in domain.constraints:
```
jduerholt, line 47:
```suggestion
    for constraint in domain.constraints.get():
```

bofire/utils/entmoot.py
```diff
@@ -0,0 +1,151 @@
...
+    model_pyo = problem_config.get_pyomo_model_core()
+    model_pyo.problem_constraints = pyo.ConstraintList()
+    entconstr.ConstraintList(constraints).apply_pyomo_constraints(
+        model_pyo, problem_config.feat_list, model_pyo.problem_constraints
+    )
+
+    return problem_config, model_pyo
+
+
+def _bofire_feat_to_entmoot(problem_config: ProblemConfig, feature: AnyInput) -> None:
```
jduerholt, line 60:
```suggestion
def _bofire_feat_to_entmoot(problem_config: ProblemConfig, feature: Union[ContrinuousInput, CategoricalInput, DiscreteInput]) -> None:
```

bofire/utils/entmoot.py
```diff
@@ -0,0 +1,151 @@
...
+    elif isinstance(feature.objective, MaximizeObjective):
+        problem_config.add_max_objective(name=feature.key)
+
+    else:
+        raise NotImplementedError(f"Did not recognise output {feature}")
+
+
+def _bofire_constraint_to_entmoot(
+    problem_config: ProblemConfig,
+    constraint: AnyConstraint,
```
jduerholt, line 117:
```suggestion
    constraint: Union[LinearEqualityConstraint, LinearInequalityConstraint, NChooseKConstraint],
```

bofire/utils/entmoot.py
```diff
@@ -0,0 +1,151 @@
...
+
+    if isinstance(constraint, LinearEqualityConstraint):
+        ent_constraint = entconstr.LinearEqualityConstraint(
+            feature_keys=constraint.features,
+            coefficients=constraint.coefficients,
+            rhs=constraint.rhs,
+        )
+
+    elif isinstance(constraint, LinearInequalityConstraint):
+        ent_constraint = entconstr.LinearEqualityConstraint(
```
jduerholt, line 134:
```suggestion
        ent_constraint = entconstr.LinearInequalityConstraint(
```

I am not sure, but this sounds more reasonable for me ...

setup.py
```diff
@@ -43,6 +43,7 @@
     extras_require={
         "optimization": [
             "botorch>=0.10.0",
+            "entmoot>=2.0",
```
jduerholt, line 46: Let us create an extra `extra` for entmoot, or what do you think? @bertiqwerty? @R-M-Lee 

Perhaps we should do it also for botorch ...

tests/bofire/strategies/test_enting.py
```diff
@@ -0,0 +1,153 @@
...
+import pytest
+from pydantic.utils import deep_update
+
+import bofire.data_models.strategies.api as data_models
+from bofire.benchmarks.api import Hartmann
+from bofire.data_models.features.api import Input
+from bofire.strategies.api import EntingStrategy
+from tests.bofire.strategies.test_base import domains
+
+VALID_ENTING_STRATEGY_SPEC = {
```
jduerholt, line 13: This structure of testing for the strategies and its data models is deprecated. Just add valid and invalid ones in `tests/bofire/data_models/strategies.py`. Then they will be automatically tested for serialization and deserialization and for throwing the correct error. For the invalid ones: only test stuff which makes really sense and not basic pydantic functionality.

The import in `test_all.py` is then also obsolote. This has to be completely tiedied up.

bofire/utils/entmoot.py
```diff
@@ -0,0 +1,151 @@
+from typing import Tuple
```
jduerholt, line 1: One can also think about putting this all into `strategies/predictives/enting.py` as these utilities will only used by this one strategy.

What do you think?

jduerholt (changes requested):
Hi @TobyBoyne,

thank you very much! This looks already quite good. I let quite a lot of (minor) inline comments/questions and suggestions.

Missing functionality would be for me only handling of pending candidates. Via `strategy.add_candidates` you can add so called pending candidates, which are candidates that are suggested by the strategy but not yet evaluated but should be taken into account for the next proposal to not propose the same again. In your case you would do the same as you do in `add_candidate_as_experiment` when you want to create proposal with more than one candidate.. So it would be good to unite this functionality.

Best,

Johannes

cc: @bertiqwerty @R-M-Lee

bofire/data_models/strategies/predictives/enting.py
```diff
@@ -0,0 +1,89 @@
...
+    MaximizeObjective,
+    MinimizeObjective,
+    Objective,
+)
+from bofire.data_models.strategies.predictives.predictive import PredictiveStrategy
+
+
+class UncParams(BaseModel):
+    beta: PositiveFloat = 1.96
+    bound_coeff: float = 0.5
```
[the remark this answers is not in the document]
TobyBoyne: I don't think it can be negative - I'll change this

bofire/data_models/strategies/predictives/enting.py
```diff
@@ -0,0 +1,89 @@
...
+    acq_sense: Literal["exploration", "penalty"] = "exploration"
+    dist_trafo: Literal["normal", "standard"] = "normal"
+    dist_metric: Literal["euclidean_squared", "l1", "l2"] = "euclidean_squared"
+    cat_metric: Literal["overlap", "of", "goodall4"] = "overlap"
+
+
+class TrainParams(BaseModel):
+    # lightgbm training hyperparameters
+    objective: str = "regression"
```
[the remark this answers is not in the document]
TobyBoyne: Yep, I can't see why (when using ENTMOOT) you would want to use anything else

bofire/data_models/strategies/predictives/enting.py
```diff
@@ -0,0 +1,89 @@
...
+
+    unc_params: "UncParams" = Field(default_factory=lambda: UncParams())
+    tree_train_params: "TreeTrainParams" = Field(
+        default_factory=lambda: TreeTrainParams()
+    )
+
+
+class EntingStrategy(PredictiveStrategy):
+    type: Literal["EntingStrategy"] = "EntingStrategy"
+    enting_params: EntingParams = Field(default_factory=lambda: EntingParams())
```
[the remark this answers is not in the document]
TobyBoyne: (See other comment)

bofire/data_models/strategies/predictives/enting.py
```diff
@@ -0,0 +1,89 @@
...
+    solver_params: Dict[str, Any] = {}
+    learn_from_candidates_coeff: float = 10.0
```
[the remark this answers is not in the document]
TobyBoyne: Anything. I probably need to better document this attribute for clarity, it links to the batch proposal stuff.

setup.py
TobyBoyne: Hm after looking again I think another `extra` is probably better, since that's how it's done for `xgb`.

bofire/utils/entmoot.py
TobyBoyne: Yes that makes sense. The file ends up being ~300 lines which isn't so bad.

tests/bofire/strategies/test_enting.py
TobyBoyne: Hm, I think I might be misunderstanding this. The file you mentioned doesn't exist, and I'm modelling my tests after `tests/bofire/strategies/test_sobo.py` which still has the import in `test_all.py`. These tests also seem to be essentially testing basic pydantic functionality - could you please give an example where this isn't the case?

Could you also please explain how the automatic testing works? I think this will help me better understand how to write these tests, thank you.

bofire/data_models/strategies/predictives/enting.py
```diff
@@ -0,0 +1,89 @@
...
+    num_boost_round: PositiveInt = 100
+    max_depth: PositiveInt = 3
+    min_data_in_leaf: PositiveInt = 1
+    min_data_per_group: PositiveInt = 1
+    verbose: int = -1
+
+
+class TreeTrainParams(BaseModel):
+    train_params: "TrainParams" = Field(default_factory=lambda: TrainParams())
+    train_lib: Literal["lgbm"] = "lgbm"
```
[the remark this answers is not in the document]
jduerholt: Hmm, ok. I understand. Then proceed as you wish. @bertiqwerty @R-M-Lee what do you think?

bofire/utils/entmoot.py
jduerholt: Ok, then let's move it just to one file for now.

bofire/data_models/strategies/predictives/enting.py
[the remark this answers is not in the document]
bertiqwerty: I would make the life for BoFire user as easy as possible. Creating the entmoot-lib-parameters where needed in the corresponding code during instantiation is fine, even though their structure is different. We do not need to replicate this here, if there is no benefit for the BoFire user.

tests/bofire/strategies/test_enting.py
jduerholt: You are right, it is still used in `test_sobo` and so forth, but this needs to be tidied up. The correct link is `tests/bofire/data_models/specs/strategies.py`. There valid and invalid specs can be registered and the valid ones are then tested in `tests/bofire/data_models/serialization` for correct serialization and deserialization and the invalid ones are tested for invalidity in `tests/bofire/data_models/test_invalid.py`. This is achieved by using fixtures which are defined in `tests/bofire/conftest.py` (no change on the fixtures is necessary from your side, you only need to add at least one valid configureation in `tests/bofire/data_models/specs/strategies.py` ).

The plan is to make use of `add_invalid` only in scenarios where one has custom validators. An example would be for example here for the `TaskInput` as we are testing here that the custom validator works: https://github.com/experimental-design/bofire/blob/ccb0bc2998f35ede3118b490e0ce387244898305/tests/bofire/data_models/specs/features.py#L173

I think in your case no `add_invalid` is needed, so just add one valid configuration via `add_valid`. You have to provide there all values for the fields, also the ones which have default values.

bofire/data_models/strategies/predictives/enting.py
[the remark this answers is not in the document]
jduerholt: So you would also argue for a flat data structure in the strategy and the corresponding data structures for `entmoot` are generated on the fly based on the flat structure and passed to the corresponding function calls? This would be also my preferred solution, as this makes it easy for the bofire user. Thing like the parameter `regression` for LGBM can then also removed, as this is generated under the hood before passed to LGBM. 

I think also that it can tested quite well. Just create a helper method that takes the flat arguments in and that returns the data structures needed for entmoot, which then can be easilty tested.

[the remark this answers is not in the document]
TobyBoyne: I think this all makes sense. I have created two methods in the `datamodels.EntingStrategy` class that dump the flat structure into the nested structure required for ENTMOOT. This also means that I have removed the unecessary parameters. I hope this solution is what you were thinking!

tests/bofire/strategies/test_enting.py
TobyBoyne: Thank you for explaining, this all makes sense.

I have implemented it, and have run `test_strategy_should_be_(de)serializable` and they're all working. This testing set up is very cool, I like it!

TobyBoyne:
I think this should cover all of the requested changes - please let me know if I missed anything.

Regarding the `candidates`, I think my solution should address this. It is slightly clunky, in my opinion. For example, if I already have generated 9 candidates, then I want to generate a 10th, I need to retrain my model on those 9 candidates sequentially, since the batch process for ENTMOOT is sequential. I can't find a neat way to store the prediction of a candidate in the current class design, and I have to discard the model as it fits on the fantasies (https://github.com/TobyBoyne/bofire/blob/d8eb2b6daecf5bc62bbb432d47f2c551ad316f39/bofire/strategies/predictives/enting.py#L281) because otherwise, the fantasy observations will be seen as a "real experiment" which is not the case.

So my solution ends up slow if you are doing multiple sequential calls to `.ask()`, but I think it's the most correct. You can see this test (https://github.com/TobyBoyne/bofire/blob/d8eb2b6daecf5bc62bbb432d47f2c551ad316f39/tests/bofire/strategies/test_enting.py#L124) for what I think the desired behaviour should look like.

jduerholt (changes requested):
Hi @TobyBoyne,

looks very good to me, just a few small comments regarding test pipelines and handling of the pending ones.

Best,

Johannes

bofire/strategies/predictives/enting.py
```diff
@@ -0,0 +1,317 @@
...
+            candidate_count (PositiveInt, optional): Number of candidates to be generated. Defaults to 1.
+
+        Returns:
+            pd.DataFrame: DataFrame with a candidates.
+        """
+        # First, fit the model on fantasies generated for any pending candidates
+        # This ensures that new points are far from pending candidates
+        real_experiments = self.experiments
+        if self.candidates is not None:
+            for i in range(len(self.candidates)):
```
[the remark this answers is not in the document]
TobyBoyne: It is possible, however it wouldn't be 'correct', I believe. Let's say we have candidates $(x_1, ..., x_5)$. When we initially generated these, we first generated $x_1$, then created a 'fantasy observation' of this point, and used that to generate $x_2$, and so on. This means that $x_5$ is conditional on the model **fitted to the fantasy observations of $(x_1, ..., x_4)$.** If we generated the fantasies at the same time, then the predictions would be different, which means $x_5$ may no longer be optimal as the fantasy observations of the other points would be different.

Ultimately this problem arises because ENTMOOT can only produce single, globally optimal candidates. There is no support for batch proposals in ENTMOOT.

If doesn't make sense, or you think the difference is too subtle, please let me know - happy to carry on this discussion!

bofire/strategies/predictives/enting.py
```diff
@@ -0,0 +1,317 @@
...
+                # iterate using indices so that each `candidate` is a DataFrame
+                candidate = self.candidates[i : i + 1]
+                # add prediction from model
+                preds = self.predict(candidate)
+                candidate = pd.concat((candidate, preds), axis=1)
+                as_experiment = self._fantasy_as_experiment(candidate)
+                self.tell(as_experiment)
```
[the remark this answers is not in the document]
TobyBoyne: Hm, good point. My original thought was that if I used `_fit`, then I wouldn't get any validation of the output of the prediction (`Domain.validate_outputs` is called in `{set/add}_experiments`). But if `preprocess_experiments_all_valid_outputs` is sufficient in `_fit`, or if I just call this method separately, then I would agree with you

tests/bofire/strategies/test_enting.py
```diff
@@ -0,0 +1,234 @@
...
+
+    input_values = proposal[benchmark.domain.get_feature_keys(Input)]
+    assert (input_values != 0).sum().sum() <= allowed_k
+
+
+@pytest.mark.slow
+def test_propose_optimal_point(common_args):
+    # regression test, ensure that a good point is proposed
+    np.random.seed(42)
+    random.seed(42)
```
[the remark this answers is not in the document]
TobyBoyne: Currently not, I will do that! However it's worth noting that passing the seed to `entmoot` does just call `np.random.seed(seed); random.seed(seed)` in the entmoot codebase, but I suppose it's worth doing anyway.

jduerholt:
Hi @TobyBoyne,

I will do a proper review this evening. Thank you already!

Best,

Johannes

bofire/strategies/predictives/enting.py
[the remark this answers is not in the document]
jduerholt: In your case here, just using `_fit` should be fine.

tests/bofire/strategies/test_enting.py
```diff
@@ -0,0 +1,234 @@
...
+    strategy = EntingStrategy(data_model)
+
+    strategy.tell(experiments)
+    proposal = strategy.ask(1)
+
+    input_values = proposal[benchmark.domain.get_feature_keys(Input)]
+    assert (input_values != 0).sum().sum() <= allowed_k
+
+
+@pytest.mark.slow
```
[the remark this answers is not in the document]
jduerholt: This makes sense, could you check if gurobi is installed in somehow the same way that you check that entmoot is installed and then add another `skipif`. This would be a bit cleaner. If not possible, then we keep it.

tests/bofire/strategies/test_enting.py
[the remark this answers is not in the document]
jduerholt: I would prefer passing the seed explicitly to entmoot via calling `strategy._get_seed()` and passing the int then to entmoot. As a general note, it would be also better in entmoot to not globally set the seeds beut using numpys `rng` functionality.

jduerholt (changes requested):
We are converging! Biggest obstacle is from my perspective currently some problem with the `entmoot` installation in the test pipeline. But this looks for me like an entmoot problem and not a bofire problem (see my comments).

Best,

Johannes

bofire/strategies/predictives/enting.py
```diff
@@ -0,0 +1,360 @@
+import warnings
+from typing import List, Optional, Tuple
+
+import numpy as np
+import pandas as pd
+import pyomo.environ as pyo
```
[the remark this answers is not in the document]
TobyBoyne: I have added these dependencies in `setup.py` in BoFire. However, these dependencies definitely do need to be fixed on entmoot's side - for example, `lightgbm` is not installed despite being required, and it should probably just install `pyomo` anyway.

bofire/strategies/predictives/enting.py
```diff
@@ -0,0 +1,370 @@
...
+    ContinuousInput,
+    DiscreteInput,
+)
+from bofire.data_models.objectives.api import MaximizeObjective, MinimizeObjective
+from bofire.strategies.predictives.predictive import PredictiveStrategy
+
+
+def domain_to_problem_config(
+    domain: Domain, seed: Optional[int] = None
+) -> Tuple[ProblemConfig, pyo.ConcreteModel]:
```
jduerholt, line 39: I think having this import specific signature is a problem. Putting it in `""` should solve it.

```suggestion
) -> Tuple["ProblemConfig", "pyo.ConcreteModel"]:
```

jduerholt (commented):
Looks overall very good to me. But somehow tests are still failing. One sees in the test logs that entmoot gets installed but somehow the suff in the `try except` block fails, so that nothing gets imported. No idea why.

I will try to set the permissions for you in a way that the test pipeline is directly triggered, without needing us for this. So you can iterate faster.

bofire/strategies/predictives/enting.py
TobyBoyne: I will change this for all signatures.

TobyBoyne:
Two sources of issues that I've found so far:

Entmoot requires `gurobipy`, even if it isn't being used. This is an issue with Entmoot (https://github.com/cog-imperial/entmoot/issues/32), but I've added it to `setup.py` for now

pyright isn't happy with `domain_to_problem_config`. This is because the functions only accept a subset of the possible feature types. To me, the solution would be to make the type hinting in `_bofire_feat_to_entmoot` less specific. However, this undoes a change that @jduerholt recommended earlier in this PR, so I just wanted to flag it so that you are aware I'm reverting it!

TobyBoyne:
I've made some progress. There are quite a few `# type: ignore`s in `enting.py` which may not all be necessary, but I used to get around pyright and the above issue about the return types of `output.get()` and `input.get()`.

However, I'm still getting issues with running the pipeline. On my machine, if I remove the license file, the tests that require Gurobi are skipped. However, the pipeline seems to still be attempting to run these, and then having issues. Any ideas?

(Note that `pyomo.common.errors.ApplicationError: Solver (gurobi) did not exit normally` error is raised when there is no license file).

jduerholt:
Using `type:ignore` when using the `get` methods is fine. Pyright is not able to work with this dynamic behavior. Currently, I cannot start the pipeline, can you resolve the merge conflicts? They came in due to a change by Behrang ;) I will then have a look again at the pipeline and the tests.

jduerholt:
Found the results from the latest pipeline execution here: https://github.com/TobyBoyne/bofire/actions/runs/8236200085/job/22522063543, and I also tried your catch for gurobi locally, (i just installed it via pip) without having a liscense and got this:

image

Any ideas?

jduerholt:
Here it is stated that it comes with a trial liscense that can be used only for small problems: https://pypi.org/project/gurobipy/

TobyBoyne:
Hi @jduerholt, I've added this in and solved merge conflicts. Looks like the tests are all passing...?

Hopefully this should be good to go! Thank you for all of the reviews! Looking forward to the next PR :)

bertiqwerty:
Thanks a lot, Toby and Johannes!

R-M-Lee:
@jduerholt, @TobyBoyne  Thanks to both of you for getting this through. Are we happy with the treatment of the optional entmoot dependencies now or was this more of a short-term fix and we still have some open to-dos?

PR #279, opened 2023-09-07
pydantic 2  - migration

bertiqwerty:
The goal of this MR is to upgrade to Pydantic 2. For the initial commit I run `bump-pydantic`  and changed some things manually. There are still some open issues, though.


bofire/data_models/strategies/predictives/botorch.py
```diff
@@ -69,13 +69,13 @@ def validate_num_raw_samples(cls, v):
...
 
-    @model_validator(mode="after")
+    @model_validator(mode="before")
     @classmethod
     def update_surrogate_specs_for_domain(cls, values):
         """Ensures that a prediction model is specified for each output feature"""
         values["surrogate_specs"] = BotorchStrategy._generate_surrogate_specs(
             values["domain"],
-            values["surrogate_specs"],
+            values["surrogate_specs"] if "surrogate_specs" in values else None,
```
bertiqwerty, line 78: @jduerholt, is this correct?

bofire/data_models/strategies/predictives/botorch.py
```diff
@@ -96,11 +96,14 @@ def update_surrogate_specs_for_domain(cls, values):
...
 
-    @model_validator(mode="after")
+    @model_validator(mode="before")
     @classmethod
     def validate_outlier_detection_specs_for_domain(cls, values):
         """Ensures that a outlier_detection model is specified for each output feature"""
-        if values["outlier_detection_specs"] is not None:
+        if (
+            "outlier_detection_specs" in values
+            and values["outlier_detection_specs"] is not None
```
bertiqwerty, line 105: @jduerholt, here again.

bofire/data_models/strategies/predictives/botorch.py
jduerholt: Hmm, I know that it definitely works, because this is always used when one does not provide any surrogates specs. I think, that, at least in the former pydantic versions, if not provided it will be None.

bofire/data_models/strategies/predictives/botorch.py
jduerholt: I only can say that it has worked :D

bertiqwerty: My question was if the change is correct or if it changes what happens :D

jduerholt:
Todo list:
- review all model validators, we should switch everywhere to mode "after", this is more failsafe
- add validators for inputs and outputs objects to check for unique keys
- intotroduce commone type for unique str tuples

jduerholt:
There are some bugs in pydantic2:
- Order of AnyFeature affects the correct deserialization
- cannot deal with `update_hyperparams` as abstract method ...

Things to consider:
- Look at `order_id` for features
- make one place for custom types

jduerholt:
@bertiqwerty: Only failing tests are now due to failure in the DOE module since on the 25th of December a new formulaic version was released and this breaks some of our tests.

jduerholt:
Hi @bertiqwerty,

tests are now passing, and I think the PR is ready for review.

What do you think, who should review it? 

Issues that are mentioned above which are still open are:
- create a central place for our custom types like `bofire/data_models/types.py`
- on several places in the code we use lists that constrain strings which have to be unique (examples are the `features` attribute from constraints, the categories attribute from categorical features etc.). In my opinion we should switch them all to  tuples and create a custom type for this.

But it could be better to do this in a seperate PR. 

What do you think?

Best,

Johannes

bertiqwerty:
Hi @jduerholt , 

Thanks a lot! The person for review would probably be Benny. I will also have a look. Having a central `types.py` makes sense to me. We should do this in a separate PR.

bofire/data_models/constraints/nchoosek.py
```diff
@@ -75,10 +73,16 @@ def relu(x):
         min_count_violation = np.zeros(experiments_tensor.shape[0])
 
         if self.max_count != len(self.features):
-            max_count_violation = relu(-1 * narrow_gaussian(x=experiments_tensor[..., indices]).sum(axis=-1) + (len(self.features) - self.max_count))  # type: ignore
+            max_count_violation = relu(
+                -1 * narrow_gaussian(x=experiments_tensor[..., indices]).sum(axis=-1)
+                + (len(self.features) - self.max_count)
+            )  # type: ignore
```
bertiqwerty, line 79: Does this `# type ignore` do anything?

bofire/data_models/constraints/nchoosek.py
```diff
@@ -75,10 +73,16 @@ def relu(x):
...
 
         if self.min_count > 0:
-            min_count_violation = relu(narrow_gaussian(x=experiments_tensor[..., indices]).sum(axis=-1) - (len(self.features) - self.min_count))  # type: ignore
+            min_count_violation = relu(
+                narrow_gaussian(x=experiments_tensor[..., indices]).sum(axis=-1)
+                - (len(self.features) - self.min_count)
+            )  # type: ignore
```
bertiqwerty, line 85: See above

bofire/data_models/strategies/predictives/botorch.py
```diff
@@ -53,61 +55,65 @@ def is_constraint_implemented(cls, my_type: Type[Constraint]) -> bool:
...
+            for m in self.surrogate_specs.surrogates:
                 if isinstance(m, MixedSingleTaskGPSurrogate):
                     raise ValueError(
                         "Categorical method FREE not compatible with a a MixedSingleTaskGPModel."
                     )
-        #  we also check that if a categorical with descriptor method is used as one hot encoded the same method is
+        # we also check that if a categorical with descriptor method is used as one hot encoded the same method is
         # used for the descriptor as for the categoricals
-        for m in values["surrogate_specs"].surrogates:
+        # descriptor_method = (
```
bertiqwerty, line 86: Remove uncommented code

bofire/data_models/strategies/predictives/botorch.py
```diff
@@ -120,28 +126,30 @@ def _generate_surrogate_specs(
...
             else:
                 _surrogate_specs.append(
                     SingleTaskGPSurrogate(
                         inputs=domain.inputs,
-                        outputs=Outputs(features=[domain.outputs.get_by_key(output_feature)]),  # type: ignore
+                        outputs=Outputs(
+                            features=[
+                                domain.outputs.get_by_key(output_feature)  # type:ignore
+                            ]
+                        ),  # type: ignore
```
bertiqwerty, line 150: When I understood this correctly, `# type: ignore` correctly, it operates on the line level. We could at some point activate `reportUnnecessaryTypeIgnoreComment` in Pyright's options.
See https://github.com/microsoft/pyright/blob/main/docs/comments.md#line-level-diagnostic-suppression

bofire/data_models/surrogates/botorch_surrogates.py
```diff
@@ -83,7 +83,8 @@ def _check_compability(self, inputs: Inputs, outputs: Outputs):
         if len(used_feature_keys) != len(inputs):
             raise ValueError("Unused features are present.")
 
-    @validator("surrogates")
+    @field_validator("surrogates")
+    @classmethod
     def validate_surrogates(cls, v, values):
         # validate that all surrogates are single output surrogates
         # TODO: this restriction has to be removed at some point
```
bertiqwerty, line 90: Make an issue out of the TODO?

bofire/data_models/surrogates/fully_bayesian.py
```diff
@@ -1,18 +1,20 @@
...
 
 
 class SaasSingleTaskGPSurrogate(TrainableBotorchSurrogate):
     type: Literal["SaasSingleTaskGPSurrogate"] = "SaasSingleTaskGPSurrogate"
-    warmup_steps: conint(ge=1) = 256  # type: ignore
-    num_samples: conint(ge=1) = 128  # type: ignore
-    thinning: conint(ge=1) = 16  # type: ignore
+    warmup_steps: Annotated[int, Field(ge=1)] = 256  # type: ignore
+    num_samples: Annotated[int, Field(ge=1)] = 128  # type: ignore
+    thinning: Annotated[int, Field(ge=1)] = 16  # type: ignore
```
bertiqwerty, line 13: Maybe we don't need these `#type ignore`s due to `Annotated`?

bofire/surrogates/diagnostics.py
```diff
@@ -485,18 +486,24 @@ def validate_shapes(cls, values):
                 )
         return values
 
+    # TODO[pydantic]: We couldn't refactor the `validator`, please replace it by `field_validator` manually.
```
bertiqwerty, line 489: Does `TODO[pydantic]` have a special meaning? And who is the one to replace stuff manually?

bertiqwerty (commented):
Thanks Johannes. I cannot approve the PR because Github says authors cannot approve their own PRs. You can probably approve and merge it?

jduerholt:
Hi @bertiqwerty, 

thx! Yes, I will go through it and then approve and merge. For me it works.

The idea with the `reportUnnecessaryTypeIgnoreComment` is a good one! Maybe we should also switch to a more recent pyright? We are still on version `1.1.305`.

Before than merging it in, should we make a last pydantic v1 release?

Best,

Johannes

bertiqwerty:
Hi Johannes, 

I think we don't need another release with Pydantic 1 but I also don't have an objection. So I am fine either way.

Updating Pyright is a good idea. This often leads to some issues. So we could create a MR for this. I can look into this.

jduerholt:
Ok, then we do not create a Pydantic 1 release. Would be nice if you can take care for the pyright update. But we should do it after merging in this PR. My plan is to have it in the latest tomorrow. Note that I moved the pyright config some time ago into the `pyproject.toml`, so the new flag for showing the unnecessary `ignore` statements should also go there, I think:

https://github.com/experimental-design/bofire/blob/33a2053e80c371791cb27364c442b83f866b1c09/pyproject.toml#L34

bofire/data_models/constraints/nchoosek.py
jduerholt: Good catch!

bofire/data_models/constraints/nchoosek.py
jduerholt: Good catch.

bofire/data_models/surrogates/botorch_surrogates.py
jduerholt: At some point, currently I am too lazy :D

bofire/surrogates/diagnostics.py
jduerholt: Very good catch, i fixed it.

jduerholt (approved):

jduerholt:
@bertiqwerty: it is now merged, you can now have a look at the pyright update ;)

bertiqwerty:
I am in favor of making a new release with the new Pydantic version.

bertiqwerty:
I can.

PR #284, opened 2023-09-13
Add aggregation modes to trainable surrogate data models

jduerholt:
This PR adds the API for defining aggregated features in the future. Note that it is currently only supported in the data models but will not be used in the functional surrogates.

They can be specified like this for all trainable surrogates:

```
'aggregations': [{'type': 'SumAggregation',
   'features': ['x_1', 'x_2'],
   'keep_features': False},
  {'type': 'MeanAggregation',
   'features': ['x_1', 'x_2'],
   'keep_features': False}],
```


simonsung06 (approved):
Looking forward the full implementation

Thanks

PR #285, opened 2023-09-13
Create surrogate for entmoot strategy

TobyBoyne:
Surrogate for Entmoot as discussed in #278.

Note that the surrogate is not called LGBM, since this interfaces with the Enting model, which contains mean/uncertainty models. Each of those models is not necessarily an LGBM model (although this is currently the only one implemented in ENTMOOT). Therefore I feel that this being an EntingSurrogate is the right level of abstraction (and then refactoring the strategy from `EntingStrategy` to `EntmootStrategy`).

### DIFF
--- a/bofire/data_models/surrogates/enting.py
+++ b/bofire/data_models/surrogates/enting.py
@@ -1,0 +1,36 @@
+from typing import Literal, Optional
+
+from pydantic import Field, validator
+from typing_extensions import Annotated
+
+from bofire.data_models.enum import CategoricalEncodingEnum
+from bofire.data_models.features.api import (
+    CategoricalDescriptorInput,
+    CategoricalInput,
+    NumericalInput,
+)
+from bofire.data_models.surrogates.surrogate import Surrogate
+from bofire.data_models.surrogates.trainable import TrainableSurrogate
+
+
+class EntingSurrogate(Surrogate, TrainableSurrogate):
+    type: Literal["EntingSurrogate"] = "EntingSurrogate"
+    train_lib: Literal["lgbm"] = "lgbm"
+    # mean model parameters
+    objective: str = "regression"
+    metric: str = "rmse"
+    boosting: str = "gbdt"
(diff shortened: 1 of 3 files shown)


bofire/data_models/surrogates/enting.py
```diff
@@ -0,0 +1,27 @@
+from typing import Literal
+
+from pydantic import Field
+from typing_extensions import Annotated
+
+from bofire.data_models.surrogates.surrogate import Surrogate
+from bofire.data_models.surrogates.trainable import TrainableSurrogate
+
+
+class EntingSurrogate(Surrogate, TrainableSurrogate):
```
jduerholt, line 10: If you want to do it like this, I would call it `LGBMEntingSurrogate`, and if you have xgb support at some point you could introduce an `XGBEntingSurrogate`. This has the advantage that the names of hyperparameters do not have to be the same but different, or if one hyperparam is only available in one model you only provide it in this model.

jduerholt: The other option would be to have a `LGBMSurrogate` which just includes the mean model and implement `LGBMEntingSurrogate` as inherited class from `LGBMSurrogate` and add the uncertainty parameters there. But I know that in the current `Entmoot` model it is setup a bit different, so you would have in this case two different `_fit` methods. The first just uses LGBM and the second one ENTMOOT. 

Thinking this further one could even build it up by having an `LGBMSurrogate` and `XGBSurrogate` class which are ENTMOOT agnostic and all care for the mean model and an `EntingUncertaintySurrogate` which adds the uncertainty part to the model. Then one could build up the `LGBMEntingSurrogate` by inheritance from `LGBMSurrogate` and `EntingUncertaintySurrogate`. This would be the super object oriented solution, but as these kind of structures are currently not supported by ENTMOOT, I think its is ok to just go with the flat `LGBMEntingSurrogate`.

What do you think?

jduerholt (commented):
I added some comments ;)

PR #287, opened 2023-09-15
Implement possibility to run hyperparameter opts in the strategy

jduerholt:
This PR adds the possibility to run a hyperparameter optimization within a BO step to ensure to have always the most accurate model.


jduerholt:
> This PR adds the possibility to run a hyperparameter optimization within a BO step to ensure to have always the most accurate model.

PR #288, opened 2023-09-18
Set default value for bounds in ContinuousInput Feature

jduerholt:
This PR implement default bounds for the `ContinuousInput` feature of `(0,1)`, which allows the user to just type:
```
ContinuousInput(key="input_1")
```
instead of
```
ContinuousInput(key="input_1", bounds = (0,1))
```
which is especially useful in situations where using features of BoFire where bounds are not needed. Less typing is then needed ;)

### DIFF
--- a/bofire/data_models/features/continuous.py
+++ b/bofire/data_models/features/continuous.py
@@ -20,7 +20,7 @@ class ContinuousInput(NumericalInput):
     type: Literal["ContinuousInput"] = "ContinuousInput"
     order_id: ClassVar[int] = 1
 
-    bounds: Tuple[float, float]
+    bounds: Tuple[float, float] = Field(default_factory=lambda: (0.0, 1.0))
     stepsize: Optional[float] = None
 
     @property


bertiqwerty:
What are the situations where you don't need bounds?

simonsung06:
A comment from my point of view about this change.

The bounds for `ContinuousInput` features are considered when the default Normalize scalar for SingleTaskGP is used (to the best of understanding [based on this function](https://github.com/experimental-design/bofire/blob/1e8306f4ad1a7a392264b405fac733292ebd74d0/bofire/data_models/features/numerical.py#L131)). This means bounds could affect the model training. Therefore, if providing bounds became less visible and not required, and the values for the `ContinuousInput` feature in the provided experiments DataFrame are between 0 and 1, then the arbitrary (0, 1) bounds would be used as the lower and upper bound in the Normalize input scalar, which may not be intended for a user who  doesn't provide bounds.

This is just a slight risk that I think is worth pointing out. But fine with me if you want to keep it and we have the users just be more aware of how they are using BoFire.

jduerholt:
@simonsung06 : this is a very good point! I will retract this PR.

PR #289, opened 2023-09-20
Feature/experiment validation

jduerholt:
This PR adds the possibility to pass as `experiments` dataframe, a frame in which the `dtype`s are not yet correct (for example all floats as strings) and still to validate/convert it into the desired format depending on the used feature types.


jduerholt:
Note that we have some overlaps between CategoricalInput/ContinuousInput and CategoricalOutput/ContinuousOutput. This should be resolved in a seperate PR by adding some middle layer classes to prevent code dublication.

PR #291, opened 2023-09-21
Refactor the candidates/experiments validators.

jduerholt:
This PR refactors the existing candidates and experiments validators.


simonsung06 (approved):
looks good to me

PR #295, opened 2023-10-02
Fix for categorical features that are fixed in fully categorical BO

simonsung06:
When all input features in an optimization are categorical/discrete, if only one category is allowed for the categorical features, then it will break. This is because the fixed feature is not included when `self.domain.inputs.get_categorical_combinations()` is used to generate the `choices` DataFrame. This results in the subsequent `merged` DataFrame having a column of NaN's for the fixed feature

This PR will fix this issue by adding in the fixed feature into `choices` before generating `merged`.

I have not included a test yet for this because I am not sure on where to place it in the folder. @jduerholt Where is the best place for this?

### DIFF
--- a/bofire/strategies/predictives/botorch.py
+++ b/bofire/strategies/predictives/botorch.py
@@ -296,6 +296,9 @@ class BotorchStrategy(PredictiveStrategy):
                     for combi in self.domain.inputs.get_categorical_combinations()
                 ]
             )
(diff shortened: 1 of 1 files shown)


jduerholt (approved):
Very good catch, thanks for fixing it. Testing is rather difficult for this, so I would merge it without a specific test for it.

PR #296, opened 2023-10-02
mapping happens outside of runner

bertiqwerty:
Otherwise, strategies implemented outside of BoFire cannot use the runner to be evaluated and crappy strategies will become part of BoFire since they could not be evaluated beforehand 😉


jduerholt:
@bertiqwerty: I will have a look over the course of the week (hopefully tmr). Thanks for the PR.

jduerholt (changes requested):
It looks good to me, I find it less elegant than the original version, as more code is needed for the user for standard tasks (before it was possible to just give the data model class as this was also a valid factory) but now it is more versatile. I think there is still a notebook in which it is not yet corrected.

bertiqwerty:
> It looks good to me, I find it less elegant than the original version, as more code is needed for the user for standard tasks (before it was possible to just give the data model class as this was also a valid factory) but now it is more versatile. I think there is still a notebook in which it is not yet corrected.

Thanks @jduerholt. I think, I fixed them all and I got rid of some type-ignores 🥳.  About the less-elegant-thing, I tend to disagree. I'd say it is less convenient. But it is more consistent with the way Bofire-users instantiate strategies. Further, you can now use it with strategies implemented outside of Bofire, e.g., during early experimentation or if you have a project-specific edge-case. These strategies could still inherit from `Strategy`. However, these outside-strategies are not registered in the mapper.

R-M-Lee (approved):

jduerholt (approved):

PR #297, opened 2023-10-03
Classification surrogates

gmancino:
# Add Classification Models for Surrogates

Adding new surrogates to allow for classification of output values (e.g. 'unacceptable', 'acceptable', or 'ideal') to use for modeling unknown constraints. Concretely, if $g_{\theta}:\mathbb{R}^d\to[0,1]^c$ represents a function governed by learnable parameters $\theta$ which outputs a probability vector over $c$ potential classes (i.e. for input $x\in\mathbb{R}^d$, $g_{\theta}(x)^\top\mathbf{1}=1$ where $\mathbf{1}$ is the vector of all 1's) and we have acceptability criteria for the corresponding classes given by $a\in\{0,1\}^c$, we can compute the expected acceptability as scalar output via $g_{\theta}(x)^\top a\in[0,1]$ which can be passed in as a constrained objective function.

### Classification Models

We add a new objective function in 'bofire/data_models/objectives/categorical.py' which can be passed to outputs of type `CategoricalOutput`. These are instantiated with a list of probability scales (i.e. the acceptability criteria vector) via the `desirability` argument and inherit the categories from the corresponding output. Using this new objective:

1. We implement an MLP ensemble method to start ('bofire/data_models/surrogates/mlp.py') which outputs a $c$-dimensional probability vector for each datapoint
2. The predicted value (stored in `{key}_{class}_prob` of the predictions) is the argmax along this probability vector, while the objective value (stored in `{key}_{class}_{des}` of the predictions) is the inner-product of the probability vector with the acceptability criteria vector
    - This value is also computed in the `constrained_objective2botorch` function, which currently undergoes the inverse of the sigmoid transformation to maintain the value in the probability space
3. We pass in the objective value to BoTorch as a constraint


gmancino:
@jduerholt, please let me know if the **style** of the updates here are appropriate. I will build more models if the initial idea makes sense.

bofire/strategies/doe/utils_categorical_discrete.py
jduerholt, line 1: Why changes in this rather unrelated module?

bofire/data_models/features/categorical.py
```diff
@@ -360,37 +359,43 @@ class CategoricalOutput(Output):
     order_id: ClassVar[int] = 8
 
     categories: TCategoryVals
-    objective: Annotated[
-        List[Annotated[float, Field(type=float, ge=0, le=1)]], Field(min_items=2)
-    ]
+    objective: CategoricalObjective = Field(
```
jduerholt, line 362: I do not understand this change? Why the need to model the desirability of the individual categories by a an own class? Why not keeping the list (ideally transforming it to a tuple?)

bofire/utils/torch_tools.py
```diff
@@ -147,24 +148,50 @@ def constrained_objective2botorch(
...
+            ],
+            [1.0 / objective.steepness, 1.0 / objective.steepness],
+            idx + 1,
+        )
+    elif isinstance(objective, CategoricalObjective):
+        # The output of a categorical objective has final dim `c` where `c` is number of classes
+        return (
+            [
+                lambda Z: -1.0
+                * objective.w
```
jduerholt, line 186: Why `objective.w`?

bofire/data_models/features/categorical.py
```diff
@@ -360,37 +359,43 @@ class CategoricalOutput(Output):
...
-        if len(objective) != len(values["categories"]):
+        weights = objective.weights
+        if len(weights) != len(values["categories"]):
             raise ValueError("Length of objectives and categories do not match.")
-        for o in objective:
-            if o > 1:
-                raise ValueError("Objective values has to be smaller equal than 1.")
-            if o < 0:
-                raise ValueError("Objective values has to be larger equal than zero")
+        for w in weights:
```
jduerholt, line 388: If we go for this structure, this validation should go to the inside of the objective class as validator there, or we just do it via the annotated stuff. But then also there.

bofire/data_models/features/categorical.py
```diff
@@ -360,37 +359,43 @@ class CategoricalOutput(Output):
...
+            if w > 1:
+                raise ValueError("Objective weight has to be smaller equal than 1.")
+            if w < 0:
+                raise ValueError("Objective weight has to be larger equal than zero")
+        # Save the categories to the objective if they do not exist
+        objective.categories = (
```
jduerholt, line 394: Hmm, why is this needed? This is not a good data structure from a software engineering point of view.

bofire/surrogates/mlp_classifier.py
```diff
@@ -0,0 +1,220 @@
...
+            mlp.eval()
+
+    def forward(self, X: Tensor):
+        r"""Compute the model output at X.
+
+        Args:
+            X: A `batch_shape x n x d`-dim input tensor `X`.
+
+        Returns:
+            A `batch_shape x s x n x m x C`-dimensional output tensor where
```
jduerholt, line 98: and what is n and m?

bofire/surrogates/mlp_classifier.py
```diff
@@ -0,0 +1,220 @@
...
+            `s` is the size of the ensemble and `C` is the number of classes.
+        """
+        return torch.stack([torch.softmax(mlp(X), dim=-1) for mlp in self.mlps], dim=-3)
```
jduerholt, line 101: Why do we softmax again the result of the log_softmax?

jduerholt (commented):
Hi Gabe, this looks really impressive. I just added some intitial questions as I am not yet fully through this beast :D

bofire/utils/torch_tools.py
```diff
@@ -147,24 +148,50 @@ def constrained_objective2botorch(
         objective (BotorchConstrainedObjective): The objective that should be transformed.
 
     Returns:
-        Tuple[List[Callable[[Tensor], Tensor]], List[float]]: List of callables that can be used by botorch for setting up the constrained objective, and
-            list of the corresponding botorch eta values.
+        Tuple[List[Callable[[Tensor], Tensor]], List[float], int]: List of callables that can be used by botorch for setting up the constrained objective,
+            list of the corresponding botorch eta values, final index used by the method (to track for categorical variables)
```
jduerholt, line 152: For what is this index needed?

bofire/strategies/doe/utils_categorical_discrete.py
gmancino: Ruff was failing - something about using isinstance is better than comparing type

bofire/data_models/features/categorical.py
gmancino: The challenge is that we need to pass the `CategoricalOutput` type to the `constrained_objective2botorch` to be evaluated by BoTorch as a constraint. To me, this seemed like a more reasonable option as it did not require refactoring the `constrained_objective2botorch` to also look for lists (tuples really). In the end, the objective should produce a scalar of some kind and having the objective just as a list does not allow for such functionality. Maybe I am missunderstanding the question...

bofire/utils/torch_tools.py
gmancino: This is an error. I was testing things locally; it will be changed in the next revision :)

bofire/data_models/features/categorical.py
gmancino: Is this referring to the reassignment of the `objective.categories`? I can change this so that `CategoricalObjective` takes in both objective and categories arguments. Then the validator just makes sure the categories are the same. Does that seem better?

Initially, I thought the repetition of categories was redundant, but maybe it makes more sense?

bofire/surrogates/mlp_classifier.py
gmancino: Copy paste typo - I will fix these as it should just be `batch_size x n x C`

bofire/surrogates/mlp_classifier.py
gmancino: Error on my part - will change to `torch.stack([mlp(X).exp() for mlp in self.mlps], dim=-3)`

bofire/utils/torch_tools.py
gmancino: Since the MLP will output a vector of length `c`, we need to keep track of two items: 1. an index related to the current objective and 2. an index related to the current objective's output. For most other methods, if the output is 1 dimensional, then these two indexes are the same. For the any method that outputs more than one objective value, then eventually these two indexes will be misaligned. 

The output that eventually gets passed to `constrained_objective2botorch` is the output from model.posterior <- which is the result of the forward pass through all models. If one of these models outputs a vector of size > 1, then using solely the objective index results in a failure to capture the underlying model's output. Is this a problem with other methods too? Do I need to add a check for all models somewhere else to measure their output size?

bofire/data_models/features/categorical.py
jduerholt: Ok, makes sense. Then let's do it as proposed by you by an own objective class which holds a tuple of desirability values which are between 0 and 1.

bofire/data_models/features/categorical.py
jduerholt: I meant that the `CategoricalObjective` should not hold the categories, it should only hold the `desirabilities` or `weights` and the categories should just be attributes of the `CategoricalOutput.`

bofire/utils/torch_tools.py
jduerholt: Ah, I get it. This makes sense. Currently, I do not see any problem there, but I will think about it ;)

bofire/data_models/features/categorical.py
gmancino: I see - I have updated accordingly

gmancino:
@jduerholt thank you so much for your feedback! I think I have addressed all of your concerns, but please let me know if there are any additional errors. If it all looks good, I plan on implementing something like in https://github.com/pytorch/botorch/issues/640 for a categorical GP. This should cover our initial basis :)

bofire/utils/torch_tools.py
```diff
@@ -147,24 +148,49 @@ def constrained_objective2botorch(
...
+        )
+    elif isinstance(objective, CategoricalObjective):
+        # The output of a categorical objective has final dim `c` where `c` is number of classes
+        return (
+            [
+                lambda Z: -1.0
+                * (
+                    Z[..., idx : idx + len(objective.desirability)]
+                    * torch.tensor(objective.desirability).to(**tkwargs)
+                ).sum(-1)
```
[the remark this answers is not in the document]
jduerholt: But maybe, I overlooked a transformation somewhere.

bofire/data_models/features/categorical.py
gmancino: Just coming back to this; is there a reason the `CategoricalObjective` should not hold the categories? I think there is an inherit challenge with decoupling the `CategoricalOutput` and `CategoricalObjective`, but both functionalities serve different purposes in BoFire...

It might make sense to have the objective hold both the categories and the desireabilities, that way the CategoricalOutput can inherit those values and do any text processing/whatever else is done with Outputs and the Objective can handle the actual computations

bofire/data_models/features/categorical.py
```diff
@@ -401,9 +404,33 @@ def validate_experimental(self, values: pd.Series) -> pd.Series:
             )
         return values
 
+    def __str__(self) -> str:
+        return "CategoricalOutputFeature"
+
     def to_dict(self) -> Dict:
         """Returns the catergories and corresponding objective values as dictionary"""
-        return dict(zip(self.categories, self.objective))
+        return dict(zip(self.categories, self.objective.desirability))
```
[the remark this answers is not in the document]
gmancino: This would mean we set up functions within the objective which take in the categories and do some processing. I think it's still unclear to me why the categories cannot be directly passed to the objective upon instantiation

bofire/strategies/predictives/predictive.py
```diff
@@ -102,16 +103,47 @@ def predict(self, experiments: pd.DataFrame) -> pd.DataFrame:
             experiments=experiments, specs=self.input_preprocessing_specs
         )
         preds, stds = self._predict(transformed)
+        column_names = list(
+            itertools.chain(
+                *[
+                    [f"{feat.key}_pred"]
+                    if not isinstance(feat, CategoricalOutput)
+                    else [f"{feat.key}_pred_{cat}" for cat in feat.categories]
```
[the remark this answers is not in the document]
gmancino: This can be added. One concern is then the user cannot see the other probabilities. For example, it could be [0.51, 0.49, 0.0] as the predicted probabilities, but if they only see the highest value the information on the closeness of the first two classes is missed. I'm fine with either way, but just want to bring a scenario like this up

gmancino:
Hi @jduerholt, I have made the corresponding changes to this PR which we have discussed. Some tests are currently failing because I changed the `TCategoryVals` to be of type `Tuple[str, ...]`, which would make the categories immutable. Hence, in each hard coded location of categories (which are currently lists) the tests fail. I know we discussed using Tuples instead of Lists for this so it may be worth changing by hand unless you have some other opinion? There may be additional technical comments on the changes made which need to be addressed first ;)

bofire/data_models/features/categorical.py
```diff
@@ -14,14 +13,18 @@
...
+)
 
 
 class CategoricalInput(Input):
     """Base class for all categorical input features.
 
     Attributes:
-        categories (List[str]): Names of the categories.
-        allowed (List[bool]): List of bools indicating if a category is allowed within the optimization.
+        categories (Tuple[str]): Names of the categories.
```
jduerholt, line 26: Let's do this in a later PR with changing the types here from list to tuple, as mentioned above. It could then even make sense to have it not as tuple but as set.

bofire/data_models/features/categorical.py
```diff
@@ -363,39 +365,52 @@ class CategoricalOutput(Output):
     order_id: ClassVar[int] = 8
 
     categories: TCategoryVals
-    objective: Annotated[
-        List[Annotated[float, Field(type=float, ge=0, le=1)]], Field(min_items=2)
-    ]
+    objective: Optional[
+        Union[CategoricalObjective, ConstrainedCategoricalObjective]
```
jduerholt, line 369: Hmm, `CategoricalObjective` is an abstract Objective. We only allow here for the actual ones. You could introduce in `data_models/objectives/api.py` a type called `AnyCategoricalObjective`, which is for now just equal to ConstrainedCategoricalObjective.

bofire/data_models/features/categorical.py
```diff
@@ -363,39 +365,52 @@ class CategoricalOutput(Output):
...
+        Args:
+            categories (Union[List[str], Tuple[str]]): List or tuple of category names
+
+        Raises:
+            ValueError: when categories do not match objective categories
+
+        Returns:
+            Tuple[str]: Tuple of the categories
+        """
+        if objective.categories != tuple(values["categories"]):
```
jduerholt, line 402: We sould have a test for it within `tests/data_models/specs/features.py` using the `add_invalid` syntax.

bofire/data_models/features/categorical.py
```diff
@@ -363,39 +365,52 @@ class CategoricalOutput(Output):
...
+            raise ValueError("categories must match to objective categories")
         return objective
 
+    @classmethod
+    def from_objective(
```
jduerholt, line 407: We need a test for this in `tests/bofire/data_models/features/categorical.py`

bofire/data_models/features/categorical.py
```diff
@@ -404,9 +419,14 @@ def validate_experimental(self, values: pd.Series) -> pd.Series:
...
 
-    def to_dict(self) -> Dict:
-        """Returns the catergories and corresponding objective values as dictionary"""
-        return dict(zip(self.categories, self.objective))
-
     def __call__(self, values: pd.Series) -> pd.Series:
-        return values.map(self.to_dict()).astype(float)
+        if self.objective is None:
+            return pd.Series(
+                data=[np.nan for _ in range(len(values))],
```
jduerholt, line 425: we need a test for this method.

bofire/data_models/objectives/api.py
```diff
@@ -25,6 +29,8 @@
 ]
 
 AnyConstraintObjective = Union[
+    CategoricalObjective,
```
jduerholt, line 32: CategoricalObjective is just the abstract class. It should not be listed here.

bofire/data_models/objectives/categorical.py
```diff
@@ -0,0 +1,106 @@
...
+            ValueError: when categories have non-unique names
+
+        Returns:
+            Tuple[str]: Tuple of the categories
+        """
+        if len(categories) != len(set(categories)):
+            raise ValueError("categories must be unique")
+        return tuple(categories)
+
+    def __call__(self, x: Union[pd.Series, np.ndarray]) -> Union[pd.Series, np.ndarray]:
```
jduerholt, line 39: Remove this method, it should stay as abstract objective.

bofire/data_models/objectives/categorical.py
```diff
@@ -0,0 +1,106 @@
...
+            categories (Union[List[str], Tuple[str]]): List or tuple of category names
+
+        Raises:
+            ValueError: when desirability count is not equal to category count
+
+        Returns:
+            Tuple[bool]: Tuple of the desirability
+        """
+        if len(desirability) != len(values["categories"]):
+            raise ValueError(
```
jduerholt, line 77: We should add a test for this via the add_invalid syntax in `data_models/specs/objectives.py`

bofire/data_models/objectives/categorical.py
```diff
@@ -0,0 +1,106 @@
...
+                "number of categories differs from number of desirabilities"
+            )
+        return tuple(desirability)
+
+    def to_dict(self) -> Dict:
```
jduerholt, line 82: We need tests for these methods.

bofire/data_models/surrogates/surrogate.py
```diff
@@ -28,3 +29,17 @@ def validate_outputs(cls, v, values):
         if len(v) == 0:
             raise ValueError("At least one output feature has to be provided.")
         return v
+
```
jduerholt, line 32: We need the actual validator that uses the is_output_implemented method, have a look here: https://github.com/experimental-design/bofire/blob/73c1db7876d72e15694610e6b068126da96b98d5/bofire/data_models/strategies/strategy.py#L38 and then we need to test if for at least one surrogate using the `add_invalid` syntax in `specs/surrogates.py`

bofire/data_models/surrogates/fully_bayesian.py
```diff
@@ -16,3 +17,16 @@ def validate_thinning(cls, value, values):
...
+        """Abstract method to check output type for surrogate models
+
+        Args:
+            outputs: objective functions for the surrogate
+            my_type: continuous or categorical output
+
+        Returns:
+            bool: True if the output type is valid for the surrogate chosen, False otherwise
+        """
+        return True if isinstance(my_type, ContinuousOutput) else False
```
jduerholt, line 32:
```suggestion
        return  isinstance(my_type, ContinuousOutput)
```

bofire/data_models/surrogates/linear.py
```diff
@@ -14,3 +18,17 @@ class LinearSurrogate(TrainableBotorchSurrogate):
...
+        """Abstract method to check output type for surrogate models
+
+        Args:
+            outputs: objective functions for the surrogate
+            my_type: continuous or categorical output
+
+        Returns:
+            bool: True if the output type is valid for the surrogate chosen, False otherwise
+        """
+        return True if isinstance(my_type, ContinuousOutput) else False
```
jduerholt, line 34:
```suggestion
        return  isinstance(my_type, ContinuousOutput)
```

bofire/data_models/surrogates/mixed_single_task_gp.py
```diff
@@ -29,3 +30,16 @@ def validate_categoricals(cls, v, values):
...
+        """Abstract method to check output type for surrogate models
+
+        Args:
+            outputs: objective functions for the surrogate
+            my_type: continuous or categorical output
+
+        Returns:
+            bool: True if the output type is valid for the surrogate chosen, False otherwise
+        """
+        return True if isinstance(my_type, ContinuousOutput) else False
```
jduerholt, line 45:
```suggestion
        return  isinstance(my_type, ContinuousOutput)
```

bofire/data_models/surrogates/mlp.py
```diff
@@ -18,4 +19,17 @@ class MLPEnsemble(TrainableBotorchSurrogate):
...
+        """Abstract method to check output type for surrogate models
+
+        Args:
+            outputs: objective functions for the surrogate
+            my_type: continuous or categorical output
+
+        Returns:
+            bool: True if the output type is valid for the surrogate chosen, False otherwise
+        """
+        return True if isinstance(my_type, ContinuousOutput) else False
```
jduerholt, line 35:
```suggestion
        return  isinstance(my_type, ContinuousOutput)
```

bofire/data_models/surrogates/mlp_classifier.py
```diff
@@ -0,0 +1,36 @@
...
+        """Abstract method to check output type for surrogate models
+
+        Args:
+            outputs: objective functions for the surrogate
+            my_type: continuous or categorical output
+
+        Returns:
+            bool: True if the output type is valid for the surrogate chosen, False otherwise
+        """
+        return True if isinstance(my_type, CategoricalOutput) else False
```
jduerholt, line 36:
```suggestion
        return  isinstance(my_type, CategoricalOutput)
```

bofire/data_models/surrogates/polynomial.py
```diff
@@ -23,3 +24,16 @@ def from_power(power: int, inputs: Inputs, outputs: Outputs):
...
+        """Abstract method to check output type for surrogate models
+
+        Args:
+            outputs: objective functions for the surrogate
+            my_type: continuous or categorical output
+
+        Returns:
+            bool: True if the output type is valid for the surrogate chosen, False otherwise
+        """
+        return True if isinstance(my_type, ContinuousOutput) else False
```
jduerholt, line 39:
```suggestion
        return  isinstance(my_type, ContinuousOutput)
```

bofire/data_models/surrogates/random_forest.py
```diff
@@ -30,5 +30,16 @@ class RandomForestSurrogate(TrainableBotorchSurrogate):
...
+        """Abstract method to check output type for surrogate models
+
+        Args:
+            outputs: objective functions for the surrogate
+            my_type: continuous or categorical output
+
+        Returns:
+            bool: True if the output type is valid for the surrogate chosen, False otherwise
+        """
+        return True if isinstance(my_type, ContinuousOutput) else False
```
jduerholt, line 45:
```suggestion
        return  isinstance(my_type, ContinuousOutput)
```

bofire/data_models/surrogates/single_task_gp.py
```diff
@@ -108,3 +112,16 @@ class SingleTaskGPSurrogate(TrainableBotorchSurrogate):
...
+        """Abstract method to check output type for surrogate models
+
+        Args:
+            outputs: objective functions for the surrogate
+            my_type: continuous or categorical output
+
+        Returns:
+            bool: True if the output type is valid for the surrogate chosen, False otherwise
+        """
+        return True if isinstance(my_type, ContinuousOutput) else False
```
jduerholt, line 127:
```suggestion
        return  isinstance(my_type, ContinuousOutput)
```

bofire/data_models/surrogates/tanimoto_gp.py
```diff
@@ -26,3 +27,16 @@ class TanimotoGPSurrogate(TrainableBotorchSurrogate):
...
+        """Abstract method to check output type for surrogate models
+
+        Args:
+            outputs: objective functions for the surrogate
+            my_type: continuous or categorical output
+
+        Returns:
+            bool: True if the output type is valid for the surrogate chosen, False otherwise
+        """
+        return True if isinstance(my_type, ContinuousOutput) else False
```
jduerholt, line 42:
```suggestion
        return  isinstance(my_type, ContinuousOutput)
```

bofire/data_models/surrogates/xgb.py
```diff
@@ -75,3 +77,16 @@ def validate_input_preprocessing_specs(cls, v, values):
...
+        """Abstract method to check output type for surrogate models
+
+        Args:
+            outputs: objective functions for the surrogate
+            my_type: continuous or categorical output
+
+        Returns:
+            bool: True if the output type is valid for the surrogate chosen, False otherwise
+        """
+        return True if isinstance(my_type, ContinuousOutput) else False
```
jduerholt, line 92:
```suggestion
        return  isinstance(my_type, ContinuousOutput)
```

jduerholt (changes requested):
Hi Gabe,

I looked over the first part of the PR and let inline comments. I will look over the rest during the Christmas days. 

Main comments are due to missing tests.

Thanks!

Best,

Johannes

bofire/strategies/predictives/predictive.py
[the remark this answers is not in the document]
gmancino: @jduerholt any comment on the above? Which functionality do you prefer?

bofire/surrogates/surrogate.py
```diff
@@ -44,27 +45,80 @@ def predict(self, X: pd.DataFrame) -> pd.DataFrame:
             Xt[c] = pd.to_numeric(Xt[c], errors="raise")
         # predict
         preds, stds = self._predict(Xt)
+        # set up column names
+        pred_cols = []
+        sd_cols = []
+        for featkey in self.outputs.get_keys():
+            if hasattr(self.outputs.get_by_key(featkey), "categories"):
```
gmancino, line 52: @jduerholt, there is a type error failing here. How do you recommend checking if the output type is categorical? Line 71 below also fails...

bofire/surrogates/mlp.py
```diff
@@ -109,40 +125,50 @@ def forward(self, X: Tensor):
...
 def fit_mlp(
     mlp: MLP,
-    dataset: RegressionDataSet,
+    dataset: MLPDataset,
     batch_size: int = 10,
     n_epoches: int = 200,
     lr: float = 1e-4,
     shuffle: bool = True,
     weight_decay: float = 0.0,
+    loss_function: Union[
```
gmancino, line 139: @jduerholt there is a type error here. I tried removing the type from the loss function and also trying nn.module. I could try nn.modules.loss unless you have any other suggestions?

bofire/surrogates/surrogate.py
jduerholt: `for feat in self.outputs.get(CategoricalOutput)`: Doing it like this, you only get categorical output features ;) The same filtereing also applies to `get_keys`. Have a look at the method in `domain.features.py`, they are super helpful and I use them really a lot, so you should familiarize with them.

bofire/surrogates/mlp.py
jduerholt: Puh, no idea, then just set a `type: ignore` behind it ...

bofire/surrogates/surrogate.py
jduerholt: If you only need the keys use `get_keys(CategoricalOutput)`, it could be that you have still a type error, then add a `type: ignore`

jduerholt:
Regarding you failing test, something went wrong in one of your merges against main, in main the line with the correct terms looks as following: https://github.com/experimental-design/bofire/blob/bf2a11999a7ba779a758c70ee17a555213676d6b/tests/bofire/strategies/doe/test_utils.py#L77

In your branch it looks likes this: https://github.com/experimental-design/bofire/blob/20ec136ffdf8bf662c9f332aac2f25bdfbf85116/tests/bofire/strategies/doe/test_utils.py#L77. In main we updated it at some point to the version looking like this `"x ** 2"` which is the new formulaic format, whereas you have in your branch still `x**2` which is the old format. 

I assume some problems with a merge.

tests/bofire/strategies/doe/test_utils.py
```diff
@@ -74,7 +74,7 @@ def test_get_formula_from_string():
     assert all(term in np.array(model_formula, dtype=str) for term in terms)
 
     # linear and quadratic
-    terms = ["1", "x0", "x1", "x2", "x0 ** 2", "x1 ** 2", "x2 ** 2"]
+    terms = ["1", "x0", "x1", "x2", "x0**2", "x1**2", "x2**2"]
```
jduerholt, line 77: One can see it also here that your branch differes from main in `doe/test_utils.py`

gmancino:
@jduerholt This is ready for another round of reviews :) Hopefully the last one!

bofire/data_models/domain/features.py
```diff
@@ -591,8 +596,8 @@ def get_by_objective(
             return Outputs(
                 features=sorted(
                     filter_by_attribute(
-                        self.get(ContinuousOutput).features,
-                        lambda of: of.objective,  # type: ignore
+                        self.get([ContinuousOutput, CategoricalOutput]).features,
```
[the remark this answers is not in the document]
gmancino: A test should be implemented in the appropriate file now.

bofire/data_models/domain/features.py
```diff
@@ -643,6 +652,14 @@ def __call__(
                 feat(experiments[f"{feat.key}_pred" if predictions else feat.key])  # type: ignore
                 for feat in self.features
                 if feat.objective is not None
+                and not isinstance(feat, CategoricalOutput)
+            ]
+            + [
+                pd.Series(data=feat(experiments.filter(regex=f"{feat.key}_pred_")), name=f"{feat.key}_pred")  # type: ignore
+                if predictions
+                else experiments[feat.key]
+                for feat in self.features
```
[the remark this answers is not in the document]
gmancino: Functionality should be updated to allow for `None` and solve the issues above.

bofire/data_models/domain/features.py
```diff
@@ -690,24 +707,24 @@ def validate_experiments(self, experiments: pd.DataFrame) -> pd.DataFrame:
...
     def validate_candidates(self, candidates: pd.DataFrame) -> pd.DataFrame:
         # for each continuous output feature with an attached objective object
-        # ToDo: adjust it for the CategoricalOutput
-        cols = list(
+        continuous_cols = list(
             itertools.chain.from_iterable(
                 [
-                    [f"{key}_pred", f"{key}_sd", f"{key}_des"]
-                    for key in self.get_keys_by_objective(Objective)
+                    [f"{obj.key}_pred", f"{obj.key}_sd", f"{obj.key}_des"]
```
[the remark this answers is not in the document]
gmancino: Done

gmancino (commented):
I should have addressed all previous concerns. The tutorials also show the functionality of the new feature. We can add extra tests, but I added the major ones for now I believe.

bofire/surrogates/mlp.py
```diff
@@ -109,40 +125,48 @@ def forward(self, X: Tensor):
...
 def fit_mlp(
     mlp: MLP,
-    dataset: RegressionDataSet,
+    dataset: MLPDataset,
     batch_size: int = 10,
     n_epoches: int = 200,
     lr: float = 1e-4,
     shuffle: bool = True,
     weight_decay: float = 0.0,
+    loss_function=nn.L1Loss,  # type: ignore
```
jduerholt, line 139: Should we somehow list and check which we support currently?

bofire/surrogates/mlp.py
```diff
@@ -109,40 +125,48 @@ def forward(self, X: Tensor):
...
             inputs, targets = data
-            if len(targets.shape) == 1:
+            if isinstance(loss_function, nn.CrossEntropyLoss):
+                targets = targets.flatten().long()
+            elif len(targets.shape) == 1 and not isinstance(
+                loss_function, nn.CrossEntropyLoss
+            ):
                 targets = targets.reshape((targets.shape[0], 1))
+            else:
+                pass
```
jduerholt, line 169: An error message? Or we can just remove the else if we do nothing.

bofire/data_models/surrogates/mlp.py
```diff
@@ -18,4 +23,44 @@ class MLPEnsemble(TrainableBotorchSurrogate):
...
+            my_type: continuous or categorical output
+        Returns:
+            bool: True if the output type is valid for the surrogate chosen, False otherwise
+        """
+        return isinstance(my_type, ContinuousOutput)
+
+
+class ClassificationMLPEnsemble(MLPEnsemble):
+    type: Literal["ClassificationMLPEnsemble"] = "ClassificationMLPEnsemble"
+    final_activation: Literal["softmax"] = "softmax"
```
jduerholt, line 56: `output_sclaer: Literal[ScalerEnum.Indetity] = ScalerEnum.Indetity`

I do not know if this works, but we have to forbid that a user assigns an output scaler, if it not works via the Literal, we have to add a field validator for this.

bofire/surrogates/surrogate.py
```diff
@@ -44,27 +45,75 @@ def predict(self, X: pd.DataFrame) -> pd.DataFrame:
...
         self.validate_predictions(predictions=predictions)
         # return
         return predictions
 
     def validate_predictions(self, predictions: pd.DataFrame) -> pd.DataFrame:
-        expected_cols = [
-            f"{key}_{t}" for key in self.outputs.get_keys() for t in ["pred", "sd"]
-        ]
+        expected_cols = []
+        check_columns = []
```
jduerholt, line 91: We are composing this kind of lists add a lof of different places, a few examples are:
- `data_models/domain/features.py`
- `data_models/strategies/predictives/predictive.py`

Can we create a utility method to create the correct col names etc. and move it into utils and just call it from different places. This will also making testing much easier.

bofire/surrogates/trainable.py
```diff
@@ -181,6 +182,22 @@ def cross_validate(
             # now do the scoring
             y_test_pred = self.predict(X_test)  # type: ignore
             y_train_pred = self.predict(X_train)  # type: ignore
+
+            # Convert to categorical if applicable
+            if isinstance(self.outputs.get_by_key(key).objective, ConstrainedCategoricalObjective):  # type: ignore
```
jduerholt, line 187: Hmm, why only of `ConstrainedCategoricalObjective`? This should apply always for every `CategoricalOutput` feature, or?

jduerholt: And in general, I am not 100% sure, what should be achieved here.

bofire/utils/torch_tools.py
```diff
@@ -188,24 +188,57 @@ def constrained_objective2botorch(
...
+                            Z[..., idx : idx + len(objective.desirability)]
+                            * torch.tensor(objective.desirability).to(**tkwargs)
+                        ).sum(-1)
+                        - 1,
+                        min=eps,
+                        max=1 - eps,
+                    )
+                )
+            ],
+            [objective.eta],
```
jduerholt, line 239: I would opt for removing the `objective.eta` and make sure that the eta which is given here is always 1 for identity, or?

tests/bofire/utils/test_torch_tools.py
```diff
@@ -739,7 +743,7 @@ def test_get_initial_conditions_generator(sequential: bool):
     ],
 )
 def test_constrained_objective2botorch(objective):
```
jduerholt, line 745: We need here also the tests for the new objective, or?

bofire/data_models/domain/features.py
```diff
@@ -690,24 +707,24 @@ def validate_experiments(self, experiments: pd.DataFrame) -> pd.DataFrame:
...
+                    for obj in self.get_by_objective(
+                        includes=Objective, excludes=CategoricalObjective
+                    )
                 ]
                 + [
                     [f"{key}_pred", f"{key}_sd"]
                     for key in self.get_keys_by_objective(
-                        excludes=Objective,
-                        includes=None,  # type: ignore
+                        excludes=Objective, includes=None  # type: ignore
```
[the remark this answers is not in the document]
gmancino: Does it make sense for `CategoricalObjective` to be None type? What would you prefer?

bofire/data_models/domain/features.py
```diff
@@ -672,23 +689,18 @@ def validate_experiments(self, experiments: pd.DataFrame) -> pd.DataFrame:
...
-        cols = list(
+        continuous_cols = list(
             itertools.chain.from_iterable(
                 [
-                    [f"{key}_pred", f"{key}_sd", f"{key}_des"]
-                    for key in self.get_keys_by_objective(Objective)
-                ]
-                + [
-                    [f"{key}_pred", f"{key}_sd"]
-                    for key in self.get_keys_by_objective(
```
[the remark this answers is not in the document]
gmancino: This should be handled in the below thread :)

bofire/data_models/features/categorical.py
```diff
@@ -359,38 +361,63 @@ class CategoricalOutput(Output):
     order_id: ClassVar[int] = 9
 
     categories: TCategoryVals
-    objective: Annotated[List[Annotated[float, Field(ge=0, le=1)]], Field(min_length=2)]
+    objective: Optional[AnyCategoricalObjective] = Field(
+        default_factory=lambda: ConstrainedCategoricalObjective(
```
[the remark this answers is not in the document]
gmancino: Should we set desireabilities to be an array of `True`'s in this instance?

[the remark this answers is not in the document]
gmancino: This actually ties into the above questions (i.e. regarding `None` types). I am also not sure of how to instantiate in this instance since it seems sort of "chicken-and-egg." Either way, we just need a default case here. Let me know what you prefer.

bofire/data_models/features/categorical.py
```diff
@@ -359,38 +361,63 @@ class CategoricalOutput(Output):
...
+            categories (List[str]): List or tuple of category names
+
+        Raises:
+            ValueError: when categories do not match objective categories
+
+        Returns:
+            Tuple[str]: Tuple of the categories
+        """
+        if self.objective.categories != self.categories:  # type: ignore
+            raise ValueError("categories must match to objective categories")
```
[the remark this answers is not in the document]
gmancino: It will be in specs/outputs in the next commit

bofire/data_models/objectives/categorical.py
```diff
@@ -0,0 +1,83 @@
...
+    w: TWeight = 1.0
+    categories: TCategoryVals
+    desirability: List[bool]
+    eta: float = 1.0
+    type: Literal["ConstrainedCategoricalObjective"] = "ConstrainedCategoricalObjective"
+
+    @field_validator(
+        "desirability",
+    )
+    def validate_categories_unique(cls, desirability: List[bool], info) -> List[bool]:
```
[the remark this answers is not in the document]
gmancino: Makes sense, will add this. I am assuming tests should be added in `specs/objectives.py`, right?

bofire/data_models/objectives/categorical.py
gmancino: They are in tests/bofire/data_models/domain/test_outputs.py. Let me know if they are insufficient or if you want them elsewhere (should be line 290)

bofire/data_models/surrogates/api.py
```diff
@@ -50,7 +55,8 @@
         SingleTaskGPSurrogate,
         MixedSingleTaskGPSurrogate,
         MixedTanimotoGPSurrogate,
-        MLPEnsemble,
+        ClassificationMLPEnsemble,
+        RegressionMLPEnsemble,
```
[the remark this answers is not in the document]
gmancino: Sure, right now, `AnyClassificatonSurrogate = ClassificationMLPEnsemble` will need to be sufficient since it is the only surrogate at this point in time

bofire/data_models/surrogates/fully_bayesian.py
```diff
@@ -18,3 +19,13 @@ def validate_thinning(cls, thinning, info):
...
+
+    @classmethod
+    def is_output_implemented(cls, my_type: Type[AnyOutput]) -> bool:
+        """Abstract method to check output type for surrogate models
+        Args:
+            my_type: continuous or categorical output
+        Returns:
+            bool: True if the output type is valid for the surrogate chosen, False otherwise
+        """
+        return isinstance(my_type, ContinuousOutput)
```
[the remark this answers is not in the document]
gmancino: Not sure what you mean here - tests in tests/bofire/data_models/specs/surrogates.py suggest this feature is working appropriately...

bofire/surrogates/mlp.py
gmancino: How do you recommend?

bofire/surrogates/mlp.py
gmancino: I will remove

bofire/surrogates/trainable.py
gmancino: CategoricalOutputs do not contain the functions to do the mapping (as per our previous discussions) from strings to categorical encodings. We need scalars for the loss function, not categories, for which Y would be an array of strings due to BoFire requiring that categorical columns be validated

bofire/utils/torch_tools.py
gmancino: I can do this, but will it break functionality elsewhere? Is eta always required for type Objective?

bofire/data_models/features/categorical.py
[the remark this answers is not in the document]
jduerholt: Let us do the following, we implement the default generation into the field validator for the CategeforicalOutput:


``` python
def validate_objective(cls, objective, info):
    if objective is None:
         return ConstrainedCategoricalObjective(categories=info.data[`categories`], desirabilities = [True]*len(info.data[`categories`])]
    else:
        pass # do the validaton already implemented in the validator.
```

By this we are then setting as default that everything is allowed, which is the same as ignorig the CategoricalOutput in the optimization. 

In the attribute, we then set the folowing:

`objective: Optional[AnyCategoricalObjective] = Field(default=None, validate_default=True)`

Does this makes sense to you?

bofire/data_models/objectives/categorical.py
```diff
@@ -0,0 +1,83 @@
...
+    ConstrainedObjective, CategoricalObjective, Objective
+):
+    """Compute the categorical objective value as:
+
+        Po where P is an [n, c] matrix where each row is a probability vector
+        (P[i, :].sum()=1 for all i) and o is a vector of size [c] of objective values
+
+    Attributes:
+        w (float): float between zero and one for weighting the objective.
+        desirability (list): list of values of size c (c is number of categories) such that the i-th entry is in {True, False}
```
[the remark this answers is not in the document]
jduerholt: `w` is just used in case of Additive and Multiplicative Sobo strategies. What happens actually there if we use there a CategoricalOuput? I have no idea :D But lets keep the `w` and checks what happens in these strategies.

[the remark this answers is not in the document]
jduerholt: `eta` is the steepness of the sigmoid based output constrained. In the MaxSigmoid it is actually called `steepness`, as in this categorical output constraint, the steepness is controlled totally by the classifier and its calibration, we should ignore `eta`, as we are implicitly assuming an `eta` of one in your iddentiy transformation for botorch, correct?

bofire/data_models/objectives/categorical.py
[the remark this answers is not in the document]
jduerholt: But you do need the validator anymore, as I introduced this global new type: https://github.com/experimental-design/bofire/blob/67f33201dee5526d1ab620a8411f83098cbce7d1/bofire/data_models/types.py#L51 which has the validation already cooked in, just set `categories: TCategoryVals` after merging main in.

bofire/data_models/surrogates/api.py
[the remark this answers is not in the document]
jduerholt: Correct!

bofire/data_models/objectives/categorical.py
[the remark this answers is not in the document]
gmancino: So I only need the validator for desireability, correct?

bofire/surrogates/mlp.py
gmancino: Something like Union[torch.nn.modules.loss.L1Loss, ...]?

bofire/data_models/surrogates/fully_bayesian.py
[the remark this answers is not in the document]
jduerholt: Yep, it is working but it is somehow inconsistent with the `is_feature_implemented`, `is_constraint_implemented` in the strategies, there always the type of the object is passed to the is_implemented method and the actual instance.

bofire/surrogates/mlp.py
jduerholt: Let's forget it. It is fine.

bofire/surrogates/surrogate.py
jduerholt: Let us put it into a new file under utils, as you want it ...

bofire/utils/torch_tools.py
jduerholt: Should not be a problem.

bofire/data_models/surrogates/mlp.py
```diff
@@ -18,4 +23,44 @@ class MLPEnsemble(TrainableBotorchSurrogate):
     weight_decay: Annotated[float, Field(ge=0.0)] = 0.0
     subsample_fraction: Annotated[float, Field(gt=0.0)] = 1.0
     shuffle: bool = True
-    scaler: ScalerEnum = ScalerEnum.STANDARDIZE
+    scaler: ScalerEnum = ScalerEnum.NORMALIZE
+
+    @classmethod
+    def is_output_implemented(cls, my_type: Type[AnyOutput]) -> bool:
```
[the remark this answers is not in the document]
gmancino: To clarify, your desired structure is: `Surrogate` -> `RegressionSurrogat(Surrogate)` and `ClassificationSurrogate(Surrogate)` where `is_output_implemented` is now a validator for type Continuous (for Regression) and Categorical (for Classification)? Then, each surrogate will be of type e.g. `(TrainableBotorchSurrogate,RergressionSurrogate)` for regression problems. Is this what you have in mind?

bofire/data_models/features/categorical.py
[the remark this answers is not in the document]
gmancino: Intuitively this makes sense, however, just applying this as is actually causes the following error: FAILED tests/bofire/data_models/serialization/test_deserialization.py::test_outputs_should_be_deserializable[outputs_spec0] 

This does not make sense since these output types are Continuous. Care to comment?

bofire/data_models/features/categorical.py
gmancino: It will be in features/test_categorical.py

bofire/data_models/surrogates/fully_bayesian.py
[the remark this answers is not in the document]
gmancino: Will update appropriately in the next commit

bofire/data_models/domain/features.py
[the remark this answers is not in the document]
gmancino: Please check bofire/utils/naming_conventions.py

bofire/data_models/objectives/categorical.py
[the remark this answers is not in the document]
gmancino: Correct, this should be complete now

bofire/strategies/predictives/predictive.py
```diff
@@ -102,17 +104,32 @@ def predict(self, experiments: pd.DataFrame) -> pd.DataFrame:
...
-                columns=["%s_pred" % feat.key for feat in self.domain.outputs.get()],
+                columns=pred_cols,
             )
+        for feat in self.domain.outputs.get():
+            if isinstance(feat, CategoricalOutput):
+                predictions.insert(
+                    loc=0,
+                    column=f"{feat.key}_pred",
+                    value=predictions.filter(regex=f"{feat.key}(.*)_prob")
+                    .idxmax(1)
```
jduerholt, line 123: This is doubled with `surrogte.py`, could we somehow also write a helper function for this and call them in both ocassions, some method called for example `postprocess_categegorical_predictions`? We could place it in `surrogate.py`. What do you think?

bofire/utils/naming_conventions.py
```diff
@@ -0,0 +1,24 @@
...
+
+def get_column_names(outputs) -> Tuple[List[str], List[str]]:
+    pred_cols, sd_cols = [], []
+    for featkey in outputs.get_keys(CategoricalOutput):  # type: ignore
+        pred_cols = pred_cols + [
+            f"{featkey}_{cat}_prob"
+            for cat in outputs.get_by_key(featkey).categories  # type: ignore
+        ]
+        sd_cols = sd_cols + [
+            f"{featkey}_{cat}_sd"
```
jduerholt, line 17: Nice. Could you also add a test for it?

bofire/utils/naming_conventions.py
```diff
@@ -0,0 +1,24 @@
+from typing import List, Tuple
+
+from bofire.data_models.features.api import (
+    CategoricalOutput,
+    ContinuousOutput,
+)
+
+
+def get_column_names(outputs) -> Tuple[List[str], List[str]]:
```
jduerholt, line 9:
```suggestion
def get_column_names(outputs: Outputs) -> Tuple[List[str], List[str]]:
```

jduerholt, line 9: And could you add a simple docstring?

tests/bofire/surrogates/test_mlp.py
```diff
@@ -13,8 +13,8 @@
     ContinuousInput,
     ContinuousOutput,
 )
-from bofire.data_models.surrogates.api import MLPEnsemble, ScalerEnum
-from bofire.surrogates.mlp import MLP, RegressionDataSet, _MLPEnsemble, fit_mlp
+from bofire.data_models.surrogates.api import RegressionMLPEnsemble, ScalerEnum
+from bofire.surrogates.mlp import MLP, MLPDataset, _MLPEnsemble, fit_mlp
```
jduerholt, line 17: Can you also add a test for the Classification Ensemble?

jduerholt (commented):
Hi Gabe, 

due to the lack of time, not a full review, but a few minor things to work on. We are very close to the finish line.

Best,

Johannes

bofire/strategies/predictives/predictive.py
gmancino: I will add it to the naming_conventions file. Unless you can tell me how to access the surrogates within the predictives to call a method specified within the surrogate class?

tests/bofire/utils/test_torch_tools.py
gmancino: These are added :)

tests/bofire/data_models/specs/features.py
```diff
@@ -120,7 +123,16 @@
     },
 )
 
-
+specs.add_valid(
```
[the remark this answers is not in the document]
gmancino: This is done in the data_models/specs/outputs.py

bofire/data_models/objectives/categorical.py
gmancino: Done

jduerholt (changes requested):
Looks very good, last two things are commented. When they are finished, we can merge it in ;)

bofire/data_models/objectives/categorical.py
```diff
@@ -0,0 +1,80 @@
...
+)
+from bofire.data_models.types import TCategoryVals
+
+
+class CategoricalObjective:
+    """Abstract categorical objective class"""
+
+
+class ConstrainedCategoricalObjective(
+    ConstrainedObjective, CategoricalObjective, Objective
```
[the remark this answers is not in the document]
gmancino: Sounds good - I will have to keep `Objective` to make sure it has the appropriate functionality for being called, etc.

gmancino:
@jduerholt the request changes are complete :)

jduerholt (approved):
Thank you very much for all your efforts and your patience with me!

PR #298, opened 2023-10-05
add polynomial kernel

dlinzner-bcs:
- added support for polynomial kernels
- add QuadraticSurrogate that performs Bayesian quadratic regression

### DIFF
--- a/bofire/data_models/surrogates/quadratic.py
+++ b/bofire/data_models/surrogates/quadratic.py
@@ -1,0 +1,22 @@
+from typing import Literal
+
+from pydantic import Field
+
+from bofire.data_models.kernels.api import (
+    LinearKernel,
+    PolynomialKernel,
+)
+from bofire.data_models.priors.api import BOTORCH_NOISE_PRIOR, AnyPrior
+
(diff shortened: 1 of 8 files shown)


bofire/data_models/surrogates/quadratic.py
```diff
@@ -0,0 +1,21 @@
...
+)
+from bofire.data_models.priors.api import BOTORCH_NOISE_PRIOR, AnyPrior
+
+# from bofire.data_models.strategies.api import FactorialStrategy
+from bofire.data_models.surrogates.botorch import BotorchSurrogate
+from bofire.data_models.surrogates.scaler import ScalerEnum
+from bofire.data_models.surrogates.trainable import TrainableSurrogate
+
+
+class QuadraticSurrogate(BotorchSurrogate, TrainableSurrogate):
```
bertiqwerty, line 16: Rather `PolynomialSurrogate`?

bertiqwerty (approved):

bofire/data_models/surrogates/quadratic.py
dlinzner-bcs: This specific one has power=2. So its quadratic.

PR #300, opened 2023-10-10
286 add surrogate models that can extrapolate

dlinzner-bcs:
- again the stupid bug that the surrogate has to be allowed for BotochSurrogate


dlinzner-bcs:
@bertiqwerty the test is there :) can you please check?

bertiqwerty (approved):
As discussed.

PR #304, opened 2023-10-26
allow for max_active=#(variables in constraint) in n_choose_k_constraints_as_bounds()

Osburg:
This PR fixed an error occuring in the edge case of an NChooseKConstraint w/ max_active=#(variables in NChooseKConstraint). In this case no additional restrictions to the bounds have to be made.


jduerholt (approved):
Looks good to me.

PR #305, opened 2023-10-27
add output_scaler to mlp

simonsung06:
Added an output_scaler option for the MLP Ensemble surrogate. This is to give extra flexibility where standardizing data often improves fitting. I have set `ScalerEnum.STANDARDIZE` as the default output_scaler because this is common practice for regression problems. Currently only standardization and identity scalers are usable. Normalization is not implemented since there is no output transform for this natively on BoTorch. If needed one day, maybe we could implement it (but i believe standardization should be sufficient for us).

### DIFF
--- a/bofire/data_models/surrogates/mlp.py
+++ b/bofire/data_models/surrogates/mlp.py
@@ -1,6 +1,6 @@
 from typing import Annotated, Literal, Sequence
 
-from pydantic import Field
+from pydantic import Field, validator
 
 from bofire.data_models.surrogates.botorch import BotorchSurrogate
 from bofire.data_models.surrogates.scaler import ScalerEnum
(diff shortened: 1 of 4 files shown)


jduerholt:
@simonsung06: I think the tests are failing due to the rdkit issue that you already fixed. I just merged it into main. Just merge it into your PR and the tests should work again!

bofire/data_models/surrogates/mlp.py
```diff
@@ -20,3 +20,4 @@ class MLPEnsemble(BotorchSurrogate, TrainableSurrogate):
     subsample_fraction: Annotated[float, Field(gt=0.0)] = 1.0
     shuffle: bool = True
     scaler: ScalerEnum = ScalerEnum.NORMALIZE
+    output_scaler: ScalerEnum = ScalerEnum.STANDARDIZE
```
jduerholt, line 23: We should add a validator here, that blocks the `Normalize` enum, so that it crashes when creating the data model and not when training it.

jduerholt (changes requested):
Thank you very much @simonsung06. Very good addition. I only added a minor comment. 

What do you think about adding this also for GPs? There we always use Standardization, but it would be nice to have the option ...

bofire/data_models/surrogates/mlp.py
simonsung06: Added the validator

bofire/surrogates/mlp.py
```diff
@@ -179,6 +186,15 @@ def _fit(self, X: pd.DataFrame, Y: pd.DataFrame):
         scaler = get_scaler(self.inputs, self.input_preprocessing_specs, self.scaler, X)
         transformed_X = self.inputs.transform(X, self.input_preprocessing_specs)
 
+        if self.output_scaler == ScalerEnum.IDENTITY:
+            output_scaler = None
+        elif self.output_scaler == ScalerEnum.STANDARDIZE:
+            output_scaler = Standardize(m=Y.shape[-1])
+        elif self.output_scaler == ScalerEnum.NORMALIZE:
```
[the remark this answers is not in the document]
simonsung06: Removed the checks from here

simonsung06:
Thanks for the suggestion. The changes have been made.

Regarding adding it for GP's too _should be_ an easy addition. Shall I do this in another PR?

jduerholt:
> Regarding adding it for GP's too _should be_ an easy addition. Would you like me to do this in another PR?

No need to hurry, if you find time, it would be nice to have it in a seperate PR.

jduerholt (approved):
Looks good to me. Thanks!

PR #306, opened 2023-10-27
updated rdkit fragment/descriptor list

simonsung06:
The descriptors list for RDKit was changed recently therefore the indexes for the fragments in the list also changed. The number and type of fragments did not change; just that additional descriptors were added to RDKit.

Updated code so that it is more robust against these RDKit changes


jduerholt (approved):
Looks good to me. Thx!

PR #308, opened 2023-11-08
Add more log ACQFs

jduerholt:
- qLogNEI as new default ACQF for SOBO
- new MOBO strategy which features all functionality of the QnehviStrategy and the QehviStrategy. These strategies are now deprecated and will be removed in a future release. 
- MOBO features the following ACQFs:
    - qLogNEHVI (default one)
    - qNEHVI
    - qEHVI
    - qLogEHVI
 - Botorch 0.9.4 as minimal required version


tests/bofire/strategies/test_mobo.py
```diff
@@ -0,0 +1,214 @@
...
+        for acqf in [
+            acquisitions.qEHVI,
+            acquisitions.qLogEHVI,
+            acquisitions.qNEHVI,
+            acquisitions.qLogNEHVI,
+        ]
+        for num_test_candidates in range(1, 3)
+    ],
+)
+def test_mobo(strategy, use_ref_point, acqf, num_test_candidates):
```
simonsung06, line 90: is num_test_candidates being used in this function?

simonsung06 (commented):
Looks good. One minor thing and then I think it's ready. I like the decision to use the new Log-based acq functions!

tests/bofire/strategies/test_mobo.py
jduerholt: Good point, this is an copy paste thing, I will remove it.

jduerholt: Fixed it.

simonsung06 (approved):
Thanks! Will merge it.

PR #309, opened 2023-11-09
gp output scaler

simonsung06:
Added output scaler option for GP's

There is repeated code for the Pydantic validator so we probably need to think about a better way to do this in the future.


jduerholt (commented):
Hi Simon,

I think I have an idea, all data models in which we added the scaler show the following inheritance:

`class SingleTaskGPSurrogate(BotorchSurrogate, TrainableSurrogate):``

What do you think about creating a new data model `TrainableBotorchSurrogate`, add the scaler stuff incl. the validator there and inherit from there?

Best,

Johannes

simonsung06:
Hi Johannes. That is a possible idea about would be suitable for now I think. I'm not sure about how pydantic2 works though and how it changes things...

For now I'll make your suggested change and commit it when i get the chance

simonsung06:
Hi @jduerholt,

I've made the suggested change by adding a new `TrainableBotorchSurrogate` data model.

In addition, the random forest data model was also modified because it originally inherits from `BotorchSurrogate` and `TrainableSurrogate` too. Whilst this is not completely necessary, I changed it to inherit `TrainableBotorchSurrogate` so `RandomForestSurrogate` will now also take `scaler` and `output_scalar` keywords arguments, which both default to  `ScalerEnum.IDENTITY`

jduerholt:
Thx, I will have a look over the course of the week!

PR #317, opened 2023-12-11
Botorch 0.9.5

jduerholt:
BoTorch version 0.9.5 was released yesterday. It features the possibility to use q>1 for nonlinear constraints which is very helpful for us due to how NChooseK constraints are implemented in the BO part of BoFire. This PR sets the min required BoTorch version for BoFire to 0.9.5.

In addition, the following two bugs are fixed:
- Rounding of continuous inputs was only working for series of length 3.
- The random seed was not set correctly set in the polytope sampler.

### DIFF
--- a/bofire/data_models/features/continuous.py
+++ b/bofire/data_models/features/continuous.py
@@ -65,7 +65,9 @@ class ContinuousInput(NumericalInput):
         allowed_values = np.arange(
             self.lower_bound, self.upper_bound + self.stepsize, self.stepsize
         )
-        idx = abs(values.values.reshape([3, 1]) - allowed_values).argmin(axis=1)  # type: ignore
+        idx = abs(values.values.reshape([len(values), 1]) - allowed_values).argmin(
+            axis=1
+        )  # type: ignore
         return pd.Series(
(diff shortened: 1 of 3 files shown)


bofire/data_models/features/continuous.py
```diff
@@ -65,7 +65,9 @@ def round(self, values: pd.Series) -> pd.Series:
         allowed_values = np.arange(
             self.lower_bound, self.upper_bound + self.stepsize, self.stepsize
         )
-        idx = abs(values.values.reshape([3, 1]) - allowed_values).argmin(axis=1)  # type: ignore
+        idx = abs(values.values.reshape([len(values), 1]) - allowed_values).argmin(
+            axis=1
+        )  # type: ignore
```
bertiqwerty, line 70: Can you double check if this `type: ignore` is really necessary?

bertiqwerty:
Thanks Johannes. I left one minor comment.

bertiqwerty (approved):

bofire/data_models/features/continuous.py
jduerholt: Good catch. It is not needed!

Issue #320, opened 2023-12-15
DoE: Bug if fixed experiments contain columns not in domain


dlinzner-bcs:
This will result in a nonsensical design matrix and validation errors. 
Solution is to assert that fixed experiments can only contain columns that are also in the domain.

PR #321, opened 2023-12-15
DoE: Fix bug if fixed_experiments contain columns that are not in domain

dlinzner-bcs:
from here: https://github.com/experimental-design/bofire/issues/320

### DIFF
--- a/bofire/strategies/doe/design.py
+++ b/bofire/strategies/doe/design.py
@@ -408,23 +408,20 @@ def find_local_max_ipopt(
     )
 
     # no columns from partially fixed experiments which are not in the domain
-    input_set = set(domain.inputs.get_keys())
     if partially_fixed_experiments is not None:
         # check if partially fixed experiments are valid
         check_partially_fixed_experiments(
             domain, n_experiments, partially_fixed_experiments
         )
-        column_set_partial = set(partially_fixed_experiments.columns)
         partially_fixed_experiments = partially_fixed_experiments[
-            list(input_set.intersection(column_set_partial))
+            domain.inputs.get_keys()
         ]
 
     # no columns from fixed experiments which are not in the domain
     if fixed_experiments is not None:
         # check if  fixed experiments are valid
         check_fixed_experiments(domain, n_experiments, fixed_experiments)
-        column_set = set(fixed_experiments.columns)
-        fixed_experiments = fixed_experiments[list(input_set.intersection(column_set))]
+        fixed_experiments = fixed_experiments[domain.inputs.get_keys()]
 
     if partially_fix_experiment is not None and fixed_experiments is not None:
         # check if partially fixed experiments and fixed experiments are valid
--- a/bofire/strategies/doe/design.py
+++ b/bofire/strategies/doe/design.py
@@ -423,7 +423,7 @@ def find_local_max_ipopt(
         # no columns from fixed experiments which are not in the domain
         fixed_experiments = fixed_experiments[domain.inputs.get_keys()]
 
-    if partially_fix_experiment is not None and fixed_experiments is not None:
+    if (partially_fixed_experiments is not None) and (fixed_experiments is not None):
         # check if partially fixed experiments and fixed experiments are valid
         check_partially_and_fully_fixed_experiments(
             domain, n_experiments, fixed_experiments, partially_fixed_experiments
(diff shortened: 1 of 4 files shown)


jduerholt:
Just one question from my side: why is this checking necessary? When you work with the fixed and partially fixed dataframes, you could just use the input columns which are part of the domain by always stating `fixed_experiments[domain.inputs.get_keys()]`, or not? This would be bit more flexible, what do you think?

jduerholt (approved):

jduerholt:
One question: what happens if the the intersection is empty?

dlinzner-bcs:
> One question: what happens if the the intersection is empty?

The logic is now: 1. Check if all needed columns are present 2. Project dataframe on columns in `domain.inputs.get_keys()`
if 1. fails, you get an informative value error.

jduerholt:
Somehow tests are failing again ...

tests/bofire/strategies/doe/test_design.py
```diff
@@ -550,3 +550,122 @@ def test_partially_fixed_experiments():
...
+    assert doe.shape == (4, 4)
+    assert np.allclose(doe.iloc[[0, 1]]["x1"], fixed_experiments["x1"])
+    assert np.allclose(doe.iloc[[0, 1]]["x2"], fixed_experiments["x2"])
+    assert np.allclose(doe.iloc[[0, 1]]["a1"], fixed_experiments["a1"])
+    assert np.allclose(doe.iloc[[0, 1]]["a2"], fixed_experiments["a2"])
+    assert np.allclose(
+        doe.iloc[[2, 3]]["x1"], partially_fixed_experiments["x1"].astype(float)
+    )
+
+    with pytest.raises(ValueError):
```
jduerholt, line 625: Ideally, we would also catch here the exact error message to be sure that we are catching the correct stuff.

bofire/strategies/doe/design.py
```diff
@@ -402,6 +402,33 @@ def find_local_max_ipopt(
...
         raise e
 
+    # determine number of experiments (only relevant if n_experiments is not provided by the user)
+    n_experiments = get_n_experiments(
+        domain=domain, model_type=model_type, n_experiments=n_experiments
+    )
+
+    if partially_fixed_experiments is not None:
+        # check if partially fixed experiments are valid
+        check_partially_fixed_experiments(
```
jduerholt, line 412: Do we want to have a test just for these added methods?

bofire/strategies/doe/design.py
```diff
@@ -402,6 +402,33 @@ def find_local_max_ipopt(
...
+            domain, n_experiments, partially_fixed_experiments
+        )
+        # no columns from partially fixed experiments which are not in the domain
+        partially_fixed_experiments = partially_fixed_experiments[
```
jduerholt, line 416: One could also set `partially_fixed_experiments` as the return value of the check method. We do the same in domain for validate_experiments and validate_candidates.

bofire/strategies/doe/design.py
```diff
@@ -583,53 +605,76 @@ def partially_fix_experiment(
...
-    n_fixed_experiments, D = np.array(fixed_experiments).shape
+    n_fixed_experiments = len(fixed_experiments.index)
 
     if n_fixed_experiments >= n_experiments:
         raise ValueError(
             "For starting the optimization the total number of experiments must be larger that the number of fixed experiments."
         )
 
-    if D != len(domain.inputs):
+    input_set = set(domain.inputs.get_keys())
```
jduerholt, line 624: You could also somehow solve this by `domain.validate_experiments(only_inputs=True)`

jduerholt (approved):
Looks good to me, I only let a few minor comments

bofire/strategies/doe/design.py
dlinzner-bcs: this does not seem to project out columns not in domain.inputs.

bofire/strategies/doe/design.py
dlinzner-bcs: Did this. Required to change the validate_candidates function slightly (add additional flags) as `partially_fixed_experiments` will contain `None`.

dlinzner-bcs: Reverted these changes. Only use `validate_candidates` for `fixed_experiments` now - where the function can stay as it is.

PR #322, opened 2023-12-16
batch constraints for DoE and __call__ for interpointEqualityConstraint

### LINKED ISSUE #310: Batch constraint for DoE
From partner interactions we learnt that some parameters can only be varried batch-wise i.e., formulation parameters of the material and process parameters e.g., temperature. Currently we only have a heuristic way to implement it that only finds suboptimal designs. Thus, we are interesting in adding a new type of constraint that enforces a batch structure in the design. The way to achieve this is by adding equality constraints across rows.

Osburg:
This PR adds a call method for batch constraints as discussed in #313 and adds support for batch constraints / `InterpointEqualityConstraint` in DoE as discussed in #310. Closes #310


tests/bofire/data_models/test_constraints.py
```diff
@@ -133,12 +139,27 @@ def test_constraints_plus():
...
+        (constraints5, 5),
+        (constraints6, 5),
     ],
 )
 def test_constraints_call(constraints, num_candidates):
     candidates = inputs.sample(num_candidates, SamplingMethodEnum.UNIFORM)
     returned = constraints(candidates)
-    assert returned.shape == (num_candidates, len(constraints))
+
+    num_rows = 0
```
jduerholt, line 150: Are we anywhere testing that the call method returns the correct values? We need to urgently refactor our tests :D

jduerholt (changes requested):
Looks good to me, what do you think, should we add another test  in `test_design.py` in which we actually test that the DoE module is able to create such designs?

tests/bofire/data_models/test_constraints.py
Osburg: At least I haven't found them :D But added one testing if the design generation works with batch constraints.

tests/bofire/strategies/doe/test_design.py
```diff
@@ -211,6 +212,30 @@ def test_find_local_max_ipopt_results():
...
+    )
+
+    result = find_local_max_ipopt(
+        domain, "linear", ipopt_options={"maxiter": 100}, n_experiments=30
+    )
+
+    x1 = np.round(result["x1"].values, 6)
+
+    assert 0 in x1 and 1 in x1
+    for i in range(10):
```
jduerholt, line 235: you could just call `domain.validate_candidates(result, only_inputs=True)` in the test. This is also doing the check.

jduerholt (approved):
Looks good to me. @KappatC ok for you to merge?

KappatC:
> Looks good to me. @KappatC ok for you to merge?

yes, please go ahead

PR #323, opened 2023-12-18
Seed handling for Sampling

jduerholt:
This PR adds random seed handling for sampling.

### DIFF
--- a/bofire/strategies/strategy.py
+++ b/bofire/strategies/strategy.py
@@ -26,6 +26,15 @@ class Strategy(ABC):
         self._experiments = None
         self._candidates = None
 
+    def _get_seed(self) -> int:
+        """Returns an integer sampled from the strategies random number generator,
+        that can be used to seed dependent generators.
+
+        Returns:
+            int: random seed.
+        """
+        return int(self.rng.integers(1, 100000))
+
     @classmethod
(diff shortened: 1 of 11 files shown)


bofire/strategies/strategy.py
```diff
@@ -26,6 +26,15 @@ def __init__(
...
         self._candidates = None
 
+    def _get_seed(self) -> int:
+        """Returns an integer sampled from the strategies random number generator,
+        that can be used to seed dependent generators.
+
+        Returns:
+            int: random seed.
+        """
+        return int(self.rng.integers(1, 100000))
```
dlinzner-bcs, line 36: Out of interest: This is just a very large number?

dlinzner-bcs: can you use `sys.maxsize`? https://stackoverflow.com/questions/7604966/maximum-and-minimum-values-for-ints

dlinzner-bcs (approved):
Looks good to me. Only the "large numbers" are arbitrary, so the space seeds for generating random generators will be artifically shrinked. I can however think of no consequence of this :D

bofire/strategies/strategy.py
jduerholt: Yep, it is just a large number, botorch is using the same I think, for this reason, I would stay with this one.

PR #324, opened 2023-12-18
implement batch constraint for doe

KappatC:


jduerholt:
@Osburg did the same here: https://github.com/experimental-design/bofire/pull/322

So with which solution should we go on?

KappatC:
A

> @Osburg did the same here: #322
> 
> So with which solution should we go on?

Oh I did not see that one. Sorry, I thought we agreed we would do the doe constraint after you do the polytope sampler and interpoint constraint, but I might as well missunderstood that. Anyways, I am fine with proceeding with Aaron's implementation, since I am still playing around with a formating that would make pyright and black to not complain.

Osburg:
@KappatC Ah, yes this is possible. Sorry in that case! Was working on the batch constraints this weekend anyways and just continued bc I was already on it...

jduerholt:
So, we close this PR and go forward with the one from @Osburg?

KappatC:
@jduerholt sure, let me play a bit around it first and then I ll close it

Osburg:
> @Osburg nw, thanks for working on this in any case. Maybe in the future we just try to coordinate a bit better, so that we don't do duplicate work
yes, sure, this was my bad! Will watch out better next time!

PR #326, opened 2023-12-20
Add noise prior

jkeupp:
MixedSingleTaskGPs are yet lacking the option to choose a non-default noise prior. This PR adds this functionality.
Are there any additions to test cases needed?


bofire/data_models/surrogates/mixed_single_task_gp.py
```diff
@@ -11,6 +11,11 @@
 )
 from bofire.data_models.surrogates.trainable_botorch import TrainableBotorchSurrogate
 
+from bofire.data_models.priors.api import (
```
jduerholt, line 14: This import block is unsorted. You can just install the ruff extension in vscode and it will do it for you on save.

bofire/data_models/surrogates/mixed_single_task_gp.py
```diff
@@ -20,6 +25,7 @@ class MixedSingleTaskGPSurrogate(TrainableBotorchSurrogate):
     categorical_kernel: AnyCategoricalKernal = Field(
         default_factory=lambda: HammondDistanceKernel(ard=True)
     )
+    noise_prior: AnyPrior = Field(default_factory=lambda: BOTORCH_NOISE_PRIOR())
```
jduerholt, line 28: Regarding the test case, you have to update the spec for `MixedSingleTaskGPSurrogate` in `tests/bofire/data_models/specs/surrogates.py` with the new argument, then it should pass the tests.

jduerholt (changes requested):
Thanks, only two minor things are missing which are mentioned inline.

jduerholt (approved):
Looks good to me

PR #327, opened 2023-12-20
Refactor test suite for data models

jduerholt:
This PR refactors the test suite for the data models, as this was still a mix between Everest and BoFire. In a follow up PR, the rest of the test suite should be tidied up.


dlinzner-bcs (approved):
It will take me days to properly go through this. Can you please quickly summarize what the goal of this refactoring is?

jduerholt:
The problem regarding our test suite is, that it is still a mix between the old everest and the new bofire and that the difference between data models and functional models is not properly reflected. Furthermore, the folder structure of the test suite is not following the actual strucuture in the code. What I did here was a tidy-up in which I copied tests (just within `tests/bofire/data_models` into the correct files and removed duplicate tests. 

A good example is how `test_features.py` looked before this PR, it contained tests regarding the feature containers (`bofire/data_models/domain/features.py`) and the actual features (`bofire/data_models/features/*`). I moved the tests to the correct places. In a next step I want to tidy-up the rest of the test suite, as we also have for example in (`tests/bofire/strategies`) tests that belong to the data models and not to the functional models.

So in general, this PR is not adding anything new, but just copying stuff to the correct places. 

Minor new things are:
- We can test invalid data_models via adding `add_invalid` in `tests/bofire/data_models/specs`. All invalid specs registered there, are tested in `tests/bofire/data_models/test_invalid.py`
- By moving stuff regarding functional aspects out of `tests/bofire/data_models`, we have an easier setup of the minimal installation test suite by just running `pytest /test/bofire/data_models`.
- I splitted `bofire/data_models/base.py` in two files, namely `bofire/data_models/base.py` and `bofire/data_models/filters.py`, as the filter methods are independet from the modified pydantic base class.

bofire/data_models/kernels/continuous.py
```diff
@@ -20,6 +22,12 @@ class MaternKernel(ContinuousKernel):
     nu: float = 2.5
     lengthscale_prior: Optional[AnyPrior] = None
 
+    @validator("nu")
+    def validate_nu(cls, v, values):
```
dlinzner-bcs, line 26: out of interest: what happened before the validator? Does BoTorch only support the `nu = p +1/2` cases?

dlinzner-bcs (approved):
Thank you @jduerholt for the overview. I approve :)

bofire/data_models/kernels/continuous.py
jduerholt: It was catched at the Gpytorch level and throws an error there if it is different from 0.5, 1.5 or 2.5. But we have not catched it in our data model, so it would fail at runtime and not before, which violates our concept ;)

jduerholt:
> Thank you @jduerholt for the overview. I approve :)

Thanks!

PR #328, opened 2023-12-21
Universal constraint sampler

### LINKED ISSUE #314: Sampling under nonlinear constraints
If I have not overlooked something we currently do not have a sampler for nonlinear equality constraints. As @simonsung06 pointed out in #307, the `OptimalityCriterionEnum.SPACE_FILLING` objective only depends on the domain, so I had the idea that we could possibly use this objective to add a sampler supporting a wide range of constraints, including nonlinear equality constraints (and batch constraints as soon as they are supported in doe). Warning: only naive ideas and handwaving arguments are following.
The general idea is to call `find_local_max_ipopt` with space filling objective on the domain in the `_ask` method. This will create a number of points which are "uniformly" distributed over the feasible set, while all constraints that `find_local_max_ipopt` can handle are supported. These points are treated as candidates for the sampling. From these candidates we draw a certain percentage to be returned by `_ask`. Intuitively, the returned candidates should represent a somewhat uniform distribution on the feasible set. The last step is intended to avoid sampling preferably in the boundary region of the feasible set: since the space filling objective will always lead to at least some points lying at the boundary of the feasible set, returning all candidates would destroy the "uniformity" of the distribution. Instead, we create a carpet of candidates populating the feasible set and then randomly pick some of them. 
For domains of high input dimension this will certainly not be a good solution to create samples in reasonable time, but in lower dimensional settings it could be an option. In the small example below we sample 50 points from a spherical surface, I hope this illustrates a little bit what the general idea is. Yellow points are the "carpet of candidates", the black points are a random sample drawn from the candidate set.

@jduerholt (or anyone else interested) What do you think about the idea, does it make any sense to you? Can we even make use of such a sampling method? If you think this could be a good idea feel free to assign this issue to me, then I will take care of it somewhen these days.  

```
domain = Domain(
    inputs=[
        ContinuousInput(key="x1", bounds = (-1,1)),
        ContinuousInput(key="x2", bounds = (-1,1)),
        ContinuousInput(key="x3", bounds = (-1,1))],
    outputs=[ContinuousOutput(key="y")],
    constraints=[NonlinearEqualityConstraint(expression="x1**2 + x2**2 + x3**2 - 1", features=["x1","x2","x3"])],
)

result = find_local_max_ipopt(domain, "linear", objective=OptimalityCriterionEnum.SPACE_FILLING, n_experiments=100, ipopt_options={"maxiter": 100, "disp":3, "linear_solver":"ma57", "hsllib":"libcoinhsl.dylib", "timing_statistics":"yes"})
result.round(3)

def plot_results_3d(result, surface_func):
    u, v = np.mgrid[0 : 2 * np.pi : 100j, 0 : np.pi : 80j]
    X = np.cos(u) * np.sin(v)
    Y = np.sin(u) * np.sin(v)
    Z = surface_func(X, Y)

    fig = plt.figure(figsize=(8, 8))
    ax = fig.add_subplot(111, projection="3d")
    ax.plot_surface(X, Y, Z, alpha=0.1, color="blue")
    ax.plot_surface(X, Y, -Z, alpha=0.1, color="blue")
    ax.scatter(
        xs=result["x1"],
        ys=result["x2"],
        zs=result["x3"],
        marker="o",
        s=40,
        color="orange",
    )
    ax.set(xlabel="x1", ylabel="x2", zlabel="x3")
    ax.xaxis.set_major_formatter(FormatStrFormatter('%.2f'))
    ax.yaxis.set_major_formatter(FormatStrFormatter('%.2f'))

    return ax

ax = plot_results_3d(result, surface_func=lambda x1, x2: np.sqrt(1 - (x1**2 + x2**2)))

sample = result[np.random.rand(len(result)) < 0.5]
ax.scatter(
    xs=sample["x1"],
    ys=sample["x2"],
    zs=sample["x3"],
    marker="o",
    s=40,
    color="black",
)
```

```
Total number of variables............................:      300
                     variables with only lower bounds:        0
                variables with lower and upper bounds:      300
                     variables with only upper bounds:        0
Total number of equality constraints.................:      100
Total number of inequality constraints...............:        0
        inequality constraints with only lower bounds:        0
   inequality constraints with lower and upper bounds:        0
        inequality constraints with only upper bounds:        0


Number of Iterations....: 100

                                   (scaled)                 (unscaled)
Objective...............:  -3.5717406289388549e+01   -3.5717406289388549e+01
Dual infeasibility......:   1.8847170877068415e+00    1.8847170877068415e+00
Constraint violation....:   1.0327337682802806e-06    1.0327337682802806e-06
Variable bound violation:   0.0000000000000000e+00    0.0000000000000000e+00
Complementarity.........:   1.0018139600569745e-11    1.0018139600569745e-11
Overall NLP error.......:   1.8847170877068415e+00    1.8847170877068415e+00


Number of objective function evaluations             = 242
Number of objective gradient evaluations             = 101
Number of equality constraint evaluations            = 242
Number of inequality constraint evaluations          = 0
Number of equality constraint Jacobian evaluations   = 101
Number of inequality constraint Jacobian evaluations = 0
Number of Lagrangian Hessian evaluations             = 0
Total seconds in IPOPT (w/o function evaluations)    =      0.206
Total seconds in NLP function evaluations            =      0.432
```

image

Osburg:
This PR adds a new sampler that supports nonlinear/linear equality/inequality constraints and NChooseKConstraints as discussed in #314. This only works for continuous inputs. In principle also batch constraints (and all other constraints you can compute a jacobian for) are supported, but since the returned sample is sampled from a larger set of candidates, the batch structure is no longer respected. By adapting the sampling logic in this last step this method can be made available for batch constraints. Closes #314.


jduerholt:
Hi @Osburg, 

Thanks for the PR, I will have a look over the Christmas days. Sorry for the delay.

Best,

Johannes

bofire/data_models/strategies/samplers/universal_constraint.py
```diff
@@ -0,0 +1,47 @@
...
+
+    Attributes:
+        domain (Domain): Domain defining the constrained input space
+        sampling_fraction (float, optional): Fraction of sampled points to total points generated in
+            the sampling process. Defaults to 0.3.
+        ipopt_options (dict, optional): Dictionary containing options for the IPOPT solver. Defaults to {"maxiter":200, "disp"=0}.
+    """
+
+    type: Literal["UniversalConstraintSampler"] = "UniversalConstraintSampler"
+    sampling_fraction: float = 0.3
```
jduerholt, line 29:
```suggestion
    sampling_fraction: Annotated[float, Field(gt=0, lt=1)] = 0.3
```

By this we have direct validation that the sampling fraction is between 0 and 1 as soons as we switch to pydantic2 which will be soon ;)

bofire/data_models/strategies/samplers/universal_constraint.py
```diff
@@ -0,0 +1,47 @@
...
+    ipopt_options: dict = {"maxiter": 200, "disp": 0}
```
jduerholt, line 30: We will or have already these ipopt options dictionary for the DoEStrategy, would it make sense to model this also as a pydantic class, in which all possible options together with its defaults are listed?

jduerholt (commented):
Looks overall good to me, just one question: is it on purpose that you inherit from the `SamplerStrategy`, because then you also inherit from their the handling of NChooseKs, which is totally fine for me, but I just wanted to ask ;)

bofire/data_models/strategies/samplers/universal_constraint.py
Osburg: Hm, good question... I think most of the options that are available (here)[https://coin-or.github.io/Ipopt/OPTIONS.html] are only interesting for expert users, so I do not know for how many people this is relevant and if it is worth it to do the effort. Most people will probably only use the `maxiter` and `disp` options.  
However, for using the HSL linear solvers you have to set some options depending on your operating system, which can be annoying. The HSL linear solvers enormously reduce the computation time for large problems. If you think many users are interested in generating large models AND are willing to take the extra effort to install nondefault linear solvers, it might be worth it to handle the linear solver selection automatically in some way... What do you think? :D

Osburg:
> is it on purpose that you inherit from the `SamplerStrategy`, because then you also inherit from their the handling of NChooseKs, which is totally fine for me, but I just wanted to ask ;)

Thanks for pointing this out :) This was not my intention, I did not pay attention apparently. So I guess I should let `UniversalConstraintSampler` inherit from `Strategy` instead?

jduerholt:
Yep, I would then recommend to inherit from `Strategy`.

bofire/data_models/strategies/samplers/universal_constraint.py
jduerholt: Hmm, I think I would let it as it is, and if necessary model it in a better way, what do you think?

Osburg: Agree.

bofire/strategies/samplers/universal_constraint.py
```diff
@@ -0,0 +1,58 @@
...
+        data_model: DataModel,
+        **kwargs,
+    ):
+        super().__init__(data_model=data_model, **kwargs)
+        assert data_model.sampling_fraction > 0 and data_model.sampling_fraction <= 1
+        self.sampling_fraction = data_model.sampling_fraction
+        self.ipopt_options = data_model.ipopt_options
+
+    def _ask(self, candidate_count: int) -> pd.DataFrame:
+        samples = find_local_max_ipopt(
```
jduerholt, line 32: We could also include pending candidates here, or? Meaning that we provide already done experiments to the sampler to which the distance should be maximized? Those would be available via `self.candidates` via the `fixed_experiments` keyword of `find_local_max_ipopt`. What do you think?

jduerholt (changes requested):
Hi @Osburg,

thanks for the update, only some minor issues remain. We could also implement the `pending_candidates` at a later point in time. But this should not be too complicated. We can do it as you want. 

Best,

Johannes

bofire/strategies/samplers/universal_constraint.py
Osburg: I assumed candidates means we do not necessarily need to have them in the set of samples (so we do not sample pending candidates w/ higher priority. Is this right?)

jduerholt: Hi Aaron,

regarding the pending candidates, the idea is the following:

We provide a set of pending candindates, and want to sample `n` new candidates which are optimal given the objective and the set of pending ones.  The `find_local_max_ipopt` returns also the fixed ones if you provide them. So you have to increase the number of samples to get from `find_local_max_ipopt` and remove then the pending ones from the dataframe before you do the random sampling.

Do you see what I mean? Can you then also add a test where pending ones are used?

Best,

Johannes

Osburg: Hi Johannes, 

thx for the clarification! I think in the new commit it is done the right way.

Cheers,
Aaron :)

jduerholt (approved):
Looks good to me! Thank you very much!

PR #332, opened 2024-01-03
Compatibility PR for Formulaic 1.0.1

jduerholt:
Over the christmas period, formulaic 1.0.1 was released, which was breaking our tests. This PR fixes it.

cc: @Osburg, I am currently not able to assign you as a reviewer, but would be nice if you also can have a look ;)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ setup(
             "botorch>=0.9.5",
             "multiprocess",
             "plotly",
-            "formulaic>=0.6.0",
+            "formulaic>=1.0.1",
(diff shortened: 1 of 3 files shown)


Osburg:
Hey @jduerholt looks good to me. If I understood correctly formulaic just handles `{...}` model terms a little differently now (and does not remove whitespaces in these model terms anymore).

jduerholt:
Yep, in formulaic before 1.0.1 `x**2` was handled as `x**2`, now it is `x ** 2`. I will then just merge it in ;)

Issue #319, opened 2023-12-12
Implement LSR-BO


jduerholt:
Implement the LSR-BO algorithm, outlined in the papers below which is very helpful for closed loop operation:

- https://www.merl.com/publications/docs/TR2023-122.pdf
- https://www.merl.com/publications/docs/TR2023-057.pdf

jduerholt:
This paper by @DavidWalz @R-M-Lee @bertiqwerty is related to it: https://arxiv.org/abs/2312.00622

PR #338, opened 2024-01-22
LSR-BO

jduerholt:
This PR implements the LSR-BO methods implemented in https://github.com/experimental-design/bofire/issues/319 and creates an infrastructure to implement new related methods by toggling between local and global optimization in `_ask`.

In addtion the `ShortestPathStrategy` is implemented with solves a quadratic program to get the shortest path under linear constraints to get from start to end.

Closes https://github.com/experimental-design/bofire/issues/319


bofire/strategies/predictives/botorch.py
```diff
@@ -267,6 +293,85 @@ def _postprocess_candidates(self, candidates: Tensor) -> pd.DataFrame:
         preds = self.predict(df_candidates)
         return pd.concat((df_candidates, preds), axis=1)
 
+    def _optimize_acqf_continuous(
+        self,
+        candidate_count: int,
+        acqfs: List[AcquisitionFunction],
+        bounds: Tensor,
+        ic_generator: Callable,
```
bertiqwerty, line 301: Is `ic` inequality constraint? Maybe a more verbose variable name?

bofire/strategies/predictives/botorch.py
```diff
@@ -267,6 +293,85 @@ def _postprocess_candidates(self, candidates: Tensor) -> pd.DataFrame:
...
+        ic_gen_kwargs: Dict,
+        nchooseks: List[Callable[[Tensor], float]],
+        fixed_features: Optional[Dict[int, float]],
```
bertiqwerty, line 304: What are `fixed_features` and `fixed_features_list`? Are these features not to be optimized? Why are there two of those? For batch proposals?

bofire/strategies/predictives/botorch.py
```diff
@@ -326,71 +431,58 @@ def _ask(self, candidate_count: int) -> pd.DataFrame:
 
         (
             bounds,
```
bertiqwerty, line 433: `bounds` are the global bounds in contrast to local bounds, right?

bofire/strategies/predictives/botorch.py
```diff
@@ -326,71 +431,58 @@ def _ask(self, candidate_count: int) -> pd.DataFrame:
...
-                    inequality_constraints=get_linear_constraints(
-                        domain=self.domain, constraint=LinearInequalityConstraint  # type: ignore
-                    ),
-                    nonlinear_inequality_constraints=nchooseks,
-                    fixed_features_list=fixed_features_list,
-                    ic_generator=ic_generator,
-                    ic_gen_kwargs=ic_gen_kwargs,
-                )
+            if self.local_search_config.is_local_step(
+                local_acqf_val.item(), global_acqf_val.item()
```
bertiqwerty, line 470: `is_local_step` in the current implementation does only make use of the local optimum, right? Couldn't we save some computation time by not always compute the global optimum?

bofire/strategies/shortest_path.py
```diff
@@ -0,0 +1,100 @@
...
+                feat = inputs.get_by_key(key)
+                assert isinstance(feat, ContinuousInput)
+                if feat.is_fixed():
+                    b[i] -= feat.fixed_value()[0] * coef  # type: ignore
+                else:
+                    A[i, keys.index(key)] = coef
+        return A, b
+
+    def step(self, start: pd.Series) -> pd.Series:
+        inputs = self.continuous_inputs
```
bertiqwerty, line 52: Can you write a doc-string for this method?

bofire/strategies/shortest_path.py
```diff
@@ -0,0 +1,100 @@
...
+        lower, upper = inputs.get_bounds(
```
bertiqwerty, line 53: Not sure, but would it make sense to get rid of the equality constraint by using the reduced domain? Then you would need compute the intersection between the boundary of the box and the connection between start and end which does not involve an optimization problem?

bertiqwerty (commented):
Thanks a lot, Johannes. Very useful method! We have some questions, see the review comments.

bofire/data_models/domain/features.py
```diff
@@ -537,6 +538,10 @@ def get_bounds(
             experiments (Optional[pd.DataFrame], optional): Dataframe with input features.
                 If provided the real feature bounds are returned based on both the opt.
                 feature bounds and the extreme points in the dataframe. Defaults to None,
+            reference_experiment (Optional[pd.Serues], optional): If a reference experiment provided,
```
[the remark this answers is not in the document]
jduerholt: Because it is already in the signature of the function, or what do you mean? Problem is somehow that we have this in most of our docstrings. How to proceed then there in general?

bofire/data_models/features/continuous.py
```diff
@@ -20,7 +21,10 @@ class ContinuousInput(NumericalInput):
     type: Literal["ContinuousInput"] = "ContinuousInput"
     order_id: ClassVar[int] = 1
 
-    bounds: Tuple[float, float]
+    bounds: Tuple[float, float] = Field(description="mama")
```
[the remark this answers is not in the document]
jduerholt: Oh, I was just testing the use of the description, and forget to remove it. I will remove it and also create an issue regarding docstrings and data model description fields, so that we can define a standard.

bofire/data_models/features/continuous.py
```diff
@@ -20,7 +21,10 @@ class ContinuousInput(NumericalInput):
...
+    local_relative_bounds: Tuple[
+        Annotated[float, Field(gt=0)], Annotated[float, Field(gt=0)]
+    ] = (math.inf, math.inf)
```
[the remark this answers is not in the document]
jduerholt: I was using `math.inf` because this belongs to the standard python library and to prevent to get serialization problems with pydantic when using a numpy type. No it has to be (`inf`, `inf`) as we are calculating the local bounds by `(max(current_val - local_relative_bounds[0], bounds[0]), min(current_val+local_relative_bounds[1]), bounds[1])`

bofire/strategies/predictives/botorch.py
jduerholt: No `ic_generator` means initial conditions generator, and that is the name for it that is also used in the botorch api. For this reason, I would like to keep it ;)

bofire/strategies/predictives/botorch.py
jduerholt: `fixed_features` is used in  botorchs `optimize_acqf` when you just optimize over a continuous space. It is a dictionary that maps indices to float values. The indices mark the features that should be kept fixed. And the floats are the respective fixed values. In case that you optimize over a mixed space using `optimize_acqf_mixed`, you have to provide a list of fixed_features dictionaries. It is then running an optimization for every dictionary in the list and returns the best one. It is documented here: https://github.com/pytorch/botorch/blob/b5b4fb7c316983f0cfb2f3257247e52c71a23b7d/botorch/optim/optimize.py#L417 and here https://github.com/pytorch/botorch/blob/b5b4fb7c316983f0cfb2f3257247e52c71a23b7d/botorch/optim/optimize.py#L855

bofire/strategies/predictives/botorch.py
jduerholt: Yes!

bofire/strategies/predictives/botorch.py
jduerholt: In principle yes, and I also thought about it, but I discarded this idea, because I already have ideas regarding improvements of the method where also the global value is used, and I coded it up in a way that we can easily plug and play these new methods, by just providing it with a new `LocalSearchConfig` class and a new `is_local_step` method and it just runs without the need for touching the actual optimization code. For this reason, I traded computation time for versatility.

bofire/strategies/shortest_path.py
jduerholt: Will do it.

bofire/strategies/shortest_path.py
jduerholt: Hmm, interesting idea. But what is with linear inequalities? We also have to account for them, or? Or do inequalities do not cause a problem? I am not sure. And formulating it as an optimization problem would give the possibility for more complex local constraints which are not just rectangular. What do you think? Keep it or reformulating it?

bofire/data_models/domain/features.py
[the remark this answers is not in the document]
bertiqwerty: Exactly. We already have it as type hint. I would not repeat it in the docstring. I would never repeat it in the docstring. We do not have to change this for everything right now. I hope that the doc-extraction-tool makes use of the type hints.

bofire/strategies/shortest_path.py
jduerholt: Done, since Monday, I have copilot access for a trial period. It is super powerful for documentation.

bofire/data_models/features/continuous.py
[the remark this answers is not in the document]
bertiqwerty: `np.inf` and `math.inf` are both of type `float` and `np.inf == math.inf` is `True`. Serialization should be fine either way. We would just save one import-statement.

bofire/strategies/predictives/botorch.py
bertiqwerty: Thanks for the Botorch-Nachhilfe :)

bertiqwerty (approved):
Thanks again Johannes! I am really grateful that we have this method now. I will approve this and you can fix/merge as you want. No show-stoppers from my side :).

bofire/strategies/shortest_path.py
jduerholt: I thought more about it, and want to keep it as optimization problem, as it is also not straight forward with the reduced domain.

PR #342, opened 2024-01-24
fix bug in nchoosek_constraints_as_bounds

### LINKED ISSUE #341: DoE with NChooseK
Hi @Osburg,

I discovered a strange behavior when setting up the following DoE including NchooseKs:

```python
from bofire.data_models.features.api import ContinuousInput
from bofire.data_models.constraints.api import LinearEqualityConstraint, LinearInequalityConstraint, NChooseKConstraint
from bofire.data_models.domain.api import Domain, Inputs, Constraints
from bofire.utils.doe import get_confounding_matrix
from bofire.data_models.strategies.api import  PolytopeSampler
import bofire.strategies.api as strategies
from bofire.strategies.doe.design import find_local_max_ipopt
import matplotlib.pyplot as plt



domain = Domain(
    inputs=Inputs(
        features = [
            ContinuousInput(key="i_1",bounds=(0,55)),
            ContinuousInput(key="i_2",bounds=(0,55)),
            ContinuousInput(key="c",bounds=(15,55)),
            ContinuousInput(key="a",bounds=(5,40)),
            ContinuousInput(key="p_1",bounds=(0,5)),
            ContinuousInput(key="p_2",bounds=(0,5)),
        ]),
        constraints = Constraints(
            constraints=[
                LinearEqualityConstraint(features=["i_1","i_2","c","a","p_1","p_2"], coefficients=[1,1,1,1,1,1,], rhs=100),
                LinearInequalityConstraint(features=["i_1", "i_2"], coefficients=[-1,-1], rhs=-15),
                LinearInequalityConstraint(features=["p_1", "p_2"], coefficients=[-1,-1], rhs=-0.5),
                LinearInequalityConstraint(features=["c","i_1", "i_2"], coefficients=[1,1,1], rhs=92),
                NChooseKConstraint(features=["i_1", "i_2"], min_count=0, max_count=1, none_also_valid=False),
                NChooseKConstraint(features=["p_1", "p_2"], min_count=0, max_count=1, none_also_valid=False),
            ]
        )
)

candidates = find_local_max_ipopt(domain,"linear")# sampling=sampling)

fig, axes = plt.subplots(ncols=4, figsize=(12, 4))
for i, combo in enumerate([["i_1", "p_1"],["i_2", "p_2"], ["i_1", "p_2"], ["i_2", "p_1"]]):
    axes[i].scatter(candidates[combo[0]], candidates[combo[1]])
    axes[i].set_xlabel(combo[0])
    axes[i].set_ylabel(combo[1])
plt.show()
```

What we are having here is a mixture design, in which we have two groups of two (`i` group with `i_1` and `i_2` and `p` group with `p_1` and `p_2`). Out of each group always one ingredient has to be used. For this reason, I set the lower bound on the features to zero, and introduced per group a linear inequality constraint to have the proper lower bound and an NChooseK constraint.

If I now plot the results, one gets for example this:

image

One can see that for example the combination high `i_1` and high `p_1` is never tried, which is really strange because then one introduces an artifical correlation between this degrees of freedom.

In the following, I setup the same design but without the subgroups:

```python
domain = Domain(
    inputs=Inputs(
        features = [
            ContinuousInput(key="i_1",bounds=(15,55)),
            ContinuousInput(key="c",bounds=(15,55)),
            ContinuousInput(key="a",bounds=(5,40)),
            ContinuousInput(key="p_1",bounds=(0.5,5)),
        ]),
        constraints = Constraints(
            constraints=[
                LinearEqualityConstraint(features=["i_1","c","a","p_1",], coefficients=[1,1,1,1], rhs=100),
                LinearInequalityConstraint(features=["c","i_1"], coefficients=[1,1], rhs=92),
            ]
        )
)

candidates = find_local_max_ipopt(domain,"linear")# sampling=sampling)

fig, ax = plt.subplots()
ax.scatter(candidates["i_1"], candidates["p_1"])
plt.show()

```

Here I get the following:
image

One sees that when setting it up without NchooseKs it combines high values of both with each other.

Do you have any idea for this strange behavior? 

One idea that I had was that it is affected by how the starting points for the ipopt are generated, currently we sample initial point by using the `PolytopeSampler`, which returns already candidates that obey NChooseK constraints. And afterwards, the DOE module put its bounds randomly on top, without taking into account on which positions the constraints are already fulfilled in the starting points. I think we should change it.  What do you think? But providing initial samples by hand that do no obey to the NchooseKs, does not help either.

Would be really happy to get your input on this!

cc @dlinzner-bcs

Osburg:
Closes #341

### DIFF
--- a/bofire/strategies/doe/utils.py
+++ b/bofire/strategies/doe/utils.py
@@ -532,7 +532,7 @@ def nchoosek_constraints_as_bounds(
                     for i in range(n_experiments):
                         ind_vanish = ind[i % len(ind)]
                         bounds[ind_vanish + i * len(domain.inputs), :] = [0, 0]
-                        if (i % len(ind) == len(ind) - 1):
+                        if i % len(ind) == len(ind) - 1:
                             np.random.shuffle(ind)
     else:
         pass
--- a/bofire/strategies/doe/utils.py
+++ b/bofire/strategies/doe/utils.py
@@ -532,6 +532,8 @@ def nchoosek_constraints_as_bounds(
                     for i in range(n_experiments):
                         ind_vanish = ind[i % len(ind)]
                         bounds[ind_vanish + i * len(domain.inputs), :] = [0, 0]
+                        if (i % len(ind) == len(ind) - 1):
+                            np.random.shuffle(ind)
     else:
         pass
 


jduerholt (commented):
Looks good to me. Thx!

jduerholt (approved):

PR #346, opened 2024-01-25
Random search with nonlinear inequality constraints

bertiqwerty:
The data model was missing the nonlinear inequality constraint

### DIFF
--- a/bofire/data_models/strategies/random.py
+++ b/bofire/data_models/strategies/random.py
@@ -7,6 +7,7 @@ from bofire.data_models.constraints.api import (
     LinearInequalityConstraint,
     NChooseKConstraint,
 )
+from bofire.data_models.constraints.nonlinear import NonlinearInequalityConstraint
 from bofire.data_models.features.api import Feature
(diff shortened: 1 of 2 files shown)


jduerholt:
Just a side node, the `UniversalSampler` is also able to deal with NonlinearConstraints, maybe we should also integrate this one into the RandomStrategy. What do you think?

bertiqwerty:
> Just a side node, the UniversalSampler is also able to deal with NonlinearConstraints, maybe we should also integrate this one into the RandomStrategy. What do you think?

What is the use case?

jduerholt (approved):
Looks good to me. Thx!

bertiqwerty:
> > What is the use case?
> 
> It uses the Spacefilling criterion from the DoE package, as it is not fitting to the DoE Strategy, Aron made a seperate strategy for it.

Could also be part of Random. Would be easier to find, I guess.

jduerholt:
> Could also be part of Random. Would be easier to find, I guess.

Will keep it in mind, when I refactor random and the samplers in the next weeks.

PR #347, opened 2024-01-30
Refactor random strategy

jduerholt:
This PR refactors the random strategy: instead of using either polytope sampling or rejection sampling, it uses now a combination of both which allows for the handling of more constraint types by the random strategy. 

Docstrings will be also updated.


bofire/strategies/random.py
```diff
@@ -21,25 +23,52 @@ def __init__(
...
-                errors.append(err)
-        if self.sampler is None:
-            raise Exception(errors)
+        self.num_base_samples = data_model.num_base_samples
+        self.max_iters = data_model.max_iters
+        self.fallback_sampling_method = data_model.fallback_sampling_method
+        self.n_burnin = data_model.n_burnin
+        self.n_thinning = data_model.n_thinning
+
+    def _get_feasible_domain_for_polytope_sampler(self) -> Domain:
```
bertiqwerty, line 32: Make this a free function with `domain: Domain` as input parameter for better encapsulation. It does not need access to all of `self`.

bofire/data_models/strategies/api.py
```diff
@@ -2,6 +2,7 @@
 
 from bofire.data_models.strategies.doe import DoEStrategy
 from bofire.data_models.strategies.factorial import FactorialStrategy
+from bofire.data_models.strategies.polytope import PolytopeSampler
```
bertiqwerty, line 5: Since there are now also strategies that are called `Sampler` the naming becomes somewhat inconsistent, no? `RandomStrategy` is also just a sampling.

bertiqwerty (approved):
Thanks Johannes. I left 2 comments. Please have a quick look and merge as you like.

bofire/data_models/strategies/api.py
jduerholt: Fully agree, but I do not have a proper solution, how would you do the naming?

bertiqwerty: Hmm... The easiest would be that all are strategies. Which they are. With Samplers we probably mean not-predictive, right? Then we would call DoE a sampler and Random also a sampler. Depends on what sampler actually means.

bofire/strategies/random.py
jduerholt: Done.

bofire/data_models/strategies/api.py
jduerholt: Yep, it means non-predictive. And something like `PolytopeStrategy` sounds weird ..., I think I will merge this PR first and we create an issue to solve this naming problem, ok? Or perhaps just schedule a meeting ...

PR #348, opened 2024-02-05
Multilinear constraint

jduerholt:
This PR implements multilinear constraints of the form `sign * x1**e1 * x2**e2 * ... * xn**en <= rhs`. In addition, it simplifies the composition of equality and inequality constraints by introducing two helper classes. Furthermore a `types.py` is introduced.

### DIFF
--- a/bofire/data_models/constraints/linear.py
+++ b/bofire/data_models/constraints/linear.py
@@ -1,14 +1,15 @@
-from typing import List, Literal, Tuple
+from typing import Annotated, List, Literal, Tuple
 
 import numpy as np
 import pandas as pd
-from pydantic import field_validator, model_validator
+from pydantic import Field, model_validator
 
 from bofire.data_models.constraints.constraint import (
-    Coefficients,
-    FeatureKeys,
+    EqalityConstraint,
+    InequalityConstraint,
     IntrapointConstraint,
 )
+from bofire.data_models.types import FeatureKeys
 
 
 class LinearConstraint(IntrapointConstraint):
@@ -23,16 +24,8 @@ class LinearConstraint(IntrapointConstraint):
     type: Literal["LinearConstraint"] = "LinearConstraint"
 
     features: FeatureKeys
-    coefficients: Coefficients
+    coefficients: Annotated[List[float], Field(min_length=2)]
     rhs: float
-
-    @field_validator("features")
-    @classmethod
-    def validate_features_unique(cls, features):
-        """Validate that feature keys are unique."""
-        if len(features) != len(set(features)):
-            raise ValueError("features must be unique")
-        return features
 
     @model_validator(mode="after")
     def validate_list_lengths(self):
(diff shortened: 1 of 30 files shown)


bofire/data_models/constraints/multilinear.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+class MultiLinearConstraint(IntrapointConstraint):
+    """
+    Represents a multi-linear constraint of the form `sign * x1**e1 * x2**e2 * ... * xn**en`.
+
+    Attributes:
+        type (str): The type of the constraint.
+        features (FeatureKeys): The keys of the features used in the constraint.
+        exponents (List[float]): The exponents corresponding to each feature.
```
bertiqwerty, line 22: Exponents are of type float but I guess you assume to only receive integers because otherwise it wouldn't be multi-linear anymore, right? Can you do `List[int]` as type hint? Or document it at least somehow?

bofire/data_models/domain/domain.py
```diff
@@ -150,11 +151,12 @@ def validate_constraints(self):
...
         keys = self.inputs.get_keys()
-        for c in self.constraints:
-            if isinstance(c, LinearConstraint) or isinstance(c, NChooseKConstraint):
-                for f in c.features:
-                    if f not in keys:
-                        raise ValueError(f"feature {f} in constraint unknown ({keys})")
+        for c in self.constraints.get(
+            [LinearConstraint, NChooseKConstraint, MultiLinearConstraint]
+        ):
+            for f in c.features:  # type: ignore
```
bertiqwerty, line 157: Why do you need to ignore this type while you didn't need to before?

bertiqwerty (approved):
Thanks Johannes! Have a look at the comments, please, and merge as you like.

bofire/data_models/constraints/multilinear.py
jduerholt: Hmm, it also works with floats, so I would like to keep it as `List[float]`, do you have a better name for it?

bertiqwerty: Hmmm... it's a polynomial with real exponants, right? I just had a look. There seems to be a name for that but I have not heard it before.
https://en.wikipedia.org/wiki/Posynomial

jduerholt: Yep, only product, for me it is somehow a generalized multilinear one ..., so is this a problem for you to keep it?

bertiqwerty: Giving it another thought, I would simply call it the `ProductConstraint`. I think, multi-linear means it is linear in each variable which this is something different.

jduerholt: Done.

bofire/data_models/constraints/linear.py
```diff
@@ -79,36 +65,13 @@ class LinearEqualityConstraint(LinearConstraint):
 
     type: Literal["LinearEqualityConstraint"] = "LinearEqualityConstraint"
 
-    # def is_fulfilled(self, experiments: pd.DataFrame, complete: bool) -> bool:
-    #     """Check if the linear equality constraint is fulfilled for all the rows of the provided dataframe.
-
-    #     Args:
-    #         df_data (pd.DataFrame): Dataframe to evaluate constraint on.
+    # def is_fulfilled(self, experiments: pd.DataFrame, tol: float = 1e-6) -> pd.Series:
```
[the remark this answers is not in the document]
jduerholt: Good catch!

bofire/data_models/domain/domain.py
jduerholt: It is needed here, because there are constraint types which do not have a features attribute.

PR #352, opened 2024-02-10
Refactor Random Strategy II

jduerholt:
In this PR the `PolytopeSampler` Strategy is removed and completely integrated into the RandomStrategy, furthermore `UniversalConstraintSampler` is renamed into `SpaceFillingStrategy` as discussed on Friday.


jduerholt:
@dlinzner-bcs @Osburg `test_n_zero_eigvals_constrained` is failing for `fully-quadratic` and I have no idea why. The sampling preserves the equality constraint and still it is failing but only for `fully-quadratic`. Any idea why?

jduerholt:
I investigated this a bit more, but I have no idea what is going on here :D

Osburg:
Hi @jduerholt, 
I think the problem lies in the sampling itself. In `RandomStrategy._sample_from_polytope()` we call `sample()` for each categorical feature with the same seed.
```
        for feat in domain.get_features([CategoricalInput, DiscreteInput]):
            samples[feat.key] = feat.sample(n, seed=seed)  # type: ignore
```
I think we should not do this since it creates regular patterns in the drawn values between the features.
In order to fix the test, incrementing the seed by one in each loop iteration does the trick for me (and should still lead to reproducable behavior). 
```
        for feat in domain.get_features([CategoricalInput, DiscreteInput]):
            seed += 1
            samples[feat.key] = feat.sample(n, seed=seed)  # type: ignore
```
I guess similar (possibly undetected) issues may occur in other places where sampling methods are called multiple times with the same seed.

jduerholt:
@Osburg, thank you very much for the catch! I overlooked it when refactoring the random strategy!

jduerholt:
@bertiqwerty: it is fixed now and ready for review!

bertiqwerty (approved):
Thanks, Johannes.

PR #353, opened 2024-02-21
Initial attempt to incorporate MultiTask GPs

jpfolch:
Initial attempt to incorporate MultiTask GPs, with long term plan of incorporating the multi-fidelity algorithm as described in [here](https://www.sciencedirect.com/science/article/pii/S0098135423000637). 

# Current issues:

- When predicting the posterior, BoFire by default returns the posterior mean and a co-variance which included the observation noise (see this [line](https://github.com/experimental-design/bofire/blob/7d6af939948d7bf30fd0d5327acc02bdfbaf86e7/bofire/surrogates/botorch.py#L26) where `observation_noise = True`). However, BoTorch's `MultiTaskGP` does not support adding observation noise yet and returns `NotImplementedError: Specifying observation noise is not yet supported by MultiTaskGP.` (a) A work-around would be to redefine the `MultiTaskGPSurrogate` to have it set to `False`, however this would lead to inconsistencies in model usage within BoFire. (b) Maybe a new flag can be incorporated into BoFire to include the option of predicting observation noise. (c) For MultiTaskGPs I could implement the new posterior myself, by adding the observation noise to the posterior covariances manually; and then simplify the code once BoTorch starts supporting observation noise.

- I am unable to use the dump functionality, when saving the BoTorch model using `torch.save(self.model, buffer)` I get the error `AttributeError: Can't pickle local object 'IndexKernel.__init__.<locals>.<lambda>'`. Still not sure if this is an issue with my local code, with GPyTorch's Index Kernel, or something else entirely.

- Need some way of specifying which column has the task_id for each observation. Currently done by calling the variable 'fid' (for fidelity, but easily changed). Need to decide whether we specify which column has the task_ids by specific name or with some other method.

- LKJ Prior (a matrix-valued prior on the inter-task correlations) depends on the number of tasks, which is unknown until the model is initialized. Currently I am setting it as a dummy `n_tasks = 1` in the prior and then changing the attribute value when initializing the surrogate, it works but perhaps there is a better way of going about it. 

# Future work:

- Need to incorporate a likelihood with heteroskedastic noise that allows for different noise levels in each task.

- Once MultiTaskGPs are working, implement a multi-fidelity strategy for querying.


jduerholt (changes requested):
Hi @jpfolch,

thank you very much for your PR. MultiTask and MultiFidelityGPs are something which I wanted to have in `BoFire` for already a long time, but never find time for!

For the start I am focusing on your point three (how to encode the task info), as I have already an idea for this ;)

My proposal is to introduce a new input feature called `TaskInput`. This `TaskInput` should inherit from the `CategoricalInput` feature and should have an additional attribute called `fidelities` of type `List[int]` in which one can assign different fidelity levels to the individual tasks/categories. I would add a validator that checks that one always has a step size of 1 between different fidelities and that it starts with zero. 

For example, having a `TaskInput` with three different tasks/categories named `process_1`, `process_2` and `process_3` the following fidelities would be valid:

- `[0,0,0]`: all have the same fidelity level
- `[0,0,1]`: `process_3` has the highest fidelity

Via the `allowed` attribute of the `TaskInput` one can then define which tasks can be proposed by the optimizer. As the `TaskInput` would be a subclass of the `CategoricalInput`, one has to setup the encoding for the GP properly. Currently, `CategoricalInput`s are alway One Hot encoded, either one gives the option for ordinal encoding for `TaskInputs`s or one does still the OneHot encoding and uses `OneHotToNumeric` input transform for the `TaskInput` as here: https://github.com/experimental-design/bofire/blob/main/bofire/surrogates/mixed_single_task_gp.py which would be from my perspective the most elegant way, but let me know what you think ;)

In addtion, we should add a validator for the `Inputs` object to forbid more than one `TaskInput`.

What do you think? I will answer your other points over the weekend (hopefully).

Best,

Johannes

tutorials/multi_task_gp_testing.py
```diff
@@ -0,0 +1,51 @@
+import json
+
+from pydantic import parse_obj_as
```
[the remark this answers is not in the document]
jpfolch: Yes, the plan is to remove it completely. It just there for now since it reproduces the errors I wrote about.

jduerholt:
> And I really like the `TaskInput` solution. I'll get started on it and push the changes.

Great, I will think in the meantime about the other open questions.

bofire/data_models/features/tasks.py
```diff
@@ -0,0 +1,32 @@
+from typing import List, Literal
+
+import numpy as np
+from pydantic import model_validator, validator
+
+from bofire.data_models.features.api import DiscreteInput
+
+
+class TaskInput(DiscreteInput):
```
jduerholt, line 9: Short question: why are you basing it on the `DiscreteInput` and not on the `CategoricalInput`? I would prefer the `CategoricalInput`, as it is more expressive and the Task is somehow more a categorical quantity than a discrete one, as the `DiscreteInput` always asumes an ordinal ordering of its values.

bofire/data_models/features/tasks.py
```diff
@@ -0,0 +1,32 @@
...
+    type: Literal["TaskInput"] = "TaskInput"
+    n_tasks: int
```
jduerholt, line 11: `n_tasks` is not necessary if you base it on the CategoricalInput as it would be just the number of categories.

bofire/data_models/features/tasks.py
jpfolch: I thought about it and here is my logic:

(a) The task themselves do not have any ordering

(b) But the GP expects an ordering of the tasks corresponding to the location of each specific task in the IndexKernel function, i.e. inputs to kernel are `k((x', i), (x, j)) ` so the task input values should always correspond to values `i, j in [0, ..., n_tasks - 1]`. 

This means that all we need to do to initialize the object InputTasks is to select the number of tasks. I am happy to change this though, and we can just encode the Categorical variables before feeding it to the GP.

jpfolch:
I've now got an implementation of `TaskInputs` inheriting from `CategoricalVariable` and it seems to be working fine when it comes to training the model and fitting. A few issues remain:

1. Using `OneHotToNumeric` has to be done outside `BoTorch`'s model since the first line of calling the posterior on `MultiTaskGPytorchModel` is `includes_task_feature = X.shape[-1] == self.num_non_task_features + 1` to check if the task features are included in the X (i.e. fully expecting the task_id to be a singular values in a column). It does not apply the input transform until after this, so under one-hot-encoding the check fails and the model returns predictions for all tasks for all inputs. Two ways of fixing this (a) try to fix within BoTorch and see if they accept the changes, or (b) change the transform within BoFire, or (c) using transform before calling `posterior` and before setting the `training_data` (current code is doing this).

2. I am running into issues with pydantic and serialization. When running `surrogate_data.model_dump_json()` I get the following warning: `Expected Union[definition-ref, definition-ref, definition-ref, definition-ref, definition-ref, definition-ref, definition-ref, definition-ref] but got TaskInput - serialized value may not be as expected`. I think this is the causing the following problem where `parse_obj_as(MultiTaskGPSurrogate, json.loads(jspec))` fails to run as it fails verification by giving the wrong arguments to the wrong classes. Any help on how to fix this would be appreciated.

bofire/data_models/features/tasks.py
jpfolch: It is now using `CategoricalInput` as a base

jduerholt:
Hi @jpfolch,

as already discussed, you can find a proper implementation for the `TaskInput` feature here https://github.com/experimental-design/bofire/pull/360 which should be merged soon. Just use this one.

Regarding your point with the one-hot input transform: I did not know this and this is a pity. Including it into botorch will need some time (as I know from experience), but your solution will not work with an optimizer as the optimizer will just call `posterior`. For this reason, we have to use an `OrdinalEncoding` for the categorical `TaskInput`. 

This is implemented here: https://github.com/experimental-design/bofire/blob/4790d637016f4e482ae8bafa6cb711b66e26b99e/bofire/data_models/features/categorical.py#L245

The actual encoding taken into account by the model is based on the attribute `input_preprocessing_specs` which is validated/generated here: https://github.com/experimental-design/bofire/blob/ccb0bc2998f35ede3118b490e0ce387244898305/bofire/data_models/surrogates/surrogate.py#L21

For the MultiTaskGP we then have to overwrite it in a way that we assign `CategoricalEncodingEnum.ORDINAL` for the task feature. If you do this, you should already be able to use the model. For using it in optimization, we then also have to so some adustments. But this will not be a problem.

So, I would recomment to merge main in and change the model to ORDINAL Encoding for the task feature. Note that in standard models, it should still use ONEHOT. You will also need to adjust the input scaler to ignore the task feature, so that it stays as integer.

Best,

Johannes

jpfolch:
Perfect, thanks! I will get started on fixing the encoding for TaskInputs. 

Regarding not being able to pickle Index kernels, @TobyBoyne found the bug within GPyTorch and submitted a [PR](https://github.com/cornellius-gp/gpytorch/pull/2482/commits/721f906fff63625f7fda4c5954f8bdb4e20e3745) which has been approved, so it should be fixed eventually, but we will need a temporary fix for the next few months at least.

TobyBoyne:
Hey, while you're waiting for the patch to hit GPyTorch, you can fix this in BoFire by manually registering the prior. I've demonstrated this in https://github.com/TobyBoyne/bofire/commit/f9df6b096867294ca297f7035dda4f68b58efcac , just copy this change over to your code and you will be able to pickle the kernel :)

jduerholt:
Hey @jpfolch,

I was curious to know what the botorch guys think about changing the order of operations regarding the input transforms in the `posterior` call so I created this issue: https://github.com/pytorch/botorch/issues/2232

Let's see ;)

Best,

Johannes

jpfolch:
Latest commit added the missing validations for the model, and added testing. Overall the multi-task model seems to work now. Only problem I am running into is that the test `test_MultiTaskGPModel` will fail randomly (this is independent of the test parameters). In particular, the problem happens when calling `fit_gpytoch_mll` and the error thrown is `RunTimeError: Must provide inverse transform to be able to sample from prior`, I will try to investigate further in the next few days.

jduerholt:
> Only problem I am running into is that the test `test_MultiTaskGPModel` will fail randomly (this is independent of the test parameters). In particular, the problem happens when calling `fit_gpytoch_mll` and the error thrown is `RunTimeError: Must provide inverse transform to be able to sample from prior`, I will try to investigate further in the next few days.

I have not yet looked at your additions, but do you think this is a bofire or a botorch problem?

jduerholt:
Hi @jpfolch, 

this looks good overall. I will do a more thorough review as soon as you request it and you know more about the issue with the fitting. To test serialization and desirilization for the added datamodels (the new gp and the new prior), please add example configs in `tests/bofire/data_models/specs/surrogates.py` and `tests/bofire/data_models/specs/priors.py`. Everything what is registered there will be automatically tested for serialization. With `add_invalid`, you can add invalid specs to test your custom validators.

Best,

Johannes

jpfolch:
Sorry, I haven't been able to work on this due to other commitments, but I should have more time now. Regarding the `fit_gpytroch_mll` problem. I am still unsure what is causing it, but I think it could be a `gpytorch` problem based on the following two issues: https://github.com/pytorch/botorch/issues/1323 and https://github.com/pytorch/botorch/issues/1860.

The best solution I have come up with, is to drop the LKJ prior all together until the issue is fixed, and use MultiTaskGPs without a prior on the task covariances.

jpfolch:
Latest commit adds the example configs to `tests/bofire/data_models/specs/surrogates.py` and `tests/bofire/data_models/specs/priors.py`. I have left the LKJ prior in the code, but currently it cannot be used: when used it defaults to `None` and throws a warning, let me know if this functionality is okay or should be changed.

bofire/benchmarks/single.py
```diff
@@ -381,6 +382,84 @@ def get_optima(self) -> pd.DataFrame:
         )
 
 
+class MultiTaskHimmelblau(Benchmark):
```
jduerholt, line 385: Can you provide tests for this class?

bofire/data_models/features/tasks.py
jduerholt, line 1: This whole file can be removed, or? Since the actual used `TaskInput` is already implemented in `task.py`. So this is outdated, or?

bofire/priors/mapper.py
```diff
@@ -13,9 +13,16 @@ def map_GammaPrior(data_model: data_models.GammaPrior) -> gpytorch.priors.GammaP
     )
 
 
+def map_LKJPrior(data_model: data_models.LKJPrior) -> gpytorch.priors.LKJPrior:
```
jduerholt, line 16: Can you add a test for this function to make sure that the map works correctly?

bofire/data_models/surrogates/multi_task_gp.py
```diff
@@ -0,0 +1,146 @@
...
+        Args:
+            my_type: continuous or categorical output
+        Returns:
+            bool: True if the output type is valid for the surrogate chosen, False otherwise
+        """
+        return isinstance(my_type, type(ContinuousOutput))
+
+    @field_validator("inputs", mode="before")
+    @classmethod
+    def validate_task_inputs(cls, v, info):
```
jduerholt, line 111:
```suggestion
    def validate_task_inputs(cls, inputs: Inputs):
```

bofire/data_models/surrogates/multi_task_gp.py
```diff
@@ -0,0 +1,146 @@
...
+        if isinstance(v, dict):
```
jduerholt, line 112: You should be able to just do:
```python
if len(inputs.get(TaskInput)) != 1:
    raise ValueError("Exactly one `TaskInput` needed for `MultiTaskGPSurrogate`.")
return inputs
```

tests/bofire/data_models/specs/surrogates.py
```diff
@@ -394,3 +396,36 @@
         "hyperconfig": None,
     },
 )
+
+specs.add_valid(
```
jduerholt, line 400: Can you also add tests via `add_invalid` in which your custom validators for the `MultiTaskGPSurrogate` complain?

bofire/surrogates/multi_task_gp.py
```diff
@@ -0,0 +1,105 @@
...
+        transformed_X = self.inputs.transform(X, self.input_preprocessing_specs)
+
+        tX, tY = torch.from_numpy(transformed_X.values).to(**tkwargs), torch.from_numpy(
+            Y.values
+        ).to(**tkwargs)
+
+        self.model = botorch.models.MultiTaskGP(  # type: ignore
+            train_X=tX,
+            train_Y=tY,
+            task_feature=X.columns.get_loc(
```
jduerholt, line 62:
```suggestion
            task_feature=transformed_X.columns.get_loc(
```

I think we have to take it from here in case of possible transformations with respect to onehots for categorical ones etc.

tests/bofire/surrogates/test_gps.py
```diff
@@ -321,6 +361,45 @@ def test_MixedSingleTaskGPHyperconfig():
         )
 
 
+def test_MultiTask_input_preprocessing():
+    # test that if wrong encoding is used, there is an error
```
jduerholt, line 365: you can move this tests for the errors into the specs firectoy via `add_invalid`

tests/bofire/surrogates/test_gps.py
```diff
@@ -321,6 +361,45 @@ def test_MixedSingleTaskGPHyperconfig():
...
+    )
+    outputs = Outputs(features=[ContinuousOutput(key="y")])
+    with pytest.raises(ValueError):
+        data_model = MultiTaskGPSurrogate(
+            inputs=inputs,
+            outputs=outputs,
+            input_preprocessing_specs={"task_id": CategoricalEncodingEnum.ONE_HOT},
+        )
+
+    # test that if there is no task input, there is an error
```
jduerholt, line 378: this can also go the add_invalid

jduerholt (changes requested):
Hi,

thank you very much, looks overall really good. Only some minor things due to tests and validators. If you have problems with the validators, just tell me, then I try to resolve them. I know they can be tricky ...

As soon as this is landed, we have to implement it to the optimizer ;)

Best,

Johannes

bofire/data_models/features/tasks.py
jpfolch: Yes, that's correct. I'll delete in the next push.

bofire/data_models/features/tasks.py
```diff
@@ -0,0 +1,54 @@
...
+
+    @model_validator(mode="before")
+    def validate_values(cls, values):
+        if "n_tasks" in values:
+            values["values"] = list(range(values["n_tasks"]))
+        return values
+
+
+class TaskInput(CategoricalInput):
+    type: Literal["TaskInputCategorical"] = "TaskInput"
```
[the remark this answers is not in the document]
jpfolch: This is in the outdated file, so it will be deleted

bofire/data_models/surrogates/multi_task_gp.py
```diff
@@ -0,0 +1,131 @@
...
+    def is_output_implemented(cls, my_type: Type[AnyOutput]) -> bool:
+        """Abstract method to check output type for surrogate models
+        Args:
+            my_type: continuous or categorical output
+        Returns:
+            bool: True if the output type is valid for the surrogate chosen, False otherwise
+        """
+        return isinstance(my_type, type(ContinuousOutput))
+
+    @field_validator("inputs", mode="before")
```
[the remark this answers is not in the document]
jpfolch: Yeah, just deleted it and everything seemed to run okay. Can't remember why I put mode `before` in the first place.

bofire/priors/mapper.py
jpfolch: Done.

jpfolch:
I've made all the requested changes. From my part everything is running okay, as far as I can tell.

Some tests are failing locally, at total of fifteen: 4/15 are kernel mapping errors, 8/15 are kernel serialization errors, 2/15 are surrogate specs (specs 4 and 5), and 1/15 is prior serialization (spec2, it says related to `union[GammaPrior, NormalPrior].` . It is unclear if these errors are related to my additions.

jpfolch:
All done, conflicts should be resolved now. I tested, and all tests passed except the EntingStrategy spec, but I don't have entmoot installed which explains it.

jduerholt:
Tests are running through, can you fix the linting errors to?

jpfolch:
I think the linting errors should be fixed now

jduerholt (approved):
Hi @jpfolch,

thank you very much! And sorry for the long process, as I am currently in parental leave, I am not looking every day into github ;)

Next step would be the multifidelity one, or the incorporation into the strategy?

Best,

Johannes

jpfolch:
Hi @jduerholt,

Thank you very much! I've also been slow as I have been away and juggling other things meanwhile. Happy to see it has been merged.

The MultiTaskGP can be used directly to model the fidelities, so I think the next step would be incorporation into the strategy? This is, assuming the strategies are the part of the code where the BO chooses which experiment to do next?

Best,
Jose

jpfolch:
Hi @jduerholt,

Sorry for the delay in reply, I have been busy finishing up work my PhD. I am currently writing up, but I should have some time to work on BoFire. So based on my [previous work](https://www.sciencedirect.com/science/article/pii/S0098135423000637) the way I recommend going about optimizing over `TaskInput` is:

1. We first choose the next point of interest by returning the candidate with the highest acqf value at the _target fidelity_.

2. We then choose which fidelity to evaluate based on one of two criteria:
(a) If the fidelity criterion is strictly ordered, we can simply choose the lowest fidelity until we reduce the uncertainty enough. We would then choose the second lowest fidelity until we reduce uncertainty, and so on... the idea being that we want to use lowest possible fidelity which we believe might still be informative of our target. This criterion is cheap to compute, generally stable, and I found it to work well. However, it does not take cost into account.
(b) A second criterion, which is more general and does take cost into account, is to consider the information gain of each fidelity _on the target fidelity_ per unit cost. This is a more general criteria, however, it can be expensive to compute accurately and numerically unstable. There is a BoTorch implementation of the method though, which is probably well optimized (the single-fidelity method is described in this [tutorial](https://botorch.org/tutorials/max_value_entropy) but there is a multi-fidelity version in the [source code](https://botorch.org/api/_modules/botorch/acquisition/max_value_entropy_search.html)). 

Approach (b) would be equivalent to first optimizing any acquisition function in the target fidelity, and then optimizing the task-input only with the MF-MES acquisition function. The benefit of this is that it gives you more flexibility in the choice acquisition functions and it is computationally cheaper than optimizing MF-MES directly (which is another possible route). 

Let me know which of the two options you would prefer, and I can get started on it.

Best,
Jose

jpfolch:
@bertiqwerty

jduerholt:
Hi @jpfolch,

totally forgot this. I will have a detailed look tomorrow. Sorry. Too much to do :(

Best,

Johannes

jduerholt:
Hi @jpfolch,

soory for the delay, the last weeks were crazy. I am not so much into multifidelity optimization, so I have a few questions:

- I also found these two botorch tutorials: https://github.com/pytorch/botorch/blob/main/tutorials/Multi_objective_multi_fidelity_BO.ipynb and this https://github.com/pytorch/botorch/blob/main/tutorials/multi_fidelity_bo.ipynb. How do them fit in your proposed options?
- Concerning option B: What do you mean with "reducing the uncertainty enough". Which uncertainty? And what is enough?

In general, I am always for starting with the easiest option ;)

Best,

Johannes

PR #355, opened 2024-02-22
Transforms

bertiqwerty:
Implement Transforms

For instance, they they can be applied to steps in the step-wise strategy. This way you can for example change data or candidates the strategies for certain steps or restrict the domain.


bofire/strategies/stepwise/transforms.py
```diff
@@ -0,0 +1,44 @@
+from typing import Optional
+
+import pandas as pd
+
+import bofire.data_models.strategies.stepwise.transforms as data_models
+from bofire.data_models.domain.api import Domain
+
+
+class Transform:
```
jduerholt, line 9: I think we do not need here a distinction between data model and functional model, as the transform does not need any non standard dependecies, or? So one can just implement it for the datamodel. The same could be done also for the conditions, or?

We should also have a loo here: https://github.com/facebook/Ax/tree/main/ax/modelbridge/transforms and think of making the transforms more prominent, as transforms of the domain, the candidates and the experiments could be also useful in other scenarios. Have a look for this to the PR of @Osburg, where he transform the complete domain between bound of -1 and 1. Having this kinds of transform, that could also be chained behind it each other could make sense, they would always take in the domain, experiments and candidates and return the transformed versions, what do you think?

bertiqwerty: Moved to data models as discussed.

bertiqwerty: @jduerholt can you have a look at the invalid/valid spec-stuff? I did something wrong and don't really get it.

R-M-Lee:
I took a quick look at this. I think that the name RemoveTransform is a little bit confusing because I would expect it to remove the transform rather than be a transform that removes something. What do you think of DropDataTransform or ExcludeTransform?

bertiqwerty:
> I took a quick look at this. I think that the name RemoveTransform is a little bit confusing because I would expect it to remove the transform rather than be a transform that removes something. What do you think of DropDataTransform or ExcludeTransform?

Yes! As so often a great naming suggestion from you!

bofire/data_models/strategies/stepwise/conditions.py
```diff
@@ -1,9 +1,18 @@
-from typing import List, Literal, Union
+from abc import abstractmethod
+from typing import List, Literal, Optional, Union
 
+import pandas as pd
 from pydantic import Field, field_validator
 from typing_extensions import Annotated
 
 from bofire.data_models.base import BaseModel
+from bofire.data_models.domain.api import Domain
```
jduerholt, line 9: Why do you have this extra class and not just put the abstract method in `Condition?`

bofire/strategies/stepwise/stepwise.py
```diff
@@ -41,40 +43,67 @@
...
+) -> Optional[T]:
+    if data is not None and transform is not None:
+        return getattr(transform, f"transform_{tf}")(data)
+
+
 class StepwiseStrategy(Strategy):
     def __init__(self, data_model: data_model, **kwargs):
         super().__init__(data_model, **kwargs)
-        self.steps = data_model.steps
+        self.stratgies = [_map(s.strategy_data) for s in data_model.steps]
```
jduerholt, line 69:
```suggestion
        self.strategies = [_map(s.strategy_data) for s in data_model.steps]
```

bofire/strategies/stepwise/stepwise.py
```diff
@@ -41,40 +43,67 @@
...
+        self.conditions = [s.condition for s in data_model.steps]
+        self.transforms = [
+            s.transform and transforms.map(s.transform) for s in data_model.steps
```
jduerholt, line 72: What is the `and` doing? I do not get it :D

bofire/strategies/stepwise/stepwise.py
```diff
@@ -41,40 +43,67 @@
...
+        ]
 
     def has_sufficient_experiments(self) -> bool:
         return True
 
-    def _get_step(self) -> Tuple[int, Step]:  # type: ignore
-        for i, step in enumerate(self.steps):
-            condition = conditions.map(step.condition)
+    def _get_step(self) -> Tuple[Strategy, Optional[Transform]]:
+        """Returns index of the current step, the step itself"""
```
jduerholt, line 79: this docstring does not matches the signature of the method

bofire/strategies/stepwise/stepwise.py
```diff
@@ -41,40 +43,67 @@
...
+        for i, condition in enumerate(self.conditions):
             if condition.evaluate(self.domain, experiments=self.experiments):
-                return i, step
+                return self.stratgies[i], self.transforms[i]
```
jduerholt, line 82:
```suggestion
                return self.strategies[i], self.transforms[i]
```

bofire/strategies/stepwise/stepwise.py
```diff
@@ -41,40 +43,67 @@
...
+        transformed_domain = tf_domain or self.domain
+        strategy.domain = transformed_domain
+        tf_exp = _apply_tf(self.experiments, transform, "experiments")
+        transformed_experiments = self.experiments if tf_exp is None else tf_exp
+        tf_cand = _apply_tf(self.candidates, transform, "candidates")
+        transformed_candidates = self.candidates if tf_cand is None else tf_cand
         # tell the experiments
-        if self.num_experiments > 0:
-            strategy.tell(experiments=self.experiments)
+        if transformed_experiments is not None and self.num_experiments > 0:
```
jduerholt, line 99:
```suggestion
        if transformed_experiments is not None and len(transformed_experiments) > 0:
```

bofire/strategies/stepwise/stepwise.py
```diff
@@ -41,40 +43,67 @@
...
         # tell pending
-        if self.num_candidates > 0:
-            strategy.set_candidates(self.candidates)
+        if transformed_candidates is not None and len(transformed_candidates) > 0:
+            strategy.set_candidates(transformed_candidates)
         # ask and return
-        return strategy.ask(candidate_count=candidate_count)
+        data_asked_for = strategy.ask(candidate_count=candidate_count)
+        if transform is not None:
+            return transform.untransform_experiments(data_asked_for)
```
jduerholt, line 107:
```suggestion
            return transform.untransform_candidates(data_asked_for)
```

why not call `data_asked_for` `candidates`?

bofire/strategies/stepwise/stepwise.py
```diff
@@ -41,40 +43,67 @@
...
+        else:
+            return data_asked_for
```
jduerholt, line 109:
```suggestion
            return candidates
```

bofire/transforms/transform.py
```diff
@@ -0,0 +1,20 @@
...
+        return experiments
+
+    def transform_candidates(self, candidates: pd.DataFrame) -> pd.DataFrame:
+        return candidates
+
+    def transform_domain(self, domain: Domain) -> Domain:
+        return domain
+
+    def untransform_experiments(self, experiments: pd.DataFrame) -> pd.DataFrame:
+        return experiments
```
jduerholt, line 20: I would also opt for an `untransform_candidates`

tests/bofire/conftest.py
```diff
@@ -74,6 +74,11 @@ def invalid_condition_spec(request) -> specs.InvalidSpec:
     return request.param
 
 
+@fixture(params=specs.transforms.invalids)
```
jduerholt, line 77: You need also a fixture for the valid ones, see below in the file.

jduerholt (changes requested):
Hi Behrang,

only a few minor comments and one test is missing, namely where it is checked that the acutal drop data transfrom works and drops the correct data, so to say, testing of its functionality.

Best,

Johannes

bofire/strategies/stepwise/stepwise.py
bertiqwerty: > What is the `and` doing? I do not get it :D

If `s.transform` is `None` we keep `None`. If it is not `None` the last value is evaluated and returned.
```
assert (None and 1) is None
assert (None or 1) == 1
```
https://en.wikipedia.org/wiki/Short-circuit_evaluation

bofire/data_models/strategies/stepwise/conditions.py
bertiqwerty: I prefer abstract classes to be really abstract and to be mixed with data members or implementations.

bofire/transforms/transform.py
bertiqwerty: We only untransform candidates. And only the new candidates that have been proposed. For the other things, the inputs are the untransformed versions, no?

bertiqwerty:
> namely where it is checked that the acutal drop data transfrom works and drops the correct data, so to say, testing of its functionality.

There was a test in the stepwise-file. I moved it to a separate file and extended it.

jduerholt (approved):
Looks good to me, last minor thing would be to not have test_transforms.py under tests/bofire/strategegies but in a new folder tests/bofire/stransforms.py so that we reflect the correct folder structure. If one is very strict we would also need the test_conditions in the data_models part of the test suite.

bertiqwerty:
> Looks good to me, last minor thing would be to not have test_transforms.py under tests/bofire/strategegies but in a new folder tests/bofire/stransforms.py so that we reflect the correct folder structure. If one is very strict we would also need the test_conditions in the data_models part of the test suite.

Thanks for the review.

Since transforms are tight to the stepwise strategy I will leave the tests there. We can move them later.

PR #363, opened 2024-03-08
separate into actual and meta strategy

bertiqwerty:


jduerholt (approved):
Thank you very much, it was originally duplicated to avoid circular imports. But your implementation solves it!

Perhaps, we should rename the files to `actual_strategy_types.py` and `meta_strategy_types.py` since we also have `types.py`

PR #373, opened 2024-03-13
Hammond -> Hamming and change map_Hamming

R-M-Lee:
I just renamed Hammond to Hamming. But in [bofire/kernels/mapper.py](https://github.com/experimental-design/bofire/compare/372-hamming-distance-instead-of-hammond-distance?expand=1#diff-b1fdeaa29fa739932b6dec7adcd767cee636333674ecc0bfc4ac5a8d7aa47fa1) there is a real change that I wanted to check with you


jduerholt (approved):
Good catch. Thank you!

PR #376, opened 2024-03-18
Qparege with constraints and batch proposals crashes

bertiqwerty:
The error was not detected because earlier
- ~~The previous constraint-test of Qparego didn't seem to have constraints.~~ _There are output constraints but no input constraints in the test._
- The previous constraint-test didn't test batch proposals even though it looked like it did at first glance.

This PR also adds the detergent benchmark problem.

### DIFF
--- a/tests/bofire/benchmarks/test_detergent.py
+++ b/tests/bofire/benchmarks/test_detergent.py
@@ -1,0 +1,18 @@
+from bofire.benchmarks.detergent import Detergent
+from bofire.data_models.strategies.random import (
+    RandomStrategy as RandomStrategyDataModel,
+)
+from bofire.strategies.api import RandomStrategy
+
+
+def test_detergent():
+    d = Detergent()
+    assert len(d.domain.inputs) == 5
+    assert len(d.domain.outputs) == 5
+    assert len(d.domain.constraints) == 2
+
+    # check that the domain has feasible points
+    random_search_dm = RandomStrategyDataModel(domain=d.domain)
+    random_search = RandomStrategy(random_search_dm)
+    candidates = random_search.ask(2)
+    assert d.domain.constraints.is_fulfilled(candidates).all()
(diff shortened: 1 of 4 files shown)


tests/bofire/strategies/test_qparego.py
```diff
@@ -142,24 +143,27 @@ def test_qparego(num_test_candidates):
 
 @pytest.mark.parametrize(
     "num_test_candidates",
-    list(range(1, 2)),
+    [1, 2],
 )
 def test_qparego_constraints(num_test_candidates):
     # generate data
-    benchmark = C2DTLZ2(dim=4)
+    benchmark = Detergent()
```
jduerholt, line 150: Can you keep the test with C2DTLZ2 here? Because it is used to test if qparego also work with output constraints which are present in C2DTLZ2. You can copy the test for the input constraints and then use the `Detergent` benchmark.

jduerholt (changes requested):
Hi @bertiqwerty,

very good catch. I totally overlooked it. I am also currently working on a PR refactoring the `batch_limit` and maxiter stuff to make it more transparent. I hope to get it landed this week.

One thing out of curiosity: are you also trying `MOBO` with `qlognehvi` on these problems? In our experience, it performs really well, but 5 outputs could be also too tough from a numeric perspective for hypervolume based methods ...

Best,

Johannes

tests/bofire/benchmarks/test_detergent.py
```diff
@@ -0,0 +1,18 @@
...
+def test_detergent():
+    d = Detergent()
+    assert len(d.domain.inputs) == 5
+    assert len(d.domain.outputs) == 5
+    assert len(d.domain.constraints) == 2
+
+    # check that the domain has feasible points
+    random_search_dm = RandomStrategyDataModel(domain=d.domain)
+    random_search = RandomStrategy(random_search_dm)
+    candidates = random_search.ask(2)
```
jduerholt, line 17: Can you also add a call to f?

```
d.f(candidates)
```

So that it is checked that it is functional?

jduerholt (approved):
Hi Behrang,

looks good for me. Only one minor thing, but just go forward with it ;)

Best,

Johannes

PR #379, opened 2024-03-19
Fix bug in `aggregate_by_duplicates`

jduerholt:
`aggregate_by_duplicates` was throwing an error when used for domains with no continuous features. This PR fixes it.


bertiqwerty (approved):
Thanks Johannes.

PR #382, opened 2024-03-20
Generic typing for Features, Inputs, and Outputs

TobyBoyne:
Implements generic typing for Features.

Currently, the dynamic typing for `Features`, `Inputs`, and `Outputs` doesn't work well with pyright. There are many `# type: ignore`s in the codebase, such as shown below, to address this. However, a solution that works with pyright is generic typing. With this PR, you can remove these comments and the type checker is still happy. The API for interacting with `Features`, `Inputs`, and `Outputs` remains unchanged,

https://github.com/experimental-design/bofire/blob/0efb4685ce72c163a44e0f1fdc325ce191dda31d/bofire/data_models/strategies/predictives/botorch.py#L198


TobyBoyne:
Ah, this didn't go well. Will fix this!

PR #385, opened 2024-03-21
Feature/linear deterministic

jduerholt:
This PR adds a new surrogate namely the `LinearDeterministicSurrogate` which is just a linear model of type `f(x) = mx +b`, where are `m` are the coefficients and `b` is the intercept. Tis is useful for multi-objective optimization with partially known objective functions (for example the number of parameters of a Neural Network in the context of hyperparameter optimization).


bertiqwerty (approved):
Looks good. Thanks!

PR #386, opened 2024-03-22
Add generic typing to Constraints

TobyBoyne:
Following on from #383, here's generic typing in Constraints!

Pyright (and whatever type checker your IDE uses) will now infer the type of constraints contained in a `Constraints` object.

It isn't perfect - you can't type hint set subtraction [1], so the `excludes` keyword isn't typed properly. But it clears up more `# type: ignore`s than before!

```python
constraints = Constraints(constraints=[LinearEqualityConstraint(...), NChooseKConstraint(...)]))
# type of constraints.features is inferred as Sequence[LinearEqualityConstraint | NChooseKConstraint]
sub_constraints = constraints.get(includes=[NChooseKConstraint])
# type of sub_constraints.features is inferred as Sequence[NChooseKConstraint]
```

[1] https://discuss.python.org/t/type-intersection-and-negation-in-type-annotations/23879

### DIFF
--- a/bofire/data_models/domain/constraints.py
+++ b/bofire/data_models/domain/constraints.py
@@ -1,31 +1,49 @@
 import collections.abc
 from itertools import chain
-from typing import List, Literal, Optional, Sequence, Type, Union
+from typing import (
+    Generic,
+    Iterator,
+    List,
+    Literal,
+    Optional,
+    Sequence,
+    Type,
+    TypeVar,
+    Union,
+    get_args,
+)
 
 import pandas as pd
 from pydantic import Field
 
 from bofire.data_models.base import BaseModel
-from bofire.data_models.constraints.api import AnyConstraint, Constraint
+from bofire.data_models.constraints.api import (
+    AnyConstraint,
+)
 from bofire.data_models.filters import filter_by_class
 
+C = TypeVar("C", bound=AnyConstraint)
+C_ = TypeVar("C_", bound=AnyConstraint)
+C__ = TypeVar("C__", bound=AnyConstraint)
+AnyConstraintTuple = get_args(AnyConstraint)
 
-class Constraints(BaseModel):
+
+class Constraints(BaseModel, Generic[C]):
     type: Literal["Constraints"] = "Constraints"
-    constraints: Sequence[AnyConstraint] = Field(default_factory=lambda: [])
+    constraints: Sequence[C] = Field(default_factory=lambda: [])
 
-    def __iter__(self):
+    def __iter__(self) -> Iterator[C]:
         return iter(self.constraints)
 
     def __len__(self):
         return len(self.constraints)
 
-    def __getitem__(self, i):
+    def __getitem__(self, i) -> C:
         return self.constraints[i]
 
     def __add__(
-        self, other: Union[Sequence[AnyConstraint], "Constraints"]
-    ) -> "Constraints":
+        self, other: Union[Sequence[C_], "Constraints[C_]"]
+    ) -> "Constraints[Union[C, C_]]":
         if isinstance(other, collections.abc.Sequence):
             other_constraints = other
         else:
@@ -78,10 +96,10 @@ class Constraints(BaseModel):
 
     def get(
         self,
-        includes: Union[Type, List[Type]] = Constraint,
-        excludes: Optional[Union[Type, List[Type]]] = None,
+        includes: Union[Type[C_], Sequence[Type[C_]]] = AnyConstraintTuple,
+        excludes: Optional[Union[Type[C__], List[Type[C__]]]] = None,
         exact: bool = False,
-    ) -> "Constraints":
+    ) -> "Constraints[C_]":
         """get constraints of the domain
 
         Args:
--- a/bofire/data_models/domain/constraints.py
+++ b/bofire/data_models/domain/constraints.py
@@ -10,22 +10,18 @@ from typing import (
     Type,
     TypeVar,
     Union,
-    get_args,
 )
 
 import pandas as pd
 from pydantic import Field
 
 from bofire.data_models.base import BaseModel
-from bofire.data_models.constraints.api import (
-    AnyConstraint,
-)
+from bofire.data_models.constraints.api import AnyConstraint, Constraint
 from bofire.data_models.filters import filter_by_class
 
-C = TypeVar("C", bound=AnyConstraint)
-C_ = TypeVar("C_", bound=AnyConstraint)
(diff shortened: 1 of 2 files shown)


jduerholt:
Nice, thanks a lot @TobyBoyne! Can you perhaps elaborate a bit on what you actually did and how it works, as I do not understand why it works ;) So, I think it is better when @bertiqwerty reviews it. If it is fine for him, can you then also file a similar PR for `Features`, `Inputs`, and `Outputs`. 

We should then also remove the following functions from domain as they are obsolete and the corresponding versions from `domain.inputs`, `domain.outputs` and `domain.constraints` should be used:
- `domain.get_features`
- `domain.get_feature_keys`
- `get_feature`

Or what do you think @bertiqwerty ?

TobyBoyne:
Of course! Just to note, this **does not change how users interface with bofire**. The only way that users will interact with this is by seeing that their type hints have automatically become more informative! Also, this has already been implemented for Features, Inputs, and Outputs in #383.

So this PR uses [Generic typing](https://docs.python.org/3/library/typing.html#generics). You will have used generic type hints whenever you do something like
```python
my_list: List[int]
x = my_list[0]
# the type of x is automatically inferred to be `int`
```
In this example, `List` accepts `int` as a `TypeVar`, which determines the type of the elements in the list.

I've done something similar here. You can now create a `Constraints` object that contains only one type of constraint, like `Constraint[NChooseKConstraint]`. 

However, you **don't want to manually type hint** the type of constraints each time. Going back to the previous example, we want something more like
```python
my_list = [2, 4, 8]
x = my_list[0]
# the type of x is automatically inferred to be `int`, no explicit type hint required!
```
This is where [TypeVar](https://docs.python.org/3/library/typing.html#typing.TypeVar) is useful. This allows is to infer the type of the elements in `Constraint.constraints` using generics. Whenever you make a new instance of a `Constraint` object, the `TypeVar` `C` stores the type of constraint passed to the constructor. This means that when you iterate through the constraints, your typechecker knows that the items you iterate over will be limited to the type of constraints that were passed to the input.

We can do even better. We use a second TypeVar, `C_`, for use in the `get` method. This allows us to keep track of the types of constraint passed to the `includes` argument, so that we can further restrict the type of constraint in the returned object. So now, when we have code like below:

```python
constraints = Constraints(constraints=[...])
for constraint in constraints.get(includes=[NChooseKConstraint]):
    print(constraint.max_count)
   # Normally, pyright would raise an error here, since AnyConstraint.max_count isn't defined.
   # But with generic types, pyright can infer that the returned constraints from the `get` call
   #   must be of type NChooseKConstraint, so this attribute is always defined!
```

I hope this clears things up :) Let me know if you have any other questions!

bertiqwerty:
> We should then also remove the following functions from domain as they are obsolete and the corresponding versions from `domain.inputs`, `domain.outputs` and `domain.constraints` should be used:
> 
>     * `domain.get_features`
> 
>     * `domain.get_feature_keys`
> 
>     * `get_feature`
> 
> 
> Or what do you think @bertiqwerty ?

Agreed. In a separate PR.

bofire/data_models/domain/constraints.py
```diff
@@ -9,23 +19,27 @@
 from bofire.data_models.constraints.api import AnyConstraint, Constraint
 from bofire.data_models.filters import filter_by_class
 
+C = TypeVar("C", bound=Union[AnyConstraint, Constraint])
```
bertiqwerty, line 22: Trailing underscores are recommended to be used to avoid conflicts Python builtins, see https://peps.python.org/pep-0008/#descriptive-naming-styles. I would use something else, e.g., numbers to to have multiple constraint types.

bofire/data_models/domain/constraints.py
```diff
@@ -78,10 +92,10 @@ def is_fulfilled(self, experiments: pd.DataFrame, tol: float = 1e-6) -> pd.Serie
...
-        includes: Union[Type, List[Type]] = Constraint,
-        excludes: Optional[Union[Type, List[Type]]] = None,
+        includes: Union[Type[C_], Sequence[Type[C_]]] = Constraint,
+        excludes: Optional[Union[Type[C__], List[Type[C__]]]] = None,
         exact: bool = False,
-    ) -> "Constraints":
+    ) -> "Constraints[C_]":
         """get constraints of the domain
 
         Args:
```
bertiqwerty, line 101: Can you remove the types from the comments? They are wrong now and we don't need them anyway.

bofire/utils/torch_tools.py
```diff
@@ -34,7 +34,7 @@
 
 def get_linear_constraints(
     domain: Domain,
-    constraint: Union[LinearEqualityConstraint, LinearInequalityConstraint],
+    constraint: Union[Type[LinearEqualityConstraint], Type[LinearInequalityConstraint]],
     unit_scaled: bool = False,
 ) -> List[Tuple[Tensor, Tensor, float]]:
     """Converts linear constraints to the form required by BoTorch.
```
bertiqwerty, line 40: Again, please remove out-dated types from comments.

bertiqwerty (approved):
Thanks, Toby. Please check my review comments and merge afterwards.

bofire/data_models/domain/constraints.py
TobyBoyne: Yes that's fair! Although just to note, TypeVar has some [additional naming rules](https://stackoverflow.com/q/74589610), which I don't think allow numbers.

bofire/data_models/domain/constraints.py
TobyBoyne: When you say we don't need types in the docstrings, are they not used to generate the docs? For example [here](https://experimental-design.github.io/bofire/ref-constraints/#bofire.data_models.domain.constraints), the docs list the return type of the `get` method according to the docstring, even though this type differs from the return type in the function signature.

bofire/utils/torch_tools.py
TobyBoyne: I've updated the signature, but please see other comment about keeping types in docstrings.

bofire/data_models/domain/constraints.py
bertiqwerty: If the doc-extractor is using the types in the docstrings we should fix the doc-extractor.

TobyBoyne: Good point. Shall I leave the comments in for now, and then leave that to another PR? Or still take out the comments?

bertiqwerty: Remove them. :)

PR #387, opened 2024-03-26
Remove `math.inf` from ContinuousInput data model

jduerholt:
This PR removes the default `(math.inf, math.inf)` from the `ContiuousInput`, as this creates with `FastAPI` as @jkeupp find out in his PR on the data models from the hackathon.


bertiqwerty (approved):
Thanks, Johannes.

PR #392, opened 2024-05-01
add to_candidates for stepwise strategy 

jkleinekorte:
Currently, stepwise strategy only returns input values for each candidate when calling to_candidate. This PR replaces the to_candidate method for stepwise strategy such that the to_candidate method of the currently used strategy is used instead. As a result, also predicted output values are returned in case a predictive strategy is used.


jduerholt (approved):
Looks good to me.

PR #395, opened 2024-05-16
name construction in a method

bertiqwerty:
Make column names of categorical features and descriptors after encoding to, e.g., one-hot accessible from the outside and re-use the same code at all places inside BoFire.


dlinzner:
I like it. approve.

dlinzner (approved):

dlinzner-bcs (approved):
the wrong dominik linzner approved...

PR #396, opened 2024-05-17
Feature type hints

bertiqwerty:
I added some `Optional`s that where missing. I also am confused about includes and excludes and added some tests. 

Now I understand how includes/excludes work but I am still confused why it is an option to pass both, includes and excludes. It complains about overlaps anyways. Further, if I pass just excludes it fails with overlap. I think the default of includes should be `None` and in case both are `None` we should just skip the `filter_by_class` part and return the whole thing. But probably I have missed something.

### DIFF
--- a/tests/bofire/data_models/domain/test_features.py
+++ b/tests/bofire/data_models/domain/test_features.py
@@ -1,9 +1,12 @@
+from collections.abc import Sequence
+
 import pytest
 from pydantic.error_wrappers import ValidationError
 
 import tests.bofire.data_models.specs.api as specs
 from bofire.data_models.domain.api import Features, Inputs, Outputs
 from bofire.data_models.features.api import (
+    AnyFeature,
     CategoricalInput,
     ContinuousInput,
     ContinuousOutput,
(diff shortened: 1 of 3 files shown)


tests/bofire/data_models/domain/test_features.py
```diff
@@ -150,3 +153,24 @@ def test_features_get_by_keys():
...
+        features = Inputs(features=[if1, if2, if3, if4, if5, if7])
+        filtered = features.get(includes=includes, excludes=excludes)
+        assert set(filtered.get_keys()) == {e.key for e in expected}
+
+    test(includes=None, excludes=[ContinuousInput], expected=[if4, if5, if7])
+    test(includes=[ContinuousInput], excludes=None, expected=[if1, if2, if3])
+    test(includes=[CategoricalInput], excludes=None, expected=[if4, if7])
+    test(includes=None, excludes=[CategoricalInput], expected=[if1, if2, if3, if5])
+    test(includes=AnyFeature, excludes=None, expected=[if1, if2, if3, if4, if5, if7])
+    with pytest.raises(ValueError):
```
jduerholt, line 169: Can you also give the error message here via: 

```suggestion
    with pytest.raises(ValueError, match="the error message"):
```

tests/bofire/data_models/domain/test_features.py
```diff
@@ -150,3 +153,24 @@ def test_features_get_by_keys():
...
+        test(includes=None, excludes=None, expected=[if1, if2, if3, if4, if5, if7])
+    with pytest.raises(ValueError):
```
jduerholt, line 171: Same as above.

jduerholt (commented):
Hi Behrang,

I was confused by the same thing this week. What do you think about raising a more informed error message when both  `include` and `exclude` are `!=None` somehow stating that one only is allowed to have one that is not None?

This would have the advantage of not changing any defaults which could cause pain ;)

Best,

Johannes

bertiqwerty:
After a little more investigation I found the reason why you can pass both to be not `None`. Includes and excludes are expanded in case they are unions like `AnyFeature`. In this case it can happen that the same class (not sub-class) is included and excluded and the overlap complain appears.

But it still makes sense to exclude subclasses of inputs. I find it neither simple nor elegant but still rather confusing. Further, I think probably all use cases can represent their filter differently. E.g., if you just want `CategoricalInputs` but not `CategoricalDescriptorInputs` you can use `exact`. If you want `CategoricalInputs` and `CategoricalDescriptorInputs` but not `MolecularDescriptorInput` you can use include and exact again. On the other hand, if you want all `NumericalInput`s and `CategoricalInput`s except `CategoricalDescriptorInputs` it might be tedious with `exact`. But who wants that?

Not sure how to proceed.

1. It would be a breaking API change to allow only one to be `!=None`. Not sure if any BoFire users rely on this. 
2. What is the purpose of the overlap-exception? Maybe it was intended to help find mis-specifications of the user. You could pass `includes=ContinuousInput, excludes=ContinuousInput` which does looks like a user error. Maybe we should check for overlap before we expand the `Union`s? 

What do you think @jduerholt ?

bertiqwerty:
In commit [c5aa866](https://github.com/experimental-design/bofire/pull/396/commits/c5aa8664d137355bb0b6e04389298c3c282cd253) I kept the both-can-be-something behavior but moved the overlap check. Now you can pass `AnyFeature` and still exclude stuff which should not break the API. As mentioned above I would prefer the only-one-can-be-something behavior if this is acceptable.

jduerholt (approved):
Hi @bertiqwerty,

looks good for me. I like that it is now also possible to provide Unions without running into the overlap error. I personally like that it is possible to provide both. 

So, if it is okay for you, I propose to go forward with your latest solution and merge it in.

Best,

Johannes

PR #402, opened 2024-06-06
Implement Priors from "Vanilla Bayesian Optimization Performs Great in High Dimensions"

jduerholt:
This PR implements the dimensionality scaled priors from the paper  "Vanilla Bayesian Optimization Performs Great in High Dimensions" from Hvarfner et al. (https://arxiv.org/pdf/2402.02229) This is also a nice blogpost regarding the method:  https://www.miguelgondu.com/blogposts/2024-03-16/when-does-vanilla-gpr-fail/

Probably, I will set the prior also as new default.


bertiqwerty:
That's a great idea!

jduerholt:
Anything missing fore merging it in? Or do you need a bit more time for review?

bofire/data_models/priors/api.py
```diff
@@ -12,8 +16,14 @@
     GammaPrior,
     NormalPrior,
     LKJPrior,
+    LogNormalPrior,
+    DimensionalityScaledLogNormalPrior,
 ]
 
+# these are priors that are generally applicable
+# and does not depend on problem specific extrea parameters
```
bertiqwerty, line 24: ...and do not depend...

bertiqwerty (approved):
Looks good!

bofire/data_models/priors/api.py
jduerholt: Fixed it. Thanks!

PR #404, opened 2024-06-13
duplicates plot

jduerholt:
This PR adds a plot that can be used to show duplicates in the data as below:

image

### DIFF
--- a/bofire/plot/duplicates.py
+++ b/bofire/plot/duplicates.py
@@ -1,0 +1,70 @@
+from typing import List
+
+import numpy as np
+import pandas as pd
+import plotly.graph_objects as go
+
+
+def plot_duplicates_plotly(
+    experiments: pd.DataFrame, duplicates: List[str], key: str, scale: float = 0.1
+) -> go.Figure:
+    """
+    Plots duplicates using Plotly.
+
+    Args:
+        experiments: The dataframe containing the experiments data.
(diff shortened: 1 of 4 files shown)


bertiqwerty:
Thanks, Johannes. Can you explain the use case?

jduerholt:
It is one of my typical methods that I use when working with real world datasets and preparing them for BO, I search for the duplicates in the data and plot them like this to see if there are obvious outliers in the duplicates ;)

bofire/plot/duplicates.py
```diff
@@ -0,0 +1,70 @@
...
+
+def plot_duplicates_plotly(
+    experiments: pd.DataFrame, duplicates: List[str], key: str, scale: float = 0.1
+) -> go.Figure:
+    """
+    Plots duplicates using Plotly.
+
+    Args:
+        experiments: The dataframe containing the experiments data.
+        duplicates: A list of strings representing the labcodes of the duplicates.
```
bertiqwerty, line 16: What are labcodes?

jduerholt: Hi @bertiqwerty ,

these is an optional column in the `experiments` data frame which can be a string based unique identifier of one experiment.

Best,

Johannes

bertiqwerty (approved):

PR #412, opened 2024-07-04
Finally a Logo

jduerholt:
This PR adds BoFire Logos to `graphics/logos` and adds them also to `README.md`. At some point we have to also include it in the docs. And we need merch, especially stickers and T-Shirts ;)


bertiqwerty (approved):
Jippi! Nice logo!

R-M-Lee:
nice work

PR #419, opened 2024-07-15
Readne example

bertiqwerty:
The example in the readme use `Inputs.sample` which is only ok as long as you don't have constraints. Hence I replaced it with `RandomSearch`. Further, I added a BO-loop, not just one ask-and-tell call.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,25 +75,33 @@ domain = Domain(
 ```
 
 You can also use one of the many benchmarks available in BoFire.
-Here, we use the Himmelblau benchmark to demonstrate the ask/tell interface for
-proposing new experiments.
+Here, we use the Detergent benchmark to demonstrate the ask/tell interface for
+proposing new experiments with multi-objective Bayesian optimization.
 
 ```python
-from bofire.benchmarks.single import Himmelblau
-
-benchmark = Himmelblau()
-samples = benchmark.domain.inputs.sample(10)
-experiments = benchmark.f(samples, return_complete=True)
-
-from bofire.data_models.strategies.api import SoboStrategy
-from bofire.data_models.acquisition_functions.api import qNEI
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -75,25 +75,33 @@ domain = Domain(
...
+for _ in range(n_experiments):
+    mobo_strategy.tell(experiments=experiments)
+    candidates = mobo_strategy.ask(candidate_count=1)
+    experiments = detergent.f(candidates, return_complete=True)
+
+# print all told experiments
+print(mobo_strategy.experiments)
 ```
 
 This gives one step in the optimization loop. We can repeat this many times to
```
TobyBoyne, line 107: This line is no longer true - remove?

README.md
```diff
@@ -75,25 +75,33 @@ domain = Domain(
...
-sobo_strategy.ask(candidate_count=1)
+from bofire.benchmarks.detergent import Detergent
+from bofire.data_models.strategies.api import QnehviStrategy, RandomStrategy
+
+# create benchmark
+detergent = Detergent()
+domain = detergent.domain
+
+# create initial data
+sampler = strategies.map(RandomStrategy(domain=domain))
```
TobyBoyne, line 91: Add a comment here, something like:
`# use a random strategy to sample while satisfying constraints`

TobyBoyne (changes requested):
Looks good! And thanks for noting the change about RandomStrategy. I've left a couple of comments but otherwise this is ready to merge.

README.md
bertiqwerty: done

README.md
bertiqwerty: done

TobyBoyne (approved):

PR #358, opened 2024-02-26
Feature Scaling in DoE

jduerholt:
Hi @Osburg,

this is my interpretation of the scaling as discussed in https://github.com/experimental-design/bofire/pull/354 and https://github.com/experimental-design/bofire/issues/343.

What do you think? No testing etc. yet. 

Feel also free, to proceed just from this PR, of course only if you want ;)

Best,

Johannes

PR #423, opened 2024-07-31
Types for bounds

jduerholt:
This PR adds a new type for Bounds (and respective tests) that can be reused in other scenarios as here for @Osburg in this PR: https://github.com/experimental-design/bofire/pull/358


bertiqwerty (approved):
Thanks, Johannes.

PR #427, opened 2024-08-19
Make feature validation part of the constraints.

jduerholt:
This PR refactors how it is validated that the inputs defined in a domain fit to the constraints added to the domain. This new implementation makes it more object oriented and less convoluted.


bertiqwerty (approved):
Thanks, Johannes.

PR #429, opened 2024-08-19
Use weights_only for load

kit1980:
`torch.load` without `weights_only` parameter is unsafe. Explicitly set `weights_only` to False only if you trust the data you load and full pickle functionality is needed, otherwise set `weights_only=True`.

If `weights_only=True` doesn't work for some cases, then explicit `weights_only=False` should be used.

Found with https://github.com/pytorch-labs/torchfix/

### DIFF
--- a/bofire/surrogates/botorch.py
+++ b/bofire/surrogates/botorch.py
@@ -64,4 +64,4 @@ class BotorchSurrogate(Surrogate):
     def loads(self, data: str):
         """Loads the actual model from a base64 encoded pickle bytes object and writes it to the `model` attribute."""
         buffer = io.BytesIO(base64.b64decode(data.encode()))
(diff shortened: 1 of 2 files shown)


kit1980:
I ended up specifying `weights_only=False` in both places, as otherwise the tests fail.
This is still valuable as in a future PyTorch release `weights_only` will default to `True`, so the tests will break if `weights_only` is not set explicitly.

Ideally investigation is needed if the full pickle functionality is logically required there, and if it's possible to change the code to make `weights_only=True` work.

kit1980:
@jduerholt  Can this be landed?

jduerholt:
Yeah, thank you very much1

jduerholt (approved):

PR #431, opened 2024-08-21
Add default generators

jduerholt:
This PR adds default generators which are used for setting up for fractional factorial designs, which circumvents to compute them over and over agains. Furthermore some of them are better than the one which are provided by the current algorithm.


R-M-Lee (commented):
Very nice! Small comment: I think people unfamiliar with statistical DoE might be confused by our use of the name "default generators". Especially since generators are of course also a python thing with a completely different meaning.

Fractional factorial designs are mentioned in the comment at the top of default_generators.py, but that's all. And it's in utils so the import also doesn't give you a clue about it's meaning. What do you think about slighly more verbose or different naming? Like default_FF_generators, default_fracfac_generators or something like that?

jduerholt:
@R-M-Lee: thanks for the suggestion. I like it and implemented it. Best, Johannes

R-M-Lee (approved):

PR #434, opened 2024-08-26
new benchmark with sum of mvnorm pdfs

R-M-Lee:
I had fun experimenting with this trying to make benchmarks that I thought were similar to real problems.

Perhaps it might prove interesting and useful to others too.

### DIFF
--- a/bofire/benchmarks/single.py
+++ b/bofire/benchmarks/single.py
@@ -522,8 +522,8 @@ class Multinormalpdfs(Benchmark):
         eigscale: float = 0.5,
         opt_on_boundary: bool = False,
         N_unimportant_inputs: int = 2,
-        means: list | None = None,
-        covmats: np.ndarray | None = None,
+        means: Optional[list] = None,
+        covmats: Optional[np.ndarray] = None,
         **kwargs,
     ) -> None:
         """Initializes the class of type Multinormalpdfs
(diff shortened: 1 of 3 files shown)


bofire/benchmarks/single.py
```diff
@@ -494,3 +495,140 @@ def __init__(self, **kwargs):
...
+
+    def _f(self, X: pd.DataFrame) -> pd.DataFrame:
+        return pd.DataFrame(
+            {
+                "y": X.apply(
+                    lambda x: sum([g.pdf(x.to_numpy()) for g in self.gaussians]), axis=1
+                ),
+                "valid_y": np.ones(len(X)),
+            },
+            index=range(len(X)),
```
jduerholt, line 621: This is not necessary as the index is set to the range by default.

bofire/benchmarks/single.py
```diff
@@ -494,3 +495,140 @@ def __init__(self, **kwargs):
...
+    def __init__(
+        self,
+        dim: int = 5,
+        n_gaussians: int = 1,
+        stdev: float = 0.4,
+        eigscale: float = 0.5,
+        opt_on_boundary: bool = False,
+        N_unimportant_inputs: int = 2,
+        means: Optional[list] = None,
+        covmats: Optional[np.ndarray] = None,
```
jduerholt, line 526: It would be nice to also add a `seed` argument of type `seed: Optional[int] = None`, to be able to reproduce randomly generated response surfaces.

jduerholt (approved):
Looks good to me. I only added two minor comments. Just merge it when you are ready. And of course, I am interested in your benchmark results ;)

bofire/benchmarks/single.py
jduerholt: We should then also test in the test that it produces the same numbers when a seed is set.

PR #439, opened 2024-09-10
Tutorial on Transfer Learning Bayesian Optimisation

jpfolch:
Added a tutorial on how to use BoFire for Transfer Learning Bayesian Optimisation. Also made two small modifications for the code to run:

- Inverse transform for ordinal encodings transforms into `int` type

- When predicting within the task using MultiTaskGPs we must add likelihood noise manually since the `observation_noise` flag is not supported by BoTorch for MultiTaskGPs

Next step would be to incorporate a step for task selection for multi-fidelity problems.


jduerholt:
Thanks @jpfolch. I will provide a review tomorrow. Best, Johannes

bofire/strategies/predictives/botorch.py
```diff
@@ -170,13 +170,21 @@ def _predict(self, transformed: pd.DataFrame) -> Tuple[np.ndarray, np.ndarray]:
         # input and further transform it to a torch tensor
         X = torch.from_numpy(transformed.values).to(**tkwargs)
         with torch.no_grad():
-            posterior = self.model.posterior(X=X, observation_noise=True)  # type: ignore
+            # observation noise is not implemented for MultiTaskGPSurrogate, has to be treated differently
+            if self.surrogate_specs.surrogates[0].type == "MultiTaskGPSurrogate":
```
jduerholt, line 174: Can you add a validator to `BotorchSurrogates` that checks that in case one `MultiTaskGP` is present, all surrogates are of type `MultiTaskGP`. Else, we will get trouble with the different required encodings. You have to add it here: https://github.com/experimental-design/bofire/blob/main/bofire/data_models/surrogates/botorch_surrogates.py

jduerholt (changes requested):
Hi @jpfolch,

looks very nice. Thank you!

I let one comment inline, and here are two other ones:

- Can you also add a validator to the botorch data model (https://github.com/experimental-design/bofire/blob/main/bofire/data_models/strategies/predictives/botorch.py) that checks if a `MultiTaskGP` is present, only one category/task is allowed in the `TaskFeature`, else we will get problems, especially as currently the fidelity/task selection is missing. You need to use a after validator (https://github.com/experimental-design/bofire/blob/6d7e065a14a403e5bbef2935631e1fe4bf1e4745/bofire/data_models/strategies/predictives/botorch.py#L105)
- Can you add to the notebook optimization campaigns that show that using information from lower fidelity enhances the convergence in comparison to not using this information?

Best,

Johannes

jpfolch:
Hi @jduerholt,

I have made the requested changes, I am a little unsure about the `_predict` method in [bofire/strategies/predictives/botorch.py](https://github.com/experimental-design/bofire/pull/439/files/6065dba0dd0404d78a5e0e9b228d1976963d6470#diff-689fd2f17db70035399835d2f7b99bc3de8e20b9dd39a31aad0ce7fc66bbda9b)[bofire/strategies/predictives/botorch.py](https://github.com/experimental-design/bofire/pull/439/files/6065dba0dd0404d78a5e0e9b228d1976963d6470#diff-689fd2f17db70035399835d2f7b99bc3de8e20b9dd39a31aad0ce7fc66bbda9b) being correct, I left some more details in the comment above.

I also added some testing, checking that the `sobo` and `qehvi` work okay with MultiTask surrogates, they seem to do! Let me know if I should move the testing to a specs file somewhere, I searched for it but couldn't really pinpoint exactly where they should go so I just tested in a separate file.

bofire/strategies/predictives/botorch.py
jpfolch: Oh, also important to mention that when I tested with a single surrogate model, or two of them, in all cases it goes through `len(posterior.mean.shape) == 2` which caused my confusion...

jduerholt:
Thanks, I will have a look the latest on Monday, but I hope to get it done tomorrow!

jduerholt (changes requested):
Hi @jpfolch,

thank you very much. Looks very good. I think  we are almost there. Have a look at my inline comments.

Regarding the notebook, only two things:
- Can you also add a "SMOKE_TEST" limitation for the iterations that you run for the notebook? So that in case of the automatest test pipleine, only two iterations per optimiuation campagin are run?
- There is a variable called `regrets_single_task_mean`. Can you rename it to `regrets_single_task_median` as it is actually the median?
- Optional one could think of creating a benchmark class for you multitask benchmark that you created and then use the automatic benchmark excutors. But this is only optional.

Best,

Johannes

bofire/data_models/strategies/predictives/botorch.py
```diff
@@ -227,3 +229,17 @@ def _generate_surrogate_specs(
         surrogate_specs.surrogates = _surrogate_specs
         surrogate_specs._check_compability(inputs=domain.inputs, outputs=domain.outputs)
         return surrogate_specs
+
+    @model_validator(mode="after")
```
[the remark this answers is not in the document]
jpfolch: > This has to be moved to `data_models/strategies/predictives/botorch.py`, as it is just a requirement for optimization.

Is it not already in the correct file? I am adding the test now though.

jpfolch:
Hi @jduerholt,

I have made the requested changes. A few points:

- I have made the `try` and `except` changes for prediction using MultiTaskGPs both in `bofire/surrogates/multi_task_gp.py` and `bofire/strategies/predictives/botorch.py` for consistency.

- I was unable to test that if a list of surrogates is used that includes a MultiTaskGP, then all models should be MultiTaskGPs using the specs; I kept getting errors. I noticed that none of the specs in `tests/bofire/data_models/specs/surrogates.py` have validation testing for the `BotorchSurrogates` surrogate and that these are instead tested in `tests/bofire/surrogates/test_torch_models.py` so that is where I added the testing. Let me know if this is okay, or if you have  a guess of what was causing the errors.

- The "SMOKE_TEST" variable in the notebook was included in a cell further down, let me know if you'd prefer it at the beginning.

- I am happy to write MultiTask / Multifidelity benchmarks in a future PR that are more complete, rather than my current silly examples, I will have a think of which ones would be best to add.

Thanks!
Best,
Jose

jduerholt:
Looks very good for me. Thank you very much!

jduerholt (approved):

PR #440, opened 2024-09-13
EqalityConstraint --> EqualityConstraint

jduerholt:
This PR fixes a typo, it is `EqualityConstraint` instead of `EqalityConstraint`. Thanks for the catch @Osburg!


Osburg (approved):
Thanks, @jduerholt !

PR #441, opened 2024-10-02
Move variance of fixed features check to predictive strategy

bertiqwerty:
I think it is okay to check for variance in fixed features only in predictive strategies. This helps to add data from only one task to the random strategy (or any strategy without surrogate) in case you have a multi-task domain. This might be useful in case you have a stepwise strategy with a random step to get initial random data from one of the tasks where you might not have historical data of.

### DIFF
--- a/bofire/strategies/predictives/predictive.py
+++ b/bofire/strategies/predictives/predictive.py
@@ -5,6 +5,7 @@ import numpy as np
 import pandas as pd
 from pydantic import PositiveInt
 
+from bofire.data_models.features.task import TaskInput
 from bofire.data_models.strategies.api import Strategy as DataModel
 from bofire.data_models.types import InputTransformSpecs
 from bofire.strategies.data_models.candidate import Candidate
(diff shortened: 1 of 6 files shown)


bofire/strategies/predictives/predictive.py
```diff
@@ -83,6 +84,25 @@ def tell(
...
         else:
             self.add_experiments(experiments)
+        # we check here that the experiments do not have completely fixed columns
+        cleaned_experiments = (
+            self.domain.outputs.preprocess_experiments_all_valid_outputs(
+                experiments=experiments
+            )
+        )
+
+        fixed_nontasks = (
```
jduerholt, line 94: Just one question: what is the purpose of a multitask optimization when your historical data is only for one task and you only want candidates in the future for the same task?

jduerholt (approved):
Looks fine for me, only see the question below ;)

bofire/strategies/predictives/predictive.py
bertiqwerty: Sorry. I wasn't clear in my formulation above. I have historical data of task 1 and I want to optimize task 2.

PR #442, opened 2024-10-02
Moving Output Objective

jduerholt:
This PR adds the possibility of smart objectives that change during the optimization campaign. So far only the `MovingMaximizeSigmoidObjective` is implemented (out of an actual need). It changes its turning point based on the best so far achieved value.


bertiqwerty (changes requested):
Hi Johannes, Thanks. Also this new feature would benefit from documentation/example, I think.

bertiqwerty (approved):
After discussing with Johannes we decided to first merge this. Johannes will add a more extensive documentation afterwards.

Issue #444, opened 2024-10-07
Linting of notebooks under tutorials and swapping qEI to qLogEI


CompRhys:
I am starting to write initial scaffold code using bofire for our work and have been looking at several of the tutorials.

 Would a PR be accepted that linted the notebooks in the pre-commit?

Would you be happy to swap qEI for qLogEI in the tutorial notebooks given that that's recommended best practice? 

```python
miniconda3/lib/python3.12/site-packages/botorch/acquisition/monte_carlo.py:393: NumericsWarning: qExpectedImprovement has known numerical issues that lead to suboptimal optimization performance. It is strongly recommended to simply replace

	 qExpectedImprovement 	 --> 	 qLogExpectedImprovement 

instead, which fixes the issues and has the same API. See https://arxiv.org/abs/2310.20708 for details.
```

CompRhys:
both tasks are finished

PR #445, opened 2024-10-07
Applying linting to notebooks

CompRhys:
Updates to use a more recent version of ruff from astral rather than the initial action written for early version before astral was formed around ruff. Applies ruff to notebooks inside the repo.

Added nbstripout action to remove unnecessary information from the tutorial notebooks to reduce the diffs that are expected when run by different users


jduerholt:
Hi @CompRhys,

thanks for your PR. I am unsure about the `nbstripout`, as I find it nice to have the outputs in the notebooks for didactic purposes, especially as we link then sometimes also directly into mkdocs. @bertiqwerty: what do you think?

It seems that you are much more experienced with Ruff than me, what do you think of using Ruff also for formatting and get rid of black? https://docs.astral.sh/ruff/formatter/

Would you mind to implement this?

Best,

Johannes

bertiqwerty:
> Hi @CompRhys,
> 
> thanks for your PR. I am unsure about the `nbstripout`, as I find it nice to have the outputs in the notebooks for didactic purposes, especially as we link then sometimes also directly into mkdocs. @bertiqwerty: what do you think?
> 
> It seems that you are much more experienced with Ruff than me, what do you think of using Ruff also for formatting and get rid of black? https://docs.astral.sh/ruff/formatter/
> 
> Would you mind to implement this?
> 
> Best,
> 
> Johannes

Hi all,

I am in favor of using a new Ruff-version and also of using `ruff format` instead of Black. 

About the notebooks, I agree with Johannes that it is good to show the output.

CompRhys:
Okay! sure can update the linting actions to make use of the pre-commit settings such that there's only one place where the ruff version is controlled. I have removed all the obvious `# noqa: <xyz>` and `# type: ignore` comments and then will go back and try fix/re-add where necessary for pyright. Switched to using the pyright ci action suggested by microsoft https://github.com/microsoft/pyright/blob/main/docs/ci-integration.md. When running the notebooks they're spammed with warnings about EI vs LogEI, I will rerun to generate the outputs but first do you guys agree that it makes sense to swap all the defaults/tutorials to use LogEI? #444 and https://github.com/Radical-AI/bofire/tree/logei-defaults

.github/workflows/lint.yaml
```diff
@@ -1,37 +1,38 @@
...
+        with:
+          enable-cache: true
+
+      - name: Install dependencies
+        run: uv pip install pre-commit ".[all]" --system
+
+      - name: Run pre-commit
+        run: pre-commit run --all-files --show-diff-on-failure
+
+      - name: Pyright
```
CompRhys, line 35: ideally we could move this inside the pre-commit action but there appear to be issues with getting pyright to work with pre-commit

bofire/outlier_detection/outlier_detection.py
```diff
@@ -83,12 +85,16 @@ def detect(self, experiments: pd.DataFrame) -> pd.DataFrame:
...
+                        experiments[self.base_gp.outputs.get_keys()[0]]
+                        - pred[self.base_gp.outputs.get_keys()[0] + "_pred"]
+                    )
+                    ** 2
+                    / pred[self.base_gp.outputs.get_keys()[0] + "_sd"] ** 2
                 )
-                ** 2
-                / pred[self.base_gp.outputs.get_keys()[0] + "_sd"] ** 2
-            ).ravel()
+                .to_numpy()
```
CompRhys, line 95: manual change check against tests

CompRhys:
I think the tests that were failing are flaky, here they passed but the same code being tested on my fork failed.

jduerholt:
> I think the tests that were failing are flaky, here they passed but the same code being tested on my fork failed.

Yeah, there are some flaky tests, espececially in the DoE module, but @Osburg is currently working on a a refactoring of the whole module which will hopefully also cure the flaky tests ;)

Thanks for the PR. @bertiqwerty, I think it makes sense that you also have a look on it as most of the pipelines and checks were implemented by you.

jduerholt:
One think that catched my attention:

What do you think of adding a pyright section to `pyproject.toml` (https://github.com/microsoft/pyright/blob/main/docs/configuration.md). There we could then for example define that it is not throwing an error when one overrites variables or methods, which would significantly reduce the number of `type: ignore` in the code (https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportIncompatibleMethodOverride).

bertiqwerty:
> One think that catched my attention:
> 
> What do you think of adding a pyright section to `pyproject.toml` (https://github.com/microsoft/pyright/blob/main/docs/configuration.md). There we could then for example define that it is not throwing an error when one overrites variables or methods, which would significantly reduce the number of `type: ignore` in the code (https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportIncompatibleMethodOverride).

I think we should keep the type ignores because incompatible override is not very nice and at least we should mark it as such.

jduerholt:
> I think we should keep the type ignores because incompatible override is not very nice and at least we should mark it as such.

Hmm, but we do it a lot in the data models (to guarantee proper serialization and deserialization using the `type` attribute), just have a look at the new `type: ignore` added in this PR ...

bertiqwerty:
This just means that we have a lot of type violations because a `Literal` is not a `str`. Maybe we should replace `type: str` by `type: Any` in the base class because the derived types are never `str` but always some kind of `Literal[...]`. I have not found a way to restrict this to any kind of `Literal`.

jduerholt:
Hmm, ok. Then we keep it as it is currently and perhaps fix this in a later PR.

CompRhys:
So when looking at pyright I removed them all and then added all the ones that created issues back and then added a bunch of extra ones where it was now complaining. I wasn't really sure as to why the changes I made created new typinging issues as I only manually changed one bit of logic to cast an dataframe to numpy before calling ravel due to an incoming pandas deprecation.

If you guys are happy with this PR as it is I think it would be good to merge and then this pyright stuff and then also changing the tutorials to follow the LogEI best practice can be done in another PR(s)

jduerholt:
I think, that there is a new pyright version, which is flagging more than the old one, that we have used so far ....

CompRhys:
I pinned the version in the action to the version that was in the `setup.py` file previously so for the action I don't think so but perhaps as I was fixing locally I probably am using the most recent version due to pylance which might be the cause

jduerholt (approved):
Thanks for this effort @CompRhys . Looks good to me! After @bertiqwerty had a look we can merge it ;) Best, Johannes

bertiqwerty (approved):
Thanks a lot @CompRhys. Looks good to me!

PR #448, opened 2024-10-11
Refactoring of `DoEStrategy` data model class

jduerholt:
This PR provides a draft for the restructured `DoEStrategy`.

The following things changed:
- The `objective` is now `criterion` to prevent confusion with the objectives of the output features.
- The former objectives like `DOptimality` are not anymore enums, but classes, for example `DOptimalityCriterion`. This allows for the clean implementation of criterion specific attributes.
- The formula is now anymore a direct attribute of the `DoEStrategy` but instead an attribute of the `criterion`.

When the functional part of the DoE module is refactored, we should have a mapper which takes the criterion and domain and returns a functional criterion class which provides then the actual objective funtion for the actual ipopt optimizer. 

On the long run, we can then also provide a common base class on the data model level for both the `DoEStrategy` and the `SpaceFillingStrategy`. In case of the latter one the only allowed criterion will then be the `SpaceFillingCriterion` which has no `formula` attribute.

@Osburg: What do you think?


dlinzner-bcs:
@jduerholt @Osburg i made some changes. We still want to support the old interface with `find_local_max_ipopt`. For this I allowed mapping a criterion to the old enum. Is this a valid compromise?

dlinzner-bcs:
@jduerholt @Osburg This PR is now ready for merging. Do you approve?

jduerholt:
There are failing tests due to a new sklearn version, I already went into the rabbit hole. Will post an issue later.

bofire/strategies/doe/objective.py
```diff
@@ -83,6 +74,71 @@ def evaluate_jacobian(self, x: np.ndarray) -> np.ndarray:
     def _evaluate_jacobian(self, x: np.ndarray) -> np.ndarray:
         pass
 
+    @abstractmethod
```
Osburg, line 77: Are these functions needed in the Objective class? Since the SpaceFilling objective actually does not have to implement any of these functions, right? (The only function that is used for any objective is get_model_matrix() in the exit message find_local_max_ipopt() i think, but imo it is not necessary there as well...). What is your opinion?

bofire/strategies/doe/objective.py
```diff
@@ -493,19 +552,66 @@ def _convert_input_to_tensor(
         )
         return torch.tensor(X.values, requires_grad=requires_grad, **tkwargs)
 
+    def get_model_matrix(self, design: pd.DataFrame) -> pd.DataFrame:
```
Osburg, line 555: This could then also be removed

bofire/strategies/doe/design.py
```diff
@@ -572,7 +199,7 @@ def find_local_max_ipopt(
     if _ipopt_options[b"print_level"] > 12:  # type: ignore
```
Osburg, line 199: In my opinion this could be removed. The objective criterion can be viewed via the IPOPT output and any other criterion can also be calculated with little effort by creating another `Objective`. What do you think? :) --> then there is also no need anymore to implement get_model_matrix() for `SpaceFilling`.
Also the metrics(), g_optimality(), a_optimality(), d_optimality() functions in doe utils could be deleted then since they are not used anywhere else.

Osburg (commented):
@dlinzner-bcs Thanks!!! :)) <3 looks good to me, left some minor comments. Feel free to ignore them if they are not helpful.
Cheers, Aaron :)

R-M-Lee:
why is the nbstripout precommit hook not being executed properly? The notebook output should not be there. Right now there are some images and outputs referencing local paths on someone's machine

bofire/strategies/space_filling.py
jduerholt, line 1: Can we get rid of this functional strategy here, and just map the data model of the SpaceFillingStrategy to the DoeStrategy in the mapper?

tests/bofire/data_models/specs/strategies.py
jduerholt, line 1: We should also test the optimality criteria data models for serialization and desirilization, should I implement this into this PR? One has to register a few new fixures for this.

jduerholt (commented):
Looks overall very good, I let some questions and I am open to also implement some of the mentioned points ;)

bofire/strategies/space_filling.py
Osburg: I would have liked to use the SpaceFillingStrategy as it is to generate the design space grid for the I-optimality criterion. To avoid circular imports it would be nice if the space filling strategy would be available indepent of the DoeStrategy (@dlinzner-bcs is it okay if I add a commit for the I-optimality - now compatible with your changes - criterion to this branch?).  
Only if there are no good reasons against keeping it this way ofc @jduerholt .

Osburg: I do not know if we need it but it was implemented to make sure that space-filling will give you uniformly random-ish samples of a surface. Because normally the space-filling objective will create a regular grid pattern of approx. equidistant points (--> i.e. not actually random). The sampling_fraction allows to create a grid with many more points than requested (1/sampling_fraction times more points). Assuming that this grid approximates the surface to sample from well, we can then imitate uniform sampling from that surface by randomly drawing points from the grid. Does this make sense to you? 

Cheers
Aaron :)

dlinzner-bcs (approved):
This looks good to go for me now. Important: This has breaking changes for:
1. DoE generation
2. SpaceFillingStrategy which is now implemented via DoE + SpaceFillingCriterion

Should I merge now?

dlinzner-bcs (approved):

dlinzner-bcs:
@Osburg can you please check again and, if applicable, approve? Then I will merge :)

Osburg (approved):
To me it looks good :) Thanks @dlinzner-bcs

tests/bofire/data_models/specs/strategies.py
dlinzner-bcs: I figured out how to add the fixtures :)

PR #453, opened 2024-11-05
Multifidelity Strategy

TobyBoyne:
Implement strategy for optimizing functions with multiple fidelities, using MultiTaskGP.

This implementation is based on [Kandasamy et al. 2016, Folch et al. 2023]. We first optimize the target fidelity to obtain an input $x$, then select the lowest fidelity that gives a variance greater than some threshold. We use a MultiTaskGP to avoid the bias terms in [Kandasamy], and to enable transfer learning across tasks.

Still to do:
- Implement scaling of UCB: the $\beta_t$ in $\beta_t^{1/2} \sigma(x) < \gamma$, eq (21) of [Folch et al. 2023]
- `_ask` currently returns the predicted mean/std at the **target** fidelity - is this behaviour correct?
- Testing, add to data model (e.g. verify that the surrogate must be a MultiTaskGP)
- Batch, asynchronous support (although this is probably for another PR)

Things I'm not sure about:
- The way I handle `allowed` in the `_ask` method seems suspicious. This seemed to me to be the easiest way to fix the fidelity at the target for the optimization, but I don't know if this is the best way.


TobyBoyne:
This implementation assumes an ordering of the fidelities, which should be relaxed since `TaskInput` does not require a fixed order (thanks for pointing this out @jpfolch!)

bofire/strategies/predictives/multi_fidelity.py
```diff
@@ -0,0 +1,70 @@
...
+
+        Args:
+            candidate_count (int): number of candidates to be generated
+
+        Returns:
+            pd.DataFrame: DataFrame with candidates (proposed experiments)
+        """
+        task_feature: TaskInput = self.domain.inputs.get_by_key(self.task_feature_key)  # type: ignore
+        task_feature.allowed = [True] + [False] * (len(task_feature.categories) - 1)
+        x = super()._ask(candidate_count)
```
R-M-Lee, line 35: I have a question about this part: it looks like `task_feature.allowed` doesn't have any influence on the `super()._ask`. Same for selecting the fidelity a couple of lines later. What am I missing here?

R-M-Lee:
> * `_ask` currently returns the predicted mean/std at the **target** fidelity - is this behaviour correct?

Hmm, I think that I would like `_ask` to return predictions for the fidelity that is being proposed. But then this needs to be very clear in the output. Perhaps too confusing. Can we just return all of them? What do you think?

> Things I'm not sure about:
> 
> * The way I handle `allowed` in the `_ask` method seems suspicious. This seemed to me to be the easiest way to fix the fidelity at the target for the optimization, but I don't know if this is the best way.

I added a comment. But I see that you replied to yourself about the assumed ordering of the fidelities, which is obviously related (currently the target fidelity is the first one and then the fidelity descends, correct?) I think this is a sensible default, to be honest. But sure, perhaps we should explicitly store the ordering of the fidelities in the strategy to avoid confusion

Thanks for starting this PR

bofire/strategies/predictives/multi_fidelity.py
TobyBoyne: `task_feature.allowed` has two impacts:
1. By setting `allowed = [True, False, False, ...]`, we fix the fidelity to be the target fidelity (thanks to [1]). This means that when we optimize the function by calling `super()._ask()`, we don't allow the underlying surrogate to optimize over the fidelity (since for the first step of the procedure, we want a candidate x for the target fidelity).
2. Any chosen fidelity must have `allowed[fidelity] == True`, thanks to [2]. We therefore need to set all tasks to be allowed, so that the validation in [2] succeeds

[1] https://github.com/experimental-design/bofire/blob/1266d7109c2af7bd6bd01522145661bf23d5cf1d/bofire/data_models/features/categorical.py#L59-L67
[2] https://github.com/experimental-design/bofire/blob/1266d7109c2af7bd6bd01522145661bf23d5cf1d/bofire/data_models/features/categorical.py#L146-L149

TobyBoyne:
@R-M-Lee  Thank you for your comments!

> > * `_ask` currently returns the predicted mean/std at the **target** fidelity - is this behaviour correct?
> 
> Hmm, I think that I would like `_ask` to return predictions for the fidelity that is being proposed. But then this needs to be very clear in the output. Perhaps too confusing. Can we just return all of them? What do you think?

I would go for your first suggestion (return only predictions for the proposed fidelity). This is because the experiment will look something like the table below. To me it makes sense that you only give one prediction (since you are only carrying out one experiment), and `fidelity` can be considered a feature of the experiment in the same way as `x1` and `x2`.

| fidelity | x1 | x2 | y_pred | y |
| --- | --- | --- | --- | --- |
| 0 | 0.0 | 0.5 | 2.1 | 2.3 |
| 1 | 1.0 | 0.7 | 0.6 | 0.3 |

> > * The way I handle `allowed` in the `_ask` method seems suspicious. This seemed to me to be the easiest way to fix the fidelity at the target for the optimization, but I don't know if this is the best way.
> 
> I added a comment. But I see that you replied to yourself about the assumed ordering of the fidelities, which is obviously related (currently the target fidelity is the first one and then the fidelity descends, correct?) I think this is a sensible default, to be honest. But sure, perhaps we should explicitly store the ordering of the fidelities in the strategy to avoid confusion

I've replied to the comment. Also, I've changed the code so that we no longer assume an ordering of fidelities - we now support any fidelities in the form accepted by `TaskInput`. That is, a list of fidelities that includes the elements `[0, 1, ..., M]` in any order.

bofire/strategies/predictives/multi_fidelity.py
```diff
@@ -0,0 +1,76 @@
...
+        Args:
+            candidate_count (int): number of candidates to be generated
+
+        Returns:
+            pd.DataFrame: DataFrame with candidates (proposed experiments)
+        """
+        if candidate_count > 1:
+            raise NotImplementedError("Batch optimization is not yet implemented")
+        task_feature: TaskInput = self.domain.inputs.get_by_key(self.task_feature_key)  # type: ignore
+        task_feature.allowed = [True] + [False] * (len(task_feature.categories) - 1)
```
jpfolch, line 38: I think this line still assumes the fidelities are ordered?

TobyBoyne: Yes, good spot!

bofire/data_models/strategies/predictives/multi_fidelity.py
```diff
@@ -0,0 +1,20 @@
+from typing import Literal
+
+from pydantic import model_validator
+
+from bofire.data_models.strategies.predictives.sobo import SoboStrategy
+
+
+class MultiFidelityStrategy(SoboStrategy):
+    type: Literal["MultiFidelityStrategy"] = "MultiFidelityStrategy"
```
jduerholt, line 9: `SoboStrategy` is currently allowing for a versatile set of acquisition functions. For this strategy only UCB is allowed or? 

Then we should overwrite here the acqf with:

```
acquisition_function: qUCB= Field(
        default_factory=lambda: qUCB()
    )
```

If you need UCB specific user settings (as discussed somewhere below), you can also add more parameters to the `qUCB` data model.

TobyBoyne: The algorithm consists of first optimizing $x$ on the target fidelity, then optimizing the selection of fidelity, $m$. Since the first step only operates on a single fidelity, we can use any acquisition function that we choose (as noted in Folch et al 2023, sec 5.2).

The selection of $m$ then uses a variance-based acquisition function, which is contained entirely in `MultiFidelityStrategy`.

jduerholt: Thank you very much for the clarification!

TobyBoyne:
Hi @R-M-Lee! I've added some tests and also a notebook to demonstrate the MF working (as a sequel to the transfer learning notebook).

I can't quite work out why the `valid` test isn't working - could you please have a look and see what I'm getting wrong? (in `tests/bofire/data_models/specs/strategies.py`, line 614). Thanks :)

jduerholt:
Hi Toby, you have to register the new strategy data model also here: https://github.com/experimental-design/bofire/blob/main/bofire/data_models/strategies/actual_strategy_type.py

TobyBoyne:
Thanks for the help, @jduerholt! I now have more validation, and some testing specs. I believe this PR is now complete, please let me know if I've missed anything @R-M-Lee!

Batch approaches and the information based approach are left for future PRs.

jduerholt:
Hey @TobyBoyne, I had one quick look. Can you also write some tests that acutally test the functional `MultifidelityStrategy` class, so that we are sure that the ask and the `_select_fidelity_and_get_predict` works? Best, Johannes

R-M-Lee:
I like it @TobyBoyne . I left one question/comment about how the fidelity is selected.

Otherwise it looks good to me

R-M-Lee:
@TobyBoyne another observation: if we do `ask` with an initial dataset that does not include evaluations of all the available tasks, we get an error. 

Here is what I get when I have 4 tasks, but the initial dataset does not include task 2 (ie, only 0, 1 and 3)

image

R-M-Lee:
another thing (sorry to keep coming up with new comments): it seems to me that running ask resets the value of task_feature.allowed to [True] * n_tasks. I think this may be a bug

bofire/strategies/predictives/multi_fidelity.py
```diff
@@ -0,0 +1,82 @@
...
+
+            X_fid = X.assign(**{self.task_feature_key: fidelity_name})
+            transformed = self.domain.inputs.transform(
+                experiments=X_fid, specs=self.input_preprocessing_specs
+            )
+            pred = self.predict(transformed)
+
+            if (pred[sd_cols] > fidelity_threshold).all().all() or m == 0:
+                pred[self.task_feature_key] = fidelity_name
+                return pred
```
R-M-Lee, line 82:
```suggestion
        fidelity_input: TaskInput = self.domain.inputs.get_by_key(self.task_feature_key)  # type: ignore
        assert self.model is not None and self.experiments is not None
        assert fidelity_input.allowed is not None

        sorted_fidelities = np.argsort(fidelity_input.fidelities)[::-1]
        _, sd_cols = get_column_names(self.domain.outputs)

        for fidelity_idx in sorted_fidelities:
            if fidelity_input.allowed[fidelity_idx]:
                m = fidelity_input.fidelities[fidelity_idx]
                fidelity_name = fidelity_input.categories[fidelity_idx]

                fidelity_threshold_scale = self.model.outcome_transform.stdvs.item()
                fidelity_threshold = (
                    self.fidelity_thresholds[m] * fidelity_threshold_scale
                )

                X_fid = X.assign(**{self.task_feature_key: fidelity_name})
                transformed = self.domain.inputs.transform(
                    experiments=X_fid, specs=self.input_preprocessing_specs
                )
                pred = self.predict(transformed)

                if (pred[sd_cols] > fidelity_threshold).all().all() or m == np.min(
                    fidelity_input.fidelities
                ):
                    pred[self.task_feature_key] = fidelity_name
                    return pred
```

R-M-Lee (changes requested):
I think we need to fix the part where I made a rough suggestion. Otherwise it looks great to me!

TobyBoyne:
Thank you @jduerholt and @R-M-Lee for your feedback! These commits contain:
- Informative error message when some fidelities are not observed
- Functional tests
    - Test that an (informative) error is raised if a task is missing
    - Test that an appropriate fidelity is selected (a) near training data and (b) far from training data
    - Test that `.ask()` works (smoke test)
- No longer assume `target_fidelity` is 0 (I think this might be enforced in `TaskInput`, but better to be explicit here)

bofire/strategies/predictives/multi_fidelity.py
TobyBoyne: Added in 1f3a00e

R-M-Lee: looks good. your solution is nicer than my suggestion :)

TobyBoyne: Thank you, but actually I think this might be a bug on my part (I'm checking the index, not the order). Fix incoming soon, I will also change the order of the fidelities in the test to confirm I've got it the right way round!

R-M-Lee:
@TobyBoyne  another thing that is not quite right I think... in the ipynb demo of MFBO, all evaluations in the loop are done at the high fidelity independent of the task proposed by the optimizer (ie, without adding the bias from the other benchmark function). This means that the plot at the end does not mean what one might naively think

TobyBoyne:
@R-M-Lee Thanks for spotting that! I have fixed it, and I still seem to be getting good results!

image

R-M-Lee (approved):

jduerholt:
Hi @TobyBoyne, thank you very much. I will also have a look today. There seems to be still a problem with the tests.  Test pipeline for 3.9 is crashing in this test: tests/bofire/strategies/test_multi_fidelity.py::test_mf_fidelity_selection

TobyBoyne:
> There seems to be still a problem with the tests. Test pipeline for 3.9 is crashing in this test: tests/bofire/strategies/test_multi_fidelity.py::test_mf_fidelity_selection

Thanks @jduerholt , this has now been fixed. In Python 3.9, GPyTorch raises a different error message when trying to *fit* a MultiTaskGP, when later versions only have issues when *ask*ing a GP. I have made a change to the `pytest` to catch both. 

**You can ignore the following, as I don't think it's that important.** But below is a discussion on why I have only fixed the test, and not changed the strategy.

Ideally, I would catch the more cryptic error and output the clearer error message, but I couldn't find a neat way to do this:
1. Changing `has_sufficient_experiments` would raise the vague `"Not enough experiments available to execute the strategy"` in `Strategy.ask`
2. It would be possible to have the `try... except` block in `MultiFidelityStrategy._fit` method, which would then raise the `"Some tasks have no experiments"` iff py3.9 is being used. But this would, in my opinion, be ugly, as it further pollutes the stack with more function calls and specific checks.

It might therefore just be easier to accept that py3.9 will have a worse error message in this case (especially as it is reaching EOL soon anyway).

bofire/strategies/predictives/multi_fidelity.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def _verify_all_fidelities_observed(self) -> None:
+        """Get all fidelities that have at least one observation.
+
+        We use this instead of overriding `has_sufficient_experiments` to provide
+        a more descriptive error message."""
+        assert self.experiments is not None
+        observed_fidelities = set(self.experiments[self.task_feature_key].unique())
+        allowed_fidelities = set(
```
jduerholt, line 100: Why not use this here? https://github.com/experimental-design/bofire/blob/289cf42d8ab2c79be28e9397609abf0c7cc6dac5/bofire/data_models/features/categorical.py#L97

jduerholt (approved):
Thanks @TobyBoyne: looks good to me. Only one minor comment ;)

bofire/strategies/predictives/multi_fidelity.py
TobyBoyne: Good point! Thanks for the comment (and for the presentation today, I enjoyed it :) )

PR #454, opened 2024-11-06
fix linting

jduerholt:

PR #455, opened 2024-11-06
Fix linting in pipeline

bertiqwerty:
Looks like `ruff format` is complaining about `bofire/data_models/surrogates/shape.py`.


bertiqwerty:
Duplicate of #454

PR #457, opened 2024-11-07
Add Trust Regions

CompRhys:
This is a WIP Draft PR to add trust regions.

This introduces a breaking change renaming LSRBO to LSRBOConfig.

TODO:

- [ ] Tests
- [ ] Ideally we could fill up the TR in one shot rather than needing to iterate


CompRhys:
@jduerholt This is now in a "working" state but it has many rough edges, would appreciate a review and thoughts of how some things may be cleaned up

bofire/data_models/strategies/predictives/botorch.py
```diff
@@ -32,6 +34,184 @@
...
+            from its own history. If false (default), all trust regions will
+            fit to available data that overlaps the trust region.
+    """
+
+    type: str
+    length_init: float = 0.8
+    length_min: float = 0.01
+    length_max: float = 1.6
+    lengthscale_adjustment_factor: float = 2.0
+    fit_region_multiplier: float = 2.0
```
jduerholt, line 71:
```suggestion
    fit_region_multiplier: float = Field(default=2.0, ge=1)
```

bofire/data_models/strategies/predictives/botorch.py
```diff
@@ -32,6 +34,184 @@
...
+    experiment_batch_sizes: list[list[PositiveInt]] = Field(
+        default_factory=lambda: [[]]
+    )
+    use_independent_tr: bool = False
+
+    length: float = Field(default=length_init, alias="length_")
+    X_center_idx: int = -1
+
+    @model_validator(mode="after")
+    def validate_fit_region_multiplier(self):
```
jduerholt, line 88: You can remove this.

jduerholt: It is catched by the suggestion from above.

bofire/data_models/strategies/predictives/botorch.py
```diff
@@ -32,6 +34,184 @@
...
+            experiments (pd.DataFrame): DataFrame containing experiments that
+                were performed and their results.
+            domain (Domain): The domain defining the problem to be optimized
+                with the strategy.
+        """
+
+        self.length = self.length_init
+
+        Y_cols = domain.outputs.get_keys()
+        if len(Y_cols) != 1:
```
jduerholt, line 132: Ah ok, makes sense. Perhaps one could then also inherit it just from SoboStrategy?

jduerholt: Or build in there?

bofire/data_models/strategies/predictives/botorch.py
```diff
@@ -137,6 +317,18 @@ def is_constraint_implemented(cls, my_type: Type[Constraint]) -> bool:
...
+    def validate_exclusive_local_search_trust_region(self):
+        # dont allow both local search and trust region to be used at the same time
+        if (
+            self.local_search_config is not None
+            and self.trust_region_config is not None
+        ):
+            raise ValueError(
+                "Local search and trust region optimization cannot be used at the same time."
+            )
+        return self
```
jduerholt, line 330: If we do it as we do it currently, we should definitly validate that we only have one output feature if we are using TURBO.

jduerholt: Based on my understanding TURBO only works with a GP surrogate with a lengthscale based kernel, so we should check the surrogate and its kernel.

bofire/strategies/predictives/botorch.py
```diff
@@ -262,29 +291,28 @@ def _setup_ask(self):
...
         ):
             fixed_features = self.get_fixed_features()
             fixed_features_list = None
         else:
             fixed_features = None
             fixed_features_list = self.get_categorical_combinations()
+
         return (
             bounds,
-            local_bounds,
```
jduerholt, line 287: Where is `local_bounds` gone?

bofire/strategies/predictives/botorch.py
```diff
@@ -332,91 +360,187 @@ def _optimize_acqf_continuous(
...
+            return_best_only=True,
+            options=self._get_optimizer_options(),  # type: ignore
+            ic_generator=ic_generator,
+            **ic_gen_kwargs,
+        )
+        return candidates, acqf_vals
+
+    def _ask_local_search(
+        self,
+        candidates: Tensor,
```
jduerholt, line 430: Let us call it `global_candidates` or something like this, what do you think?

bofire/strategies/predictives/botorch.py
```diff
@@ -332,91 +360,187 @@ def _optimize_acqf_continuous(
...
         assert candidate_count > 0, "candidate_count has to be larger than zero."
         if self.experiments is None:
             raise ValueError("No experiments have been provided yet.")
 
+        if (
+            self.trust_region_config is not None
+            and not self.trust_region_config.has_sufficient_experiments(
+                self.experiments, self.domain
+            )
+        ):
```
jduerholt, line 493: Can you add a comment to describe in natural language what is the scenario here, and what we are doing in this case?

bofire/strategies/predictives/botorch.py
```diff
@@ -332,91 +360,187 @@ def _optimize_acqf_continuous(
...
+            reference_experiment = self.experiments.loc[
+                self.trust_region_config.X_center_idx
+            ]
+            # instantiate a random strategy and ask it for the new experiments.
```
jduerholt, line 497: I propose to move this as a method to the trust region data model, that receives a domain and spits out the local domain, which we then here path to the random strategy to sample new solutions. What do you think?

bofire/strategies/predictives/botorch.py
```diff
@@ -332,91 +360,187 @@ def _optimize_acqf_continuous(
...
+            )
+            # we have to ask for candidate_count to not break the validation elsewhere
+            # in bofire. This is inefficient and breaks the low discrepancy sampling
+            # benefits of using a single sobol sequence.
+            candidates = sampler.ask(candidate_count)
+            for key in self.domain.outputs.get_keys():
+                candidates[f"{key}_pred"] = 0
+                candidates[f"{key}_sd"] = 0
+                candidates[f"{key}_des"] = 1
+                candidates[f"valid_{key}"] = 1
```
jduerholt, line 541: Candidates do not need a valid key, this line can be removed.

bofire/strategies/predictives/botorch.py
```diff
@@ -491,30 +614,17 @@ def _ask(self, candidate_count: int) -> pd.DataFrame:  # type: ignore
             and has_local_search_region(self.domain)
             and candidate_count == 1
         ):
-            local_candidates, local_acqf_val = self._optimize_acqf_continuous(
+            return self._ask_local_search(
+                candidates=candidates,
```
jduerholt, line 618:
```suggestion
                candidates=global_candidates,
```

jduerholt (commented):
Hi Rhys,

thank you very very much. I did a first pass (not looking at tests so far) and really like it. Just have a look at my comments.

Best,

Johannes

bofire/strategies/predictives/botorch.py
jduerholt: Found it ;)

bofire/strategies/predictives/botorch.py
CompRhys: Essentially if you're new best solution sits at the edge of the old optimization box you might not have a sufficient number of points to fit a reasonable surrogate. In the original TurBO work I believe that the model was still fit on all the data in the history of a given trust region so this wasn't an issue. I am partial to the local model idea in MORBO where you take experiments within 2L of the reference point to train on (this will scale better in high-throughoput experimental settings as M^3 <<< N^3 for M<N). In that instance you want to sobol or otherwise sample a new set of experiments until you have enough to trust a local surrogate. 

This is a part where bofire wasn't that friendly because bofire validates that you return the requested number of candidates whereas I would want to return all the missing experiments in one shot.

bofire/data_models/strategies/predictives/botorch.py
CompRhys: Yes so the reference method does indeed scale the inputs to a hypercube and then warp the trust region based on the normalized lengthscales of the GP. From my reading I think the scaling to a hypercube at the start in important but the warping of the box is optional (it's not used in morbo for instance).

The issue is that I am not sure how to add these things here with the minimal footprint. It's in general the challenge of why things like TurBO and MORBO aren't in Ax

bofire/strategies/predictives/botorch.py
CompRhys: Okay! I am honestly not sure what f"{key}_sd" and f"{key}_des" are fwiw, I couldn't find it clearly written anywhere and am not sure what they are and whether these placeholders matter, I guess sd is standard deviation but not sure about des.

bofire/strategies/predictives/botorch.py
CompRhys: I took a quick look it this and it will introduce circular dependencies

bofire/data_models/strategies/predictives/botorch.py
CompRhys: This should probably be an abstract method here and then move this init to the `TuRBOConfig`

jduerholt: Yeah, I think this would make sense to have it as abstract method, especially as there and move it to `TuRBOConfig`, expecially as the there are TR methods that support multiobjective optimization.

bofire/data_models/strategies/predictives/botorch.py
jduerholt: The turbo validation is moved to the `TURBOConfig` (see comment above), we could then also add to the Turboconfig, if we want to have the warping of the box or not. Based on this hyperparam we could then check if the chosen surrogate is suitable. What do you think?

bofire/strategies/predictives/botorch.py
jduerholt: Ah ok, should we then also add a flag to the `TurBOConfig` that toggles between using local models for the TRs or one global model? This would be nice as we could then reproduce the literature to see if it actually works and gives even more flexibility. 

You are right regarding the bofire friendliness here, we could add a configuration flag to toggle this validation behavior.

jduerholt:
@CompRhys: I let some comments, when we have decided on how to handle these things, I would then have a look at the tests in the next iteration.

bofire/strategies/predictives/botorch.py
CompRhys: We can toggle the behaviour depending on `fit_region_multiplier` being `None`

jduerholt:
@CompRhys: short info: I try to get over it, today or tomorrow. Best, Johannes

bofire/strategies/predictives/botorch.py
```diff
@@ -332,91 +360,202 @@ def _optimize_acqf_continuous(
...
 
+        if (
+            self.trust_region_config is not None
+            and not self.trust_region_config.has_sufficient_experiments(
+                self.experiments, self.domain
+            )
+        ):
+            # fill the trust region with random samples if we're using a trust
+            # region approach and don't have enough experiments to fit a
+            # reasonable GP surrogate
```
jduerholt, line 556: I think this linting error is caused by starting the comment with the word `region`: https://github.com/microsoft/pylance-release/issues/3765. Just format it differently.

bofire/strategies/predictives/botorch.py
jduerholt: @CompRhys :

Yeah, we should document it:

- `_pred`: predicted value
- `_sd`: predicted standard deviation
- `_des`: stands for desirability, and it is the predicted objective value, from the objective module attached to a `ContinuousOutputFeature` . It is called `_des` instead of `_obj` just for legacy and tech debt reasons ...

tests/bofire/data_models/specs/local_search_config.py
jduerholt, line 1: We should have something like this also for the TurboConfig. But is is a bit tedious to add a new spec type in the test suite. But as soon as it is added serialization, desrialization and validation is explicitly tested. I will add this to the PR when the rest is ready, so you do not need to bother.

tests/bofire/data_models/specs/strategies.py
jduerholt, line 1: Could you add here at least one valid SOBO Strategy example with a valid trust region config to test the serialization etc. for it? And also invalid combinations to test the validators added to the `BotorchStrategy`regarding the trust region.

jduerholt (changes requested):
Hi @CompRhys,

this looks really nice. I let some comments, could then add tets for the new functionality especially the methods regarding the TR ;)

I really like it!

Best,

Johannes

bofire/strategies/predictives/botorch.py
```diff
@@ -332,91 +360,202 @@ def _optimize_acqf_continuous(
...
+                            inp.lower_bound,  # type: ignore
+                        ),
+                        min(
+                            reference_experiment[inp.key]
+                            + self.trust_region_config.length / 2,  # type: ignore
+                            inp.upper_bound,  # type: ignore
+                        ),
+                    ),
+                )
+                for inp in inputs.get(ContinuousInput)
```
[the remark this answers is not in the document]
CompRhys: Actually I think that the PR doesn't work because I've missed the logic here to actually adjust the `local_relative_bounds` of the domain based on the turbostate. 

The intent was that it would only impact the ContinuousInputs as was attempting to reuse the local search stuff. If local search is validated in that way then can also add that here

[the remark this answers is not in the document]
jduerholt: Good point with updating the local_relative_bounds!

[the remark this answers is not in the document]
jduerholt: And I think, you also need to something to change in line 627 where the local search is started, furtheremore also the optimize_local needs some update so that it works both for TR and LSRBO?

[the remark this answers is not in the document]
CompRhys: I've already put a validation that you can't do both at the same time in the `BotorchStrategy` data model. In principal you might be able to remove this in the future but lets start with the minimal case

```python
    def validate_exclusive_local_search_trust_region(self):
        # dont allow both local search and trust region to be used at the same time
        if (
            self.local_search_config is not None
            and self.trust_region_config is not None
        ):
            raise ValueError(
                "Local search and trust region optimization cannot be used at the same time."
            )
        return self
```

CompRhys:
Not sure why it's trying to test the base class and then giving validation errors

```python
strategy_spec = SoboStrategy: {'domain': {'type': 'Domain', 'inputs': {'type': 'Inputs', 'features': [{'type': 'ContinuousInput', 'key... 'failure_counter': 0, 'experiment_batch_sizes': [[]], 'use_independent_tr': False, 'length': 0.8, 'X_center_idx': -1}}

    def test_strategy_should_be_serializable(strategy_spec: Spec):
        spec = strategy_spec.typed_spec()
>       obj = strategy_spec.cls(**spec)
E       TypeError: Can't instantiate abstract class TrustRegionConfig without an implementation for abstract methods 'init_trust_region', 'update_trust_region', 'validate_domain', 'validate_surrogates'

tests/bofire/data_models/serialization/test_serialization.py:60: TypeError
```

bofire/data_models/strategies/predictives/botorch.py
```diff
@@ -32,6 +34,237 @@
...
+        elif self.failure_counter == self.failure_streak:  # Shrink trust region
+            self.length /= self.lengthscale_adjustment_factor
+            self.failure_counter = 0
+
+        self.best_value = min(self.best_value, Y_best)  # type: ignore
+        if self.length < self.length_min:
+            self.experiment_batch_sizes.append([])
+
+        for inp in domain.inputs.get(ContinuousInput):
+            inp.local_relative_bounds = (self.length / 2, self.length / 2)
```
CompRhys, line 263:
```suggestion
            inp.local_relative_bounds = (self.length / 2, self.length / 2)  # type: ignore
```

bofire/data_models/strategies/predictives/botorch.py
```diff
@@ -50,28 +283,27 @@ def is_local_step(self, acqf_local: float, acqf_global: float) -> bool:
...
-    type: Literal["LSRBO"] = "LSRBO"
+    type: Literal["LSRBOConfig"] = "LSRBOConfig"
     gamma: Annotated[float, Field(ge=0)] = 0.1
 
     def is_local_step(self, acqf_local: float, acqf_global: float) -> bool:
         return acqf_local >= self.gamma
 
 
-AnyLocalSearchConfig = LSRBO
+AnyLocalSearchConfig = Union[LSRBOConfig]
```
CompRhys, line 305:
```suggestion
AnyLocalSearchConfig = LSRBOConfig
```

bofire/data_models/strategies/predictives/botorch.py
```diff
@@ -50,28 +283,27 @@ def is_local_step(self, acqf_local: float, acqf_global: float) -> bool:
...
+AnyTrustRegionConfig = Union[TuRBOConfig]
```
CompRhys, line 306:
```suggestion
AnyTrustRegionConfig = TuRBOConfig
```

jduerholt:
Hi @CompRhys, I saw that you made some commits but also that you transferred it into a draft PR. Just ping me here, when I should review again! Best, Johannes

CompRhys:
I want to actually have a chance to check that it works rather than just doesn't fail before moving to have it merged, thought it best to make as draft whilst not actively working on it

jduerholt:
Ok, thanks for the info! I saw some nice stuff from you regarding SEBO in botorch/Ax. Nice! At some point, I also have to make it available here ;)

jduerholt:
It appears to me that there are also quite a few changes needed in botorch itself to make SEBO really usable. Can you give me a "ping" when you have the feeling that it is now really usable? So that I can try to port it also to BoFire. And ofc it would be nice to also get this TR PR here at some point ready, as I really like it!

CompRhys:
The majority of those changes do not impact how SEBO might be implemented here because the homotopy stuff is in botorch and then it's just using qnehvi so would require a similar amount of wrapper code here to implement. 

I would say that it would be a great thing to add given that sparsity in high-dimensional formulation/mixtures problems is common in material science.

A missing part is a a sum-to-one constraint an equality constraints aren't yet in Ax. Not sure if that's an API design or bandwidth issue - potentially also just that equality constraints are harder to optimize so maybe ties to your `cyipopt` stuff relevant. And then you're into the rabbit whole of it being a pain to package `cyipopt` as you can't install via pip.

jduerholt:
I agree, so far we do not had so many high-dim optimizations, but I think it would be a great benefit. And I think it should be relatively easy to implement in BoFire. I have an idea ;) We will have a hackathon next Thursday (on-site, in Germany), and I will try to come up with a solution then. So I hope, that we have Sebo in BoFire at the end of next week together with all the other constraints ;)

CompRhys:
Looking forward to it, this PR https://github.com/pytorch/botorch/pull/2588 to botorch should have exposed everything needed to allow all the constraints allowed by botorch.

jduerholt:
Yeah, I saw this, optimization over mixed domains is still missing there to have it fully compabible with BoFire. But I created already a PR for it: https://github.com/pytorch/botorch/pull/2639

PR #458, opened 2024-11-12
Fix enum serialization problems

jduerholt:
As outlined here https://stackoverflow.com/questions/69541613/how-to-json-serialize-enum-classes-in-pydantic-basemodel, the `json` module has problems with serializing plain enums, so far we had to use always pydantics `model_dump_json`. This can lead to problems and confusion. 

Setting up enums which also inherit from `str` solves this issue. In py3.11 is a new datatype available which does exactly this called `StrEnum` (https://docs.python.org/3/library/enum.html#enum.StrEnum). But due to backwards compatibility, I implemented the solution to explicitly inherit from `str`.

### DIFF
--- a/bofire/data_models/enum.py
+++ b/bofire/data_models/enum.py
@@ -1,13 +1,13 @@
 from enum import Enum
 
 
-class SamplingMethodEnum(Enum):
+class SamplingMethodEnum(str, Enum):
     UNIFORM = "UNIFORM"
     SOBOL = "SOBOL"
     LHS = "LHS"
 
 
-class CategoricalMethodEnum(Enum):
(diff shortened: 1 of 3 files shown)


bertiqwerty:
Thanks Johannes. What do you mean with
>  so far we had to use always pydantics model_dump_json. This can lead to problems and confusion.
?

bertiqwerty:
Does it work with old BoFire objects that already have been stored somewhere?

jduerholt:
Could lead to some Problems. But I think it is still good to merge it in, as this creates in general less problems. One can write a two liner to fix the old jsons.

bertiqwerty (approved):

PR #460, opened 2024-11-15
Add pandas data manipulation transform

jduerholt:
- Adds a `ManipulateDataTransform`: This new transform can be used to manipulate `experiments` and `candidates` within a `StepwiseStrategy` by using string expressions in combination with `pandas.DataFrame.eval`. 
- It refactors the `StepwiseStrategy` in a way that it is not necessary that the domains of all strategies have to be equal to each other. They only have to be *compatible* to each other. See  method`validate_domain_compatibility` for further details. This means that for different steps, different representations can be used as long as the final candidates generated in a step are valid by means of the master domain of the strategy. This can be assured by using proper untransformations of the candidates.


bertiqwerty:
Thanks Johannes! I appreciate removing the too strict requirements of the `Domain` to be equal across all steps. I left some comments.

jduerholt:
Hi @bertiqwerty, I do not see any comments ;) Best, Johannes

bertiqwerty (approved):

PR #464, opened 2024-11-19
update entmoot version

R-M-Lee:


R-M-Lee:
Hi @TobyBoyne , could you please take a look? Do we need a new entmoot release to fix this? I would like to use entmoot via bofire with pyomo and Gurobi cloud in the background as per our exchange a few months ago

TobyBoyne:
Hi @R-M-Lee. If I remember correctly, Pyomo 6.7.1 introduced a bug that broke ENTMOOT. We then fixed the version of Pyomo to 6.7.0 in this PR (https://github.com/cog-imperial/entmoot/pull/41). I think that we just need to fix Pyomo to the same version in BoFire in order to resolve this.

R-M-Lee:
Thanks for the reply! I'll give it a try

R-M-Lee:
@TobyBoyne there are some tests failing for what I believe is an unrelated reason and we'll clear that up separately. But one ENTMOOT-specific thing caught my eye in https://github.com/experimental-design/bofire/actions/runs/11936229378/job/33306622223?pr=464

`FAILED tests/bofire/strategies/test_enting.py::test_domain_to_problem_config - ValueError: No support for feat_type 'binary'. Check feature 'if3'.
FAILED tests/bofire/strategies/test_enting.py::test_convert_constraint_to_entmoot - ValueError: No support for feat_type 'binary'. Check feature 'if3'.`

ENTMOOT doesn't like binary features apparently, but one is getting created when we do the domain conversion. Were there changes related to this in ENTMOOT too?

TobyBoyne:
If I had to guess, it's do to with these lines [1]. The logic is a bit curious - we only check to see if the feature is binary *if* no bounds are provided. Since I believe the bofire implementation provides the trivial bounds `[0, 1]`, we never actually test if the feature is binary.

Some quick fixes would be:
- In BoFire, change the bounds for a binary feat to be `None` [2]
- In ENTMOOT, fix this incorrect logic.

A long term fix would be to use the BoFire feature API in ENTMOOT directly. I'm not sure how you would feel about this circular dependency (I personally think it would be fine, since we would only need the minimal BoFire installation in ENTMOOT), however this would take a non-trivial amount of time to implement (discussion at https://github.com/cog-imperial/entmoot/issues/45)

[1] https://github.com/cog-imperial/entmoot/blob/4057ae37a4b84fc148032ce60e5f71990f490986/entmoot/problem_config.py#L287-L288

[2] https://github.com/experimental-design/bofire/blob/d2bd87d0139072a0c740e9d88314f4ebe2b62208/bofire/strategies/predictives/enting.py#L110

PR #465, opened 2024-11-20
`get_by_keys` with exclude option

jduerholt:
Just a small QOL feature, we can now also exclude by feature keys.


bertiqwerty (approved):
Thanks, Johannes.

PR #466, opened 2024-11-22
Suppress import warning prints

chrihaas:
Suppress import warnings related to the packages rdkit, modred, xgboost, and entmoot after installing BoFire via

`pip install bofire[optimization]`


jduerholt (approved):
Thanks Christian.

PR #467, opened 2024-11-22
[QOL] cancel tests on same PR

CompRhys:
Often I end up pushing then spotting an issue that's quick to fix. In the instance that pushes are made in quick succession this will reduce the CI bill (free? but still worth saving energy)


jduerholt (approved):
Nice! Thank you very much. So far, CI ist still free ;)

PR #470, opened 2024-12-03
Fix NumberOfExperimentsCondition

chrihaas:
Fix condition, by shifting the conditional state by one.

In a stepwise strategy, the next step should be reached if n_experiments_step = n_experiments_available.


jduerholt (approved):
Thank you Christian.

jduerholt:
@bertiqwerty: also fine for you?

bertiqwerty:
Hmm... There might be some places in our code where we relied on the old behavior, I think. But in principle, I agree with the change. So let's merge it and we fix our code.

bertiqwerty (approved):
Thanks, Christian.

jduerholt:
> Hmm... There might be some places in our code where we relied on the old behavior, I think. But in principle, I agree with the change. So let's merge it and we fix our code.

Same for us ;) But we will also fix it.

PR #476, opened 2024-12-04
kernels working on a given set of features

e-dorigatti:


e-dorigatti:
NB - tests for this feature are missing

e-dorigatti:
Example with mol features:

```
import pandas as pd

import bofire.strategies.api as strategies
from bofire.data_models.domain import api as domain_api
from bofire.data_models.features import api as features_api
from bofire.data_models.kernels import api as kernels_api
from bofire.data_models.molfeatures import api as molfeatures_api
from bofire.data_models.priors.api import HVARFNER_LENGTHSCALE_PRIOR
from bofire.data_models.strategies import api as strategies_api
from bofire.data_models.surrogates import api as surrogates_api


domain = domain_api.Domain(
    inputs=domain_api.Inputs(
        features=[
            features_api.ContinuousInput(key="x1", bounds=(-1, 1)),
            features_api.ContinuousInput(key="x2", bounds=(-1, 1)),
            features_api.CategoricalMolecularInput(
                key="mol", categories=["CO", "CCO", "CCCO"]
            ),
        ]
    ),
    outputs=domain_api.Outputs(features=[features_api.ContinuousOutput(key="f")]),
)


strategy = strategies.map(
    strategies_api.SoboStrategy(
        domain=domain,
        surrogate_specs=surrogates_api.BotorchSurrogates(
            surrogates=[
                surrogates_api.SingleTaskGPSurrogate(
                    inputs=domain.inputs,
                    outputs=domain.outputs,
                    input_preprocessing_specs={
                        "mol": molfeatures_api.Fingerprints(),
                    },
                    kernel=kernels_api.AdditiveKernel(
                        kernels=[
                            kernels_api.RBFKernel(
                                ard=True,
                                lengthscale_prior=HVARFNER_LENGTHSCALE_PRIOR(),
                                features=["x1", "x2"],
                            ),
                            kernels_api.TanimotoKernel(
                                features=["mol"],
                            ),
                        ]
                    ),
                )
            ]
        ),
    )
)


strategy.tell(
    experiments=pd.DataFrame(
        [
            {"x1": 0.2, "x2": 0.4, "mol": "CO", "f": 1.0},
            {"x1": 0.4, "x2": 0.2, "mol": "CCO", "f": 2.0},
            {"x1": 0.6, "x2": 0.6, "mol": "CCCO", "f": 3.0},
        ]
    )
)
candidates = strategy.ask(candidate_count=1)
print(candidates)
```

jduerholt:
@e-dorigatti: This looks really great and will make the GP section so much easier, we can then get rid of a lot of GP classes! Thank you very much!

e-dorigatti:
Hi @jduerholt thanks for the kind feedback and pointers :) this sounds like a very important use-case to implement and in the next weeks I will look into it. My first idea would be to go via the `input_preprocessing_specs` and try to mimic what `Fingerprints` is doing, probably by recycling some functions from `CategoricalInput`, and indeed creating a new kernel for the Hamming distance.

Going through an input transform sounds (a lot?) more involved, why do you think that would be preferable to the preprocessing specs? Maybe I should learn more about botorch first.. ;)

jduerholt:
Hi @e-dorigatti,

a few more notes/explanations from my side:
- There are two types of transforms in principle:
    1. Everything that we have in the `input_transform_specs`, this is executed **before** the actual optimization/model fitting. For examples, categorical features are transformed/expanded into Onehots or molecular ones into fingerprints, than the model is fitted upon this representation and also the optimizer on top of the acquisition function is working on top of this represenation. Currently, we can give there either one of `CategoricalEncodingEnum`s or one of the molfeatures classes. This is also a bit of inconsistent here, and we have to tidy it up at some point. The molfeatures classes was added later, and we found there that simple enums were not enough as the transform need to be parameterized.
    2. Then we have the botorch input transforms, this is executed on the fly during gp fitting, optimization. A transform(s) is added to the GP and the data is transformed before it is actually entering the kernel. We use them for their flexibility, for example we could have two models in an optimization and the one could be equipped with a `Normalize` transform and the other one with an `InputStandardize` transform. Then one GP would be fitted to the normalized (between 0 and 1) data and the other one to the standardized data (zero mean, unit varriance), and still both of them could be used in combination during optimization. Another example is filtering of features. It could be the case, that we have build two surrogates, but in one of them one feature is not present (perhaps, because we know that it has not predictive power for the output of this model) and now we want to use both surrogates in optimization. In this case, BoFire is automatically combining both Models in botorch's `ModelList` and adds a filter features transform before the first one that filters those features out that should be not part of the optimization (could be in principle also done with the approach here in the PR). Another example is the `PiecewieseLinearGP`. For the optimizer, the input to the model are the nodes of the piecewiselinear, but the nodes are on the fly duing optimization expanded (in a differentiable manner) to the discretized piecewiselinear on which then the Wasserstein Kernel is applied. Currently, desite chosing which scaler to use, botorch input transform are not explicitly exposed to BoFires API.
- We decided at some point that, categoricals have to be encoded for use in surrogates/optimization always as one-hot, as this allows both for dealing with them as normal features treated as the continuous ones without special kernels and using them in a MixedGP botorch `OneHotToNumeric` transform. Of course, also other design choices would be possible here, which have drawbacks in itself.
- Regarding this PR, an 80% solution which would make different implementations for and MixedSingleTaskGPMixedTanimotoGPSurrogate obsolete would be to have the HammingKernel being able to act directly on the onehot encoded values. This would mean that one would move the `OneHotToNumeric` transform into the kernel. Furthermore, also the MultiTaskGP could then benefit from this as we would then have the same flexibility there. Of course we could also lift the restriction of having categoricals always as one-hots. 
- In a next step, one should then make the already eposed scaling transforms (`Normalize`, `Standardize`) configurable by giving the user the explicit option which features these transform are acting on.

In general, botorch input transforms are a great way to build up complex differentiable optimization pipelines, on the other hand it is very challenging to come up with a general framework, parameterizied by feature names that can compose different input transforms on different features including the bookkeeping of indices.

I hope his posts helps ;)

e-dorigatti:
Thank you for the thorough explanation @jduerholt, this clears up a lot of doubts! So the key difference is that the optimizer of the acquisition function is working with the features that result after applying the input preprocessing specs, but before the input transform.

What are the reasons that led you to decide that categoricals have to be one-hotted? Currently because I don't know your motivations it looks a bit silly since you use the input transform to do the one-hot, then the OneHotToNumeric to undo the one-hot, as the Hamming kernel does not need one-hots in the first place. If we now add more code to deal with the fact that this specific transform (and only this one?) changes the input dimensionality, it would seem like over-complicating things even further (but it would not be that complicated in the end). Maybe, given that with this PR you can have kernels focusing on specific (subsets of) features, it is now unnecessary to force categoricals to be one-hot? Moreover, because molecules are in the end a special type of categorical, it would make sense imho to convert the molecular featurizers to transforms rather than preprocessing.

I am now wondering whether it is necessary to expose botorch's input transforms, or it would be okay to put these transformations within the kernels themselves, so that users would specify the features that the kernel is acting on as well as how they should be transformed. I think this would be simpler to implement, and not less powerful than botorch's input transforms unless I am missing something. It also makes sense because, in some way, kernels are the only reason why we need input transforms in the first place, and different kernels could work on different representations of the same feature.

What do you think? In the end, I think that both options can be implemented relatively easily, but perhaps one is more disruptive than the other. I think you have a better perspective and ability to judge :)

jduerholt:
Will provide you a thorough answer tomorrow ;)

jduerholt:
> What are the reasons that led you to decide that categoricals have to be one-hotted? Currently because I don't know your motivations it looks a bit silly since you use the input transform to do the one-hot, then the OneHotToNumeric to undo the one-hot, as the Hamming kernel does not need one-hots in the first place.

The reason was that I wanted to have things too flexible:
- I wanted to be able to combine a SingleTaskGPSurrogate which needs to onehot the categoricals with a MixedSingleTaskGPSurrogate in the same optimization. For this, there is the need that for both of them the same features enter the GP. So I decided for this complicated solution using the OneHotToNumeric transform, as I saw that it was already available in botorch. And then I just said that `CategoricalnputFeature`s always has to be one hot.
- But this kind of constraint is already relaxed for the following features which all inherit from `CategoricalInputFeature`:
    - `CategoricalDescriptorInput`: Here also `CategoricalEncodingEnum.DESCRIPTOR` is allowed.
    - `CategoricalMolecularInput`: Here also the vairous mofeature transforms are allowed. 
    - `TaskInput`: If you want to use it within a `MultiTaskGPSurrogate`, it has to be ordinal encoded.

>  Moreover, because molecules are in the end a special type of categorical, it would make sense imho to convert the molecular featurizers to transforms rather than preprocessing.

You are right, they are special forms of the categorical features. But I would not convert them to botorch input transforms, for the following reasons:
- Botorch input transforms are super nice if you have to change something between the input representation for the optimizer and the representation of the GP. Especially, if you want it in differentiable manner, as we do it in the `PiecewiseLinearGPSurrogate`. This is not really the case for the molecular featureizers. The featurization is non differentiable and we can do it easily upfront.
- The current setup for the transforms is working and I do not see a need to change it.

> I am now wondering whether it is necessary to expose botorch's input transforms, or it would be okay to put these transformations within the kernels themselves, so that users would specify the features that the kernel is acting on as well as how they should be transformed. I think this would be simpler to implement, and not less powerful than botorch's input transforms unless I am missing something. It also makes sense because, in some way, kernels are the only reason why we need input transforms in the first place, and different kernels could work on different representations of the same feature.

Two reasons for keeping it:
1. Practical reason; The transforms are deeply baked into botorch and also well tested. If we want to do it in the kernels, we have to implement it ourselves.
2. Computational reason: the botorch/gpytorch gps store the training data `x*`which is used then used at inference time to calculate the covariance matrix/kernel with respect to the test data `x`: `k(x,x*)`. The trick is that it stores the **transformed** training data, so you do not need to transform the training data at every inference step. All these mechanism are already implemented and well tested especially on the botorch side. For this reason, in general I want to keep the trasforms, but of course there could be exceptions for edge cases for it.

What I would do now:
- Release the restriction that `CategoricalInputs`s has to be always one hot encoded and allow also for ordinal encodings.
- Fix `get_categorical_combinations` in `BotorchStrategy` so that it also works for categoricals encoded as ordinals.
- Implement in `get_categorical_combinations` also the generation of the molecular ones. Currently you can only run optimizations over molecular spaces if you do not have only continuus ones. 
- Drop the options in `BoTorchStrategy` for the `categorical_method`, `descriptor_method` und `discrete_method`. Depending on the surrogate, we had here so far the possibility to use continuous relaxations of them during the optimization, which ofc can lead to wrong optima. So I would drop this support (makes the code much easier), especially as we get the GA by @LukasHebing and @sleweke-bayer for mixed spaces. 
- Change the `ScalerEnum`s into actual classes in which we can define on which features/feature types the corresponding transforms should apply.
- Try to write the `MixedSingleTaskGPSurrogate` in plain bofire as a `SingleTaskGPSurrogate` which the explicit kernels and see that we get the same results.
- Remove all necessary GP implementations and write helper methods that helps you to generate the complex specs.

This should overall, simplify our codebase a lot and will give in general much more flexibility. Of course, there will be also edge cases. What do you think?

Ofc, this should not all be done in this PR by you. So I would suggest, that we try to get this PR (which does not breaks anything) in as fast as possible, agree on the changes from above and then distribute the taks. Fine for you?

e-dorigatti:
Thank you for the clarifications @jduerholt. Following your suggestions I have implemented the Hamming kernel working on one-hot features, which is instantiated by the kernel mapper transparently for the user, and verified that one can get the same results as MixedSingleTaskGP (or even better with a different kernel)

e-dorigatti:
Apologies, I got lost debugging those tests that were failing on GH but not locally, then I noticed other PRs are suffering from the same issue. According to the [print statements](
https://github.com/experimental-design/bofire/pull/476/commits/01a01e13681c840a18359412e791df00b6bd6385) I added it seems like an issue with differentiation in the formulaic package?

```
>       assert np.allclose(B, J)
E       assert False
E        +  where False = <function allclose at 0x7f392a5de930>(array([[0., 1., 0., 0., 0., 2.],\n       [0., 0., 1., 0., 0., 1.],\n       [0., 0., 0., 1., 6., 0.]]), array([[[0, 1, 0, 0, 0, 2],\n        [0, 0, 1, 0, 0, 1],\n        [0, 0, 0, 1, 0, 6]]]))
E        +    where <function allclose at 0x7f392a5de930> = np.allclose

tests/bofire/strategies/doe/test_objective.py:52: AssertionError
----------------------------- Captured stdout call -----------------------------
model 1 + x1 + x2 + x3 + x3 ** 2 + x1:x2
diff x1 0 + 1 + 0 + 0 + 0 + x2
diff x2 0 + 0 + 1 + 0 + 0 + x1
diff x3 0 + 0 + 0 + 1 + 0 + (2*x3)
```

Where `diff x3` should be `0 + 0 + 0 + 1 + (2*x3) + 0` (full output [here](https://github.com/experimental-design/bofire/actions/runs/12414754510/job/34659535002#step:5:182))

bofire/data_models/domain/features.py
```diff
@@ -622,6 +623,21 @@ def _validate_transform_specs(
                     raise ValueError(
                         f"Forbidden transform type for feature with key {key}",
                     )
+            checked_keys.add(key)
```
jduerholt, line 626: From my understanding this whole part here is not necessary, as this is already checked above. Every feature has a method `valid_transform_types` and then it is checked if the assigned transform type is valid based on the registered ones in the feature.

bofire/data_models/kernels/kernel.py
```diff
@@ -1,5 +1,15 @@
...
 
 class Kernel(BaseModel):
     type: str
+
+
+class AggregationKernel(Kernel):
+    pass
+
+
+class ConcreteKernel(Kernel):
```
jduerholt, line 14: Should we give it a more verbose name like `FeatureSpecificKernel`? But ConcreteKernel is also fine for me ;)

bofire/kernels/categorical.py
```diff
@@ -0,0 +1,25 @@
+import torch
+from gpytorch.kernels.kernel import Kernel
+from torch import Tensor
+
+
+class HammingKernelWithOneHots(Kernel):
```
jduerholt, line 6: Can you add a docstring which refers to the botorch implementation and just states that this is the same but based on a one-hot representation?

bofire/kernels/categorical.py
```diff
@@ -0,0 +1,25 @@
...
+    has_lengthscale = True
+
+    def forward(
```
jduerholt, line 9: Are you sure that this actually the same as the botorch implementation. From my understanding you are introducing here one lenghtscale per one-hotted column, whereas as the botorch had one lengthscale per feature. In the end, this appears to me to be the same as a RBF kernel under the assumption that one always has a zero or one, or?

For me the step is missing here where we go from one-hot back to ordinal encoding ..., but I could be wrong.

tests/bofire/kernels/test_categorical.py
```diff
@@ -0,0 +1,18 @@
+import torch
+from botorch.models.kernels.categorical import CategoricalKernel
+from botorch.models.transforms.input import OneHotToNumeric
+
+from bofire.kernels.categorical import HammingKernelWithOneHots
+
+
+def test_hamming_with_one_hot():
```
jduerholt, line 8: I am not 100% sure that this test really is complete enough, as you are only testing here for one categorical feature, furthermore you are testing with `ard_num_dims=1` which is the default, and uses one lenthscale for everything. In this case, by coincidence one gets the same results. 

But I would already expect that the HammingDistanceKernel only works with `ard_num_dims=1` or three. As it treats in the current setup every one-hot expanded column as an independet feature.

jduerholt (commented):
Hi @e-dorigatti,

thank you very much for the additions and sorry for the delay in review.

I have some doubts that the new kernel does exactly the same as the old one. See my comments. If I am correct, I would propose the following next steps, as I am also assuming that it makes too much efforts to come up with a correct version:

- Remove the new kernel
- Merge it as is with the new feature dependend setup as new expert feature
- I will implement then the suggestions that I made last year to lift the restriction that categorical features has to always be one-hot encoded. 

What do you think?

Best,

Johannes

e-dorigatti:
Hi @jduerholt, sorry for the mess, I was probably too eager to start vacation :) I revised the implementation of the kernel to use the one-hot to numeric transformation and added tests for using multiple categoricals at the same time both with and without ARD I also took care of your other comments about renaming, docstrings, redundant checks, etc.

Let me know if you spot any other issues.

tests/bofire/surrogates/test_gps.py
```diff
@@ -293,6 +295,97 @@ def test_SingleTaskGPHyperconfig():
...
+                ),
+                TanimotoKernel(features=["x_mol"]),
+            ]
+        ),
+    )
+
+    gp_mapped = surrogates.map(gp_data)
+    gp_mapped.fit(experiments)
+    pred = gp_mapped.predict(experiments)
+    assert pred.shape == (4, 2)
```
jduerholt, line 385: Should we also test here that the features in active_dims are correct?

bofire/kernels/mapper.py
```diff
@@ -1,22 +1,37 @@
...
 from gpytorch.kernels import Kernel as GpytorchKernel
 
 import bofire.data_models.kernels.api as data_models
 import bofire.priors.api as priors
+from bofire.kernels.categorical import HammingKernelWithOneHots
 from bofire.kernels.fingerprint_kernels.tanimoto_kernel import TanimotoKernel
 from bofire.kernels.shape import WassersteinKernel
 
 
+def _compute_active_dims(
```
jduerholt, line 16: Can you add a test just for this method?

jduerholt (commented):
@e-dorigatti: Looks good to me. Only some minor stuff. Thanks!

After this PR is merged, we can simplify a lot in BoFire ;)

jduerholt (approved):
Perfect @e-dorigatti! Thank you very much. Based on this, we should be able to tidy up a lot in BoFire!

e-dorigatti:
Fantastic, thank you @jduerholt for the support and guidance, happy to have contributed :)

PR #478, opened 2024-12-05
update DoE tutorials

R-M-Lee:
do not direct users to `find_local_max_ipopt` any more. They should just create a strategy and do `ask` unless they are really expert users.


R-M-Lee:
@rosonaeldred I made this pull request and modified the basic notebook. Let's put further changes here then get Johannes and Dominik to review. I won't change the tutorials again here unless you want me to

rosonaeldred:
I'm a bit confused following this ticket, its history, and the two repos -- and that you also want to already merge 3 commits from your branch. Maybe we can take time next week to sort this out.

dlinzner-bcs:
@R-M-Lee @rosonaeldred already done here https://github.com/experimental-design/bofire/pull/448 #488

R-M-Lee:
there were some other additions to the basic tutorial in this branch. It would be good if we get them merged too

rosonaeldred:
looking at the two, dominic's changes include his refactoring so that the formulas are inside the DOptimality criterion and also includes pictures, so i would try to close without merging @R-M-Lee  (also, the merge of jupyter notebooks where one contains images is a special kind of hell, but i opened up a working tree to just visually compare the two and i would just keep dominic's changes)

PR #479, opened 2024-12-05
Categorical deterministic surrogate

jduerholt:
This PR adds categorical deterministic surrogates which can be used to map categorical input features to continuous output values in order to add penalty values for using specific categories in the optimization.

CC: @LukasHebing


LukasHebing (approved):
all clear to me. thanks again Johannes!

PR #480, opened 2024-12-11
Add categoricals for `FractionalFactorialStrategy`

jduerholt:
This PR adds categorical features for the fractional factorial strategy.

It is done by creating a full factorial design for the categoricals and running the (fractional) two level factorial design of the continuous ones for every categorical combination. Doing this makes then also the `FactorialStrategy` obsolete, as one can use the `FractionalFactorialStrategy` now also with only categoricals.

### DIFF
--- a/bofire/data_models/strategies/factorial.py
+++ b/bofire/data_models/strategies/factorial.py
@@ -1,4 +1,7 @@
+import warnings
 from typing import Literal, Type
+
+from pydantic import model_validator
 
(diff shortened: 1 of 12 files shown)


jduerholt:
@bertiqwerty: can you review it? The failing test is due to a new version of sklearn shipped on Monday. We see this failing test currently on several PRs. I try to provide a fix in a seperate one.

Jimbo994 (approved):

PR #482, opened 2024-12-12
Add infrastructure for changing acquisition function optimizers

sleweke-bayer:
The `BotorchStrategy` is expanded to modularize the acquisition function optimizer. Every acquisition function optimizer has to derive from the newly created base class `AcquisitionFunctionOptimizer`. The current Botorch-based optimizer is moved into the separate class `BotorchAcqfOptimizer`.

This PR includes a **breaking change in the data model**: A new data model for acquisition function optimizers is created. The Botorch-optimizer params stored in the `BotorchStrategy` are moved to the new `BotorchAcqfOptimizer`.

Due to the modularization, we can later add different optimizers for the acquisition function (e.g., genetic algorithm).

### DIFF
--- a/bofire/strategies/predictives/acqf_optimizers.py
+++ b/bofire/strategies/predictives/acqf_optimizers.py
@@ -1,0 +1,254 @@
+from abc import ABC, abstractmethod
+from typing import Callable, Dict, List, Optional, Tuple, Type
+
+from botorch.acquisition.acquisition import AcquisitionFunction
+from botorch.optim.optimize import (
+    optimize_acqf,
+    optimize_acqf_list,
+    optimize_acqf_mixed,
+)
+from torch import Tensor
+
+from bofire.data_models.constraints.api import (
+    LinearEqualityConstraint,
+    LinearInequalityConstraint,
+    NChooseKConstraint,
+    ProductConstraint,
+)
+from bofire.data_models.domain.api import Domain
+from bofire.data_models.strategies.predictives import acqf_optimizers as data_models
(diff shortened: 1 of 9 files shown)


bofire/strategies/predictives/acqf_optimizers.py
```diff
@@ -0,0 +1,254 @@
...
+    ):
+        pass
+
+    @classmethod
+    def from_spec(cls, data_model: BaseDataModel) -> "AcquisitionFunctionOptimizer":
+        """Used by the mapper to map from data model to functional strategy."""
+        return cls(data_model=data_model)
+
+    @abstractmethod
+    def optimize(
```
jduerholt, line 47: From my understanding the signature of the `optimize` method should be a bit simpler and maybe also more torch agnostic. From my perspective, we only need the domain, the bounds, and the acqf(s), the rest can be generated on the fly from domain and is also the optimizer specific, or?

I mean that for example a GA wants to have the constraints in a different format then botorch wants them. In principal, also the bounds could be generated on the fly out of the domain, but this would create problems with the local and global bounds, as this is controlled on a different level. Of course, one could also think about putting the LSR logic into the generic optimizer class, then we would only need the acqfs and the domain and maybe also the input processing specs.

What do you think about making it more agnostic here?

jduerholt (commented):
Hi @sleweke-bayer,

thank you very much. I do not had a detailed look but focused more on the architecture and had especially one comment there. What do you think?

Best and a happy new year,

Johannes

jduerholt:
Implemented in https://github.com/experimental-design/bofire/pull/535

PR #484, opened 2024-12-16
Group split kfold

jkeupp:
This pull request introduces the GroupShuffleSplit functionality to the cross_validate method in bofire/surrogates/trainable.py. It adds a new parameter group_split_column to ensure that the splits are made such that the same group is not present in both training and testing sets. 

Additionally, corresponding tests have been added in tests/bofire/surrogates/test_cross_validate.py to validate the new functionality.

For now this only works for stand-alone cross_validation calls of the surrogate. Next step, which is to use it during hyperparameter optimization will be part of a future PR.


jduerholt (changes requested):
Looks good. Thanks, mostly remarks with respect to testing.

jduerholt:
@jkeupp: Do you keep this in mind, so that we can integrate it soon? Best, Johannes

jduerholt (approved):
Looks good to me. Thanks!

Issue #483, opened 2024-12-13
(Python) functions as non-linear constraints in DoE


dlinzner-bcs:
Multiple use case owners have their constraints in terms of complicated functions. Sometimes they have them as (python) functions already available. Since we already use automatic differentiation, we could support the non-symbolic functions as non-linear constraints.

Osburg:
@dlinzner-bcs sounds cool :) How would it be intended to map the variables names to the function inputs then? By requiring the function to have input names corresponding to the variable names and the inputs are given as a dict {"var1": values_for_var1, "var2": values_for_var2...} which we can generate from the model matrix?

Osburg:
Closing this bc #485 was accepted

PR #485, opened 2024-12-18
I optimality criterion and nonlinear constraints defined by python function

Osburg:
This PR adjusts the I optimality criterion suggested in #428 to the new structure of doe from #448 . Also it adds support for nonlinear constraints defined by python functions as suggested by @dlinzner-bcs in #483


Osburg:
@dlinzner-bcs I tried to implement your suggestion from #483, does it roughly look like you imagined?
Cheers
Aaron :)

dlinzner-bcs:
Hi Aaron! Oh very nice thank you! I will check it out.

dlinzner-bcs:
Hi Aaron! That covers the use case i had in mind really well. Thank you :) i added a surrogate use case as a test. So all is left is to merge main again 🙈

Osburg:
At your service 🫡. I guess all these pyright errors are not my fault, right? :D 
@dlinzner-bcs pls feel free to approve and merge if you are okay with the changes.

Cheers
Aaron :)

bofire/strategies/doe/objective.py
```diff
@@ -163,6 +177,185 @@ def get_model_matrix(self, design: pd.DataFrame) -> pd.DataFrame:
...
+            domain=domain,
+            model=model,
+            n_experiments=n_experiments,
+            delta=delta,
+            transform_range=transform_range,
+        )
+
+        # uniformly fill the design space
+        if np.any([isinstance(obj, EqualityConstraint) for obj in domain.constraints]):
+            warnings.warn(
```
dlinzner-bcs, line 222: Would this also work? This seems like less code.
 ```   
    data_model = data_models.DoEStrategy(
        domain=domain,
        optimization_strategy="partially-random",
        criterion=SpaceFillingCriterion(delta=delta),
        ipopt_options={"maxiter": 500, "disp": 0},
    )
    sampler = strategies.DoEStrategy(data_model=data_model)
    result = sampler.ask(n_space_filling_points).values
```

dlinzner-bcs (changes requested):
Thank you @Osburg! Looks great and will be very helpful to us!

bofire/strategies/doe/objective.py
Osburg: Agree that this would be nicer, but I think this leads to circular imports :/

dlinzner-bcs (approved):
This is good to go :) Thank you @Osburg.

PR #490, opened 2025-01-02
factorial strategy missing in api

bertiqwerty:


jduerholt:
Failing tests seems to be a py 3.9 issue with entmoot, or? This is also the case in your other PR. Can you have a look into this? From my point of view we could also remove 3.9 support.

TobyBoyne:
Hi @jduerholt, yes there was a small bug fix in entmoot, and I also pinned he gurobipy version to <12. I would hope that wouldn't introduce more bugs...

From looking at the tests, I can see that many are failing due to using | to combine types instead of Union, which isn't supported in py3.9 - is this what is causing issues?

jduerholt:
FactorialStrategy is deprecated, FractionalFactorialStrategy can do exactly the same.

PR #491, opened 2025-01-02
Make `None` as candidate count possible during benchmarks

bertiqwerty:


jduerholt:
Happy new year @bertiqwerty, tests are failing. Something seems wrong. Can you have a look?

PR #492, opened 2025-01-07
Make candidate_count enforcement less strict

jduerholt:
This PR makes the `candidate_count` enforcement less strict:

- A warning instead of an error is raised if the strategy spits out more candidates than expected. This was also requested by @CompRhys 
- In the `FractionalFactorialStrategy` also values despite `None` for candidate count are allowed, nevertheless it always returns the predefined number of candidates and ignore the value given for `candidate_count`.


bertiqwerty (approved):
Thanks. Agreed.

bertiqwerty:
> Thanks. Agreed.

Actually, I wanted to respond this to the other PR but it also fits here.

PR #493, opened 2025-01-07
Drop support for Python 3.9

jduerholt:
This PR drops the support for Python 3.9. It will go EOL this year (https://devguide.python.org/versions/) and causes especially trouble with Entmoot but also in other occassions. 

cc: @TobyBoyne @LukasHebing

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -19,7 +19,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.12"
+          python-version: "3.10"
 
       - name: Set up uv
         uses: astral-sh/setup-uv@v2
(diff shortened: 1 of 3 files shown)


jduerholt:
We have here some pyright issue, the old version (`1.1.305`) makes problems with py3.10 and throws errors which are not really errors. Switching to newer versions produces a lot of real errors, for which I do not have currently the time to fix. I would propose to merge it as is to get rid of the entmoot problems and handle the pyright stuff seperately. Maybe somebody is also volunteering. What do you think?

bertiqwerty (approved):
Thanks. Agreed.

CompRhys:
> We have here some pyright issue, the old version (`1.1.305`) makes problems with py3.10 and throws errors which are not really errors. Switching to newer versions produces a lot of real errors, for which I do not have currently the time to fix. I would propose to merge it as is to get rid of the entmoot problems and handle the pyright stuff seperately. Maybe somebody is also volunteering. What do you think?

I don't mind grinding through some linting errors, brainless work to do in the evening

jduerholt:
> I don't mind grinding through some linting errors, brainless work to do in the evening

Sounds great! I would recommend to switch directly to the newest version then.

CompRhys:
@jduerholt Will get to this either tonight or over the weekend to try clear the linting issues

PR #494, opened 2025-01-08
Progessbar for Hyperopt

jduerholt:
So far the hyperopt runner in bofire was only showing a progress bar when not using the FractionalFactorialStrategy, unfortunately, this is the default for singletaskgps. This PR adds the progress bar.


Jimbo994 (approved):
Looks good. Nice to have the pbar when running hyperopt!

PR #495, opened 2025-01-09
Make the BoFire Data Models OpenAI compatible

jduerholt:
By playing around with pydanticAI and GPT4o to setup domains for optimization problems described in natural language (which works extremely well), I went into errors for all data model attributes that are tuples as openai does not support tuples in json schema (https://github.com/pydantic/pydantic-ai/issues/372). This PR replaces the tuples by lists.

### DIFF
--- a/bofire/benchmarks/detergent.py
+++ b/bofire/benchmarks/detergent.py
@@ -57,11 +57,11 @@ class Detergent(Benchmark):
 
         self._domain = Domain.from_lists(
             inputs=[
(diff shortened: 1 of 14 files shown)


LukasHebing (commented):
everything looks fine! Only the linting errors hints at tuple usage in bofire/benchmarks/...

LukasHebing:
Also, pydantic seems to convert bounds, given as tuples to list. So that won`t break anything in application

jduerholt:
@LukasHebing: I updated the linting.

PR #504, opened 2025-01-28
Feature/add desirability function in range

LukasHebing:
This PR adds the desirability function `InRangeDesirability`

This objective is $1.$ inside a specified range, and $0.$ outside this range, defined by the `bounds` of the objective data-model.
 
This is used in our app by the users to handle quality constraints.

The data-model is added in `bofire/data_models/objectives/desirabilities.py`

An example is given within the desirability notebook `tutorials/advanced_examples/desirability_objectives.ipynb`


LukasHebing:
We have a linting error pointing out that `get_objective_callable` is too complex, which I totally agree on. It seems, that we now reached the limit of allowed if-statements, as the number of objectives increased. 

I would like to restructure the definition of the torch callables (preferrably put them into the data-models?), but that would be a bit much for this PR.

@jduerholt: Can we somehow increase the level of allowed complexity for the linting?

jduerholt (approved):
Looks good to me. The complexity threshold can be adjusted here. As quick fix you can increase it to 19 (https://github.com/experimental-design/bofire/blob/ac38bcc6922661d567f44990ea9b1ef07c68d86a/pyproject.toml#L121)

We should refactor this part in a seperate PR, but I would not include it directly in the data models because than we would have torch also as part of the minimal installation, so you would need torch for running the data models.

Just increase the thresshold and merge it in. Thanks!

LukasHebing:
> Looks good to me. The complexity threshold can be adjusted here. As quick fix you can increase it to 19 (
> 
> https://github.com/experimental-design/bofire/blob/ac38bcc6922661d567f44990ea9b1ef07c68d86a/pyproject.toml#L121
> 
> )
> We should refactor this part in a seperate PR, but I would not include it directly in the data models because than we would have torch also as part of the minimal installation, so you would need torch for running the data models.
> 
> Just increase the thresshold and merge it in. Thanks!

I increased the complexity as suggested. Thanks @jduerholt :)

PR #497, opened 2025-01-10
Adding Desirability functions as objectives

LukasHebing:
- added a file with data-models 3 different Desirability functions: Increasing, Decreasing, and Peak Desirability
- Some options for the function design (bounds, clipping, concave/convex form
- Added torch executables into torch_tools.py
- Tests for data-models and objectives added
- A demonstration notebook, on how the functions look is added as example

PR #505, opened 2025-01-29
Fix pyright issues

CompRhys:
Fixes the pyright issues from #497 and #504.


.pre-commit-config.yaml
```diff
@@ -25,3 +25,10 @@ repos:
     hooks:
       - id: nbstripout
         args: [--drop-empty-cells, --keep-output]
+
```
jduerholt, line 28: Can we remove this comment block?

jduerholt (changes requested):
@CompRhys: Looks really good to me. I overlooked that here already a lot was solved. Thanks!

I only let one comment, but this should be super easy to fix!

.pre-commit-config.yaml
```diff
@@ -25,3 +25,10 @@ repos:
...
+# repos:
+#   - repo: https://github.com/RobertCraigie/pyright-python
+#     rev: v1.1.305
+#     hooks:
+#     - id: pyright  # TODO: work out how to install the package in the
+#                    # pre-commit venv
```
CompRhys, line 34:
```suggestion
```

.pre-commit-config.yaml
CompRhys: Ideally if you stick to pyright the best way to be consistent would be to try would out how to put it into the pre-commit. However I didn't manage to do that here and so will remove the dead commented block.

jduerholt (approved):
Looks good to me. Thank you very much!

CompRhys:
thanks, I can't recreate the remaining issues locally I really thought I'd squashed them all with the extra push last night

PR #510, opened 2025-02-04
Randomization for FractionalFactorials

jduerholt:
This PR adds an option for randomizing the order of candidates proposed by the `FractionaFactorialStrategy`.

### DIFF
--- a/bofire/data_models/strategies/fractional_factorial.py
+++ b/bofire/data_models/strategies/fractional_factorial.py
@@ -46,6 +46,10 @@ class FractionalFactorialStrategy(Strategy):
         int,
         Field(description="Number of reducing factors", ge=0),
     ] = 0
+    randomize_runorder: bool = Field(
+        default=False,
+        description="If true, the run order is randomized, else it is deterministic.",
+    )
 
     @classmethod
     def is_constraint_implemented(cls, my_type: Type[Constraint]) -> bool:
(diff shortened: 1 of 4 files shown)


jduerholt:
@dlinzner-bcs: also the https://github.com/experimental-design/bofire/blob/main/tutorials/doe/basic_examples.ipynb is failing. Any idea, why?

Osburg:
Hey Johannes :)

To me it looks like it is running without an error, am I wrong?

Screenshot from 2025-02-05 14-25-15

jduerholt:
@Osburg: you are right. The error stems from another one, and I solved it in a seperate one.  If you are already looking at this PR, can you also just review ;)

jduerholt:
@Osburg: here it is failing again; https://github.com/experimental-design/bofire/actions/runs/13155774927/job/36712478341?pr=511

Osburg (approved):
think it looks good :)

Osburg:
The notebook fails since ipopt sometimes has problems to fulfill the more complicated nonlinear constraints (like the cones) to the (fixed) tolerance level in the validation of candidates. I would suggest to set "raise_validation_error=False" for the cone examples (since the constraints are only mildly violated, i.e. above the threshold but still on the cone surface when visually inspecting the design). An alternative would be to increase the number of max iterations, but then the notebook will run longer, which is not worth it i think, the jobs are running for a very long time now already... What do you think?

PR #511, opened 2025-02-05
Make BoFire compatible to BoTorch 0.13.0

jduerholt:
This PR makes BoFire compatible to the most recent BoTorch release. We should also file a new release in the near future. But, before, I want to have the PR on the randomization in fractional factorials in (@dlinzner-bcs) and maybe also need to have a look at the DoEStrategy.

### DIFF
--- a/bofire/utils/torch_tools.py
+++ b/bofire/utils/torch_tools.py
@@ -144,7 +144,9 @@ def get_interpoint_constraints(
     return constraints
 
 
-def get_nchoosek_constraints(domain: Domain) -> List[Callable[[Tensor], float]]:
+def get_nchoosek_constraints(
+    domain: Domain,
+) -> List[Tuple[Callable[[Tensor], float], bool]]:
     """Transforms NChooseK constraints into a list of non-linear inequality constraint callables
     that can be parsed by pydantic. For this purpose the NChooseK constraint is continuously
     relaxed by countig the number of zeros in a candidate by a sum of narrow gaussians centered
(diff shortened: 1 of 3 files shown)


bofire/utils/torch_tools.py
```diff
@@ -182,24 +184,32 @@ def min_constraint(indices: Tensor, num_features: int, min_count: int):
...
+                    True,
+                )
             )
     return constraints
 
 
-def get_product_constraints(domain: Domain) -> List[Callable[[Tensor], float]]:
+def get_product_constraints(
+    domain: Domain,
+) -> List[Tuple[Callable[[Tensor], float], bool]]:
```
bertiqwerty, line 212: The longish type `Tuple[Callable[[Tensor], float], bool` is somehow difficult to understand. Wouldn't it make sense to use a type like
```python
class Constraint(BaseModel):
    evaluate: Callable[[Tensor], float]
    is_intrapoint: bool
```
and then use `list[Constraint]` as return type?

bertiqwerty (approved):
Thanks Johannes. I just left one minor comment.

bofire/utils/torch_tools.py
jduerholt: In principle yes, but I would like to keep it as is, becuase this is exactly the format which is required by the botorch api. Else, we have an additional conversion step. If ok for you, I would merge the current version of the PR.

bertiqwerty: Could you add the role of the bool to the docstring?

PR #512, opened 2025-02-05
Fix failing doe notebooks

Osburg:
The doe basic examples often fail as pointed out by @jduerholt in #510. This PR avoids raising an error by setting  raise_validation_error=False in `ask()` for the cone examples to avoid raising errors due to slightly violated constraints.


jduerholt (approved):
Thank you very much. Looks good to me!

PR #520, opened 2025-02-14
Polynomial feature interaction kernel

e-dorigatti:
Hi, now that we have kernels that work only on subset of features, it would be nice to also combine these kernels to include interactions amongst them. This can be done already but it is a bit tedious for higher-order interactions and quite inefficient since the kernels would be re-computed multiple times with the same inputs.

Essentially one could now replace

```
k1 = RBFKernel(features=["x1", "x2"])
k2 = MaternKernel(features=["x2", "x3"])
k3 = HammingDistanceKernel(features=["x4", "x5"])

k = AdditiveKernel(kernels=[
    ScaleKernel(base_kernel=k1),
    ScaleKernel(base_kernel=k2),
    ScaleKernel(base_kernel=k3),
    ScaleKernel(base_kernel=MultiplicativeKernel(kernels=[k1, k2])),
    ScaleKernel(base_kernel=MultiplicativeKernel(kernels=[k1, k3])),
    ScaleKernel(base_kernel=MultiplicativeKernel(kernels=[k2, k3])),
])
```

with

```
k = PolynomialFeatureInteractionKernel(
    kernels=[k1, k2, k3],
    max_degree=2,
)
```

### DIFF
--- a/bofire/kernels/aggregation.py
+++ b/bofire/kernels/aggregation.py
@@ -1,4 +1,3 @@
-import functools
 import itertools
 from typing import Any, Optional
 
@@ -24,23 +23,19 @@ class PolynomialFeatureInteractionKernel(gpytorch.kernels.Kernel):
         self.kernels = kernels
         self.max_degree = max_degree
         self.indices = [
-            kk
-            for n in range(1, self.max_degree + 1)
-            for kk in (
+            list(
                 itertools.combinations_with_replacement(range(len(kernels)), n)
                 if include_self_interactions
                 else itertools.combinations(range(len(kernels)), n)
             )
+            for n in range(1, self.max_degree + 1)
         ]
 
-        self.coefs = torch.nn.Parameter(
-            torch.sqrt(torch.ones(len(self.indices)) / len(self.indices))
-        )
-
+        n = sum(len(idx) for idx in self.indices)
         lengthscale = (
-            torch.zeros(*self.batch_shape, len(self.indices))
+            torch.zeros(*self.batch_shape, n)
             if len(self.batch_shape)
-            else torch.zeros(len(self.indices))
+            else torch.zeros(n)
         )
         self.register_parameter(
             name="raw_lengthscale", parameter=torch.nn.Parameter(lengthscale)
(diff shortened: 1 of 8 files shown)


bofire/kernels/aggregation.py
```diff
@@ -0,0 +1,103 @@
...
+        outputscale_prior: Optional[gpytorch.priors.Prior] = None,
+        outputscale_constraint: Optional[gpytorch.constraints.Interval] = None,
+        *args: Any,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(*args, **kwargs)
+
+        self.kernels = kernels
+        self.max_degree = max_degree
+        self.indices = [
```
jduerholt, line 25: Can you add some comments here on what you are doing?

jduerholt: And what this indices are used for?

bofire/kernels/aggregation.py
```diff
@@ -0,0 +1,103 @@
...
+        self.register_parameter(
+            name="raw_outputscale", parameter=torch.nn.Parameter(outputscale)
+        )
+        if outputscale_prior is not None:
+            if not isinstance(outputscale_prior, gpytorch.priors.Prior):
+                raise TypeError(
+                    "Expected gpytorch.priors.Prior but got "
+                    + type(outputscale_prior).__name__
+                )
+            self.register_prior(
```
jduerholt, line 49: For what is this prior used?

jduerholt: Should be documented in the datamodel where the kernel should be explained.

jduerholt (changes requested):
@e-dorigatti: Thanks for this PR. Looks overall good. I let some comments, especially regarding documentation, but this should be integrated quickly. Could be good to rebase the feature branch with main so that the we do not see here changes which were already merged into main. 

Thanks for the PR!

e-dorigatti:
Thanks for the review @jduerholt, I've added the documentation and comments, and integrated the changes from main.

jduerholt (approved):
Thanks! Looks good to me!

Issue #521, opened 2025-02-14
cyipotp mandatory asof 0.0.17?


jkeupp:
hi guys!

i'm currently struggling with 0.0.17, in particular since  it seems that for any functional (bofire[all] or similar) installations, it seems that cyipopt, consumed by the doe subpackages has become mandartory:

```
ModuleNotFoundError: No module named 'cyipopt'
Traceback:
[THINGS OMITTED]
File "/app/.venv/lib/python3.13/site-packages/bofire/strategies/api.py", line 1, in <module>
    from bofire.strategies.doe_strategy import DoEStrategy
File "/app/.venv/lib/python3.13/site-packages/bofire/strategies/doe_strategy.py", line 12, in <module>
    from bofire.strategies.doe.branch_and_bound import (
    ...<2 lines>...
    )
File "/app/.venv/lib/python3.13/site-packages/bofire/strategies/doe/branch_and_bound.py", line 16, in <module>
    from bofire.strategies.doe.design import find_local_max_ipopt
File "/app/.venv/lib/python3.13/site-packages/bofire/strategies/doe/design.py", line 18, in <module>
    from bofire.strategies.doe.objective import get_objective_function
File "/app/.venv/lib/python3.13/site-packages/bofire/strategies/doe/objective.py", line 10, in <module>
    from cyipopt import minimize_ipopt  # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This is a problem, as currently we wish to keep deployments simple and not install this thing. particularly not via conda, which seems to be the only viable option asof right now (sry, i've not looked into it very deeply... )

any suggestions?

bertiqwerty:
I stumbled across the same thing. Thanks for bringing this up @jkeupp and thanks for fixing @jduerholt.

Btw., you can install it without conda on a Ubuntu Docker container by apt-get-installing the necessary packages for IPOPT (https://coin-or.github.io/Ipopt/INSTALL.html) and and pip-installing cyipopt.

jduerholt:
We will also implement a fallback to scipy.minimize as discussed here: https://github.com/experimental-design/bofire/pull/525

jkeupp:
thanks a lot for the quick fix, @jduerholt ! Sorry for not reviewing myself the PR, i was basically offline between last friday and today.

Resolved, closing this!

PR #522, opened 2025-02-14
Fix that BoFire is not working without cyipopt

jduerholt:
@jkeupp reported that the current BoFire version on pypi is not working when cyipopt is not installed. The reason is that cyipopt is used in the instantiation of I-optimal criterion and that we do not guard the import there. This PR fixes it and adds a test where we just install the optimization package of BoFire to prevent it in the future.

After having this in, we should also make a release.

cc: @Osburg @dlinzner-bcs @bertiqwerty

### DIFF
--- a/bofire/strategies/doe/objective.py
+++ b/bofire/strategies/doe/objective.py
@@ -205,12 +205,10 @@ class IOptimality(ModelBasedObjective):
 
         try:
             from cyipopt import minimize_ipopt  # type: ignore
-        except ImportError as e:
-            warnings.warn(e.msg)
-            warnings.warn(
-                "please run `conda install -c conda-forge cyipopt` for this functionality.",
+        except ImportError:
+            raise ImportError(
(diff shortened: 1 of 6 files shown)


jduerholt:
Closes https://github.com/experimental-design/bofire/issues/521

bofire/strategies/doe/objective.py
```diff
@@ -204,6 +203,15 @@ def __init__(
                 If None is provided, the default options (maxiter = 500) are used.
         """
 
+        try:
+            from cyipopt import minimize_ipopt  # type: ignore
+        except ImportError as e:
+            warnings.warn(e.msg)
+            warnings.warn(
+                "please run `conda install -c conda-forge cyipopt` for this functionality.",
```
bertiqwerty, line 211: I would not put out warnings for missing libraries Because I think more often then not it is intended that library is missing. And if someone really wants to use something they have not installed, they will get an error anyway.

bertiqwerty (approved):

PR #523, opened 2025-02-14
`get_by_key` using regex

jduerholt:
This PR adds the possibility to use regex for getting a feature by a key.

### DIFF
--- a/bofire/data_models/domain/features.py
+++ b/bofire/data_models/domain/features.py
@@ -114,18 +114,18 @@ class _BaseFeatures(BaseModel, Generic[F]):
             return Outputs(features=cast(Tuple[AnyOutput, ...], new_feature_seq))
         return Features(features=new_feature_seq)
 
-    def get_by_key(self, key: str, regex: bool = False) -> F:
+    def get_by_key(self, key: str, use_regex: bool = False) -> F:
         """Get a feature by its key.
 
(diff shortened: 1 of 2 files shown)


bertiqwerty:
Thanks Johannes. Let me make a proposal. I will push it to this branch.

bofire/data_models/domain/features.py
```diff
@@ -113,17 +114,33 @@ def is_outfeats(feats):
...
 
         Returns:
             Feature: Feature of interest
 
         """
-        return {f.key: f for f in self.features}[key]
+        try:
+            if use_regex:
+                return next(
+                    f for f in self.features if re.match(key, f.key) is not None
```
bertiqwerty, line 138: Are you sure you want `re.match` instead of `re.search`? `re.match` matches just the beginning of the string. `re.search` matches the regex somewhere in the string.

jduerholt: You are right, it should `re.search`. It was an copilot error ;)

jduerholt:
> Thanks Johannes. Let me make a proposal. I will push it to this branch.

@bertiqwerty:  So, you proceed working on this PR and then we merge it as soons as it ready?

bertiqwerty:
I just changed to search. From my point of view this is ready now.

Issue #503, opened 2025-01-28
Is pyright actually helping?


CompRhys:
I said that I would go through a fix some of the linting issues a few weeks back but there are so many pointless pyright complaints that result in `# type: ignore` needing to be added everywhere that it makes me question whether there's a benefit to using pyright if the API is so complex that adhering to the type pyright rules becomes an additional burden even for correct code.

jduerholt:
@bertiqwerty , what do you think?

CompRhys:
My main issue with pyright is that there's no good way I can see to get it to run in pre-commit due to needing to install the package in the pre-commit env which I can't find a good solution for. As such it's something that will continually break because people "know" that their code works and so for development velocity often it makes sense to merge and fix later.

There are also weird redefinition choices that I needed to add in #505

bertiqwerty:
Just for my understanding, do you think Pyright as a tool is not worth it or would you want to ditch static type checking alltogether?

CompRhys:
Unless there's not a good way to get the tool to work locally in the same way as in CI my thoughts are it's an annoyance rather than a benefit, cf just using a ruff rule to say you have to include type hints (without a checker). In the linked PR i fixed all the pyright issues I get locally pinning the pyright version to 1.1.305 rather than using the newer versions but it still seemingly divergences due to py3.10 in CI and me using py3.12 locally. There are too many ways to use this particular tool in a manner than even fixing all issues locally it will break the linting CI.

jduerholt:
@bertiqwerty: what do you think about using such a ruff rule to enforce that the user actually uses types?

LukasHebing:
I am also a fan of correct typing (@bertiqwerty). But with `pyright`, I had several problems:
- I couldn`t reproduce the behavior locally
- A lot of fails were actually wrong (code worked as it should, with the given types)
- For some reformulations (same line of code), it took me hours to find the "pyright-correct" version. Very annoying.

However, in some places, it was helpful, e.g. to think about optional, None inputs that I forgot to test.

bertiqwerty:
>     * For some reformulations (same line of code), it took me hours to find the "pyright-correct" version. Very annoying.

Can you show some examples?

LukasHebing:
> > For some reformulations (same line of code), it took me hours to find the "pyright-correct" version. Very annoying.
> 
> Can you show some examples?

```python
if not any(isinstance(feat.objective, obj) for obj in allowed_objectives):
    continue
```
[...]

PR #524, opened 2025-02-14
Make pyright a seperate test job

jduerholt:
As discussed in issue https://github.com/experimental-design/bofire/issues/503, we have currently problems with pyright that leads to fails in the linting pipeline, so we currently ignore the linting pipeline when looking at PRs, which is not good, because one does not see on the first glance if there are maybe also problems with ruff etc. 

This PR makes pyright a seperate pipeline job so that one sees if there are other issues.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -32,11 +32,6 @@ jobs:
       - name: Run pre-commit
         run: pre-commit run --all-files --show-diff-on-failure
 
-      - name: Pyright
-        uses: jakebailey/pyright-action@v2
-        with:
-          version: 1.1.305  # is there any reason not to use the latest?
-
   types:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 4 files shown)


jduerholt:
Maybe also @Jimbo994 can review, as this fixes especially linting error in the gp slice method ;)

Best,

Johannes

Jimbo994 (approved):
Noted and approved. Thanks for fixing.

.github/workflows/lint.yaml
```diff
@@ -32,6 +32,25 @@ jobs:
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+
+      - name: Set up uv
+        uses: astral-sh/setup-uv@v2
+        with:
+          enable-cache: true
+
+      - name: Install dependencies
```
bertiqwerty, line 51: I think you don't need to install the pre-commit hooks.

bertiqwerty (commented):
Thanks Johannes.

PR #529, opened 2025-02-25
Strategy TidyUp

jduerholt:
This PR changes the following things:
- Removes the deprecated qehvi and qnehvi strategies, they are replaced by the MoboStrategy
- It refactors the tests so that they should not anymore fail randomly in the strategies section


jduerholt:
@bertiqwerty: it is ready for review now ;)

bertiqwerty (approved):
Thanks, Johannes.

jduerholt:
@Osburg here are also agains some examples where the tests fail due to problems in doe ...

PR #532, opened 2025-03-03
Blocking for Fractional Factorial Designs

jduerholt:
This PR adds blocking for fractional factorial designs, currently it is not supported for fractional factorial designs. But this can be implemented in a straight foward manner. Only thing necessary is to add the blocking generators to `/bofire/bofire/utils/default_fracfac_generators.py` and update the getters. I will take care for this in a seperate PR when this one is merged.


bofire/strategies/fractional_factorial.py
```diff
@@ -34,27 +50,79 @@ def _get_continuous_design(self) -> pd.DataFrame:
...
         design = pd.DataFrame(fracfact(gen=gen), columns=continuous_inputs.get_keys())
+
+        if self.n_blocks > 1 and self.n_repetitions % self.n_blocks != 0:
+            block_generator = get_block_generator(
+                n_factors=len(continuous_inputs),
+                n_blocks=self.n_blocks,
+                n_repetitions=self.n_repetitions,
+                n_generators=self.n_generators,
+            )
+            design["block"] = apply_block_generator(
```
R-M-Lee, line 61: is it a problem if there are other features called "block" or if the blocking feature key is "block"?

tests/bofire/data_models/specs/strategies.py
```diff
@@ -593,6 +594,93 @@
...
+                        categories=["a", "b", "c"],
+                        allowed=[True, True, True],
+                    ),
+                ],
+            ),
+        ),
+        "block_feature_key": "d",
+    },
+    error=ValueError,
+    message="Number of blocks 3 is not possible with 1 repetitions.",
```
R-M-Lee, line 680: I would have naively thought that the default would set the number of repetitions to be equal to the number of allowed categories in the blocking factor (then it would work I think). Is there a reason why we do not do this?

In general, I find the interplay between n_repetitions and n_blocks a bit confusing.

R-M-Lee (approved):
Hi @jduerholt , nice work. I left a couple of comments

tests/bofire/data_models/specs/strategies.py
jduerholt: I programmed it in the Minitab way, there you say how many repetitions you want to perform and then it shows you how many blocks are possible for the chosen design given the number of repetitions, in our case the user has to specify the number of repetitions and blocks and we check if it is possible. 

Note that you can also have blocking without any repetitions.

R-M-Lee: ok, fair enough. I'm not familiar with minitab so I can just accept that

bofire/strategies/fractional_factorial.py
jduerholt: Solved it ;)

Issue #463, opened 2024-11-19
bump entmoot version


R-M-Lee:
there are some bugfixes in a newer version of entmoot

R-M-Lee:
closing because this was solved in #536

PR #536, opened 2025-03-04
Bump Entmoot and LightGBM

### LINKED ISSUE #530: upgrade lightgbm
Our vulnerability scan complains about lightgbm at version `4.0.0`, and requires at least `4.6.0`. Is there a reason, this is fixed at 4.0 in the `pyproject.toml`?

```
"lightgbm==4.0.0",
```

@jduerholt

TobyBoyne:
Bumps the version of Entmoot (which provided bug fixes), and LightGBM (which fixed a vulnerability). Also provides a slight shift in the strategy so that `test_strategy_can_be_loaded_from_data_model` works even without Entmoot installed, which avoids an ugly `pytest.importorskip("entmoot")`.

Closes #530, #463.
Passes all Enting tests with a Gurobi license and `--runslow` on my machine - I don't remember if all of these run on the CI pipeline on GitHub.


jduerholt (approved):
Thanks. Looks good to me, any comments from you @R-M-Lee ?

setup.py
```diff
@@ -22,7 +22,7 @@
         "cvxpy[CLARABEL]",
         sklearn_dependency,
     ],
-    "entmoot": ["entmoot>=2.0", "lightgbm==4.0.0", "pyomo==6.7.1", "gurobipy"],
+    "entmoot": ["entmoot~=2.0.4", "lightgbm~=4.6.0", "pyomo==6.7.1", "gurobipy"],
```
R-M-Lee, line 25: sorry, I forgot what we discussed before. There was something with the pyomo version, wasn't there? We need to keep it at 6.7.1?

R-M-Lee (approved):

pyproject.toml
```diff
@@ -70,7 +70,7 @@ all = [
     "cvxpy[CLARABEL]",
     "scikit-learn>=1.0.0,<1.6",
     "entmoot>=2.0",
-    "lightgbm==4.0.0",
+    "lightgbm>=4.0.0",
```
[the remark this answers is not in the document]
R-M-Lee: and in setup.py it's slightly different again. What is the logic?

R-M-Lee:
thank you very much @TobyBoyne . Just left a couple of questions

pyproject.toml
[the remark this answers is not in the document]
TobyBoyne: Completely slipped by, not quite sure how. Sorry!

TobyBoyne:
@R-M-Lee I've made the requirements more consistent. However, I did also wonder why we include the dependencies in BoFire at all - would it be better to just install entmoot, and let entmoot handle its own dependencies?

jduerholt:
> @R-M-Lee I've made the requirements more consistent. However, I did also wonder why we include the dependencies in BoFire at all - would it be better to just install entmoot, and let entmoot handle its own dependencies?

I would also favor this.

TobyBoyne:
> I would also favor this.

Great! I think this would be a lot cleaner. I think we initially included the versions of these packages in BoFire because Entmoot didn't handle its dependencies correctly, but this should be fixed soon (https://github.com/cog-imperial/entmoot/pull/54, just waiting on confirmation that it doesn't break anything).

TobyBoyne:
Following the merging of https://github.com/cog-imperial/entmoot/pull/54, I've now removed all the entmoot sub-dependencies from BoFire, and bumped the entmoot version to `entmoot~=2.0.5`. This PR should now be ready to merge, @jduerholt.

R-M-Lee:
Looks good to me 😺

jduerholt:
Why do we have a `setup.py` and a ``pyproject.toml`, we got rid of the `setup.py` a few weeks ago, it looks that it was merged by error again in. @TobyBoyne: can you remove it? Also the tests are failing, and I do not know why.

pyproject.toml
```diff
@@ -69,10 +69,7 @@ all = [
...
     "cvxpy[CLARABEL]",
     "scikit-learn>=1.0.0,<1.6",
-    "entmoot>=2.0",
-    "lightgbm==4.0.0",
-    "pyomo==6.7.1",
-    "gurobipy",
+    "entmoot~=2.0.5",
     "xgboost>=1.7.5",
     "rdkit>=2023.3.2",
     "mordred",
```
jduerholt, line 75: Somehow, it is installing a super old mordred version, can you say here "mordred>=1.2.0"?

TobyBoyne: Ah - mordred specifies that they need numpy==1.*, I assume an older version of mordred did not pin the numpy version which is why it is doing that. Entmoot specified numpy~=2, which caused issues. I'll unspecify the version of numpy needed by entmoot now.

TobyBoyne:
Thanks for the reviews, @jduerholt! I have removed `setup.py`, and removed the `numpy~=2.0.0` dependency in entmoot. The tests are still failing, but now it's the tests that are failing in main, so I don't think this is my fault!

jduerholt:
Thanks @TobyBoyne! Have a nice weekend!

PR #537, opened 2025-03-05
Bump image for doc pipelines

jduerholt:
In the doc pipelines, we are still using Ubuntu 20.04, which is not anymore supported by github, here we change to `ubuntu-latest`.


jduerholt:
@bertiqwerty: can you have a look? ;)

bertiqwerty (approved):
Thanks, Johannes.

PR #553, opened 2025-03-17
Working Stepsize Implementation

jduerholt:
## Motivation

Currently, one can provide a `stepsize` for a continuous input feature but it is not exploited. Here we make the first steps towards a working stepsize implementation. In case that a continuous input is not part of a constraint, it is rounded after candidate generation to the closest allowed value. For predictive strategies, the predictions are then calculated for the rounded value. 

It is a draft as it will be incorporated after the refactoring of the acqf optimization is included. 

cc: @LukasHebing 

Y

## Test Plan

Unit tests.

### DIFF
--- a/bofire/data_models/features/continuous.py
+++ b/bofire/data_models/features/continuous.py
@@ -3,7 +3,7 @@ from typing import Annotated, ClassVar, List, Literal, Optional, Tuple
 
 import numpy as np
 import pandas as pd
-from pydantic import Field, model_validator
+from pydantic import Field, PositiveFloat, model_validator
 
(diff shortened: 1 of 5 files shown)


LukasHebing:
Do I understand that correctly, that the rounding step in the optimization is done in `postprocess_candidates`, and after the optimization? So that normally, the optimizer would use the same range?

That would be a good solution. For the GA, I would also include the possibility to optimize step- factors in the same way as "discrete" variables. 

Shall we make this behavior configurable for all optimizers? E.g. having a `treat_continuous_inputs_with_step_as_discrete` option in the abstract optimizer class?

PR #555, opened 2025-03-19
Add chemistry focus to documentation

TobyBoyne:
## Motivation

Where the BoFire paper has a focus on chemistry applications, the documentation is far more general. This PR updates the docs to provide more of a sense of the application in chemistry, while still providing a general overview.

Yes

## Test Plan

N/A

### DIFF
--- a/docs/userguide_surrogates.md
+++ b/docs/userguide_surrogates.md
@@ -60,6 +60,7 @@ HammingDistanceKernel|Similarity is defined by the [Hamming distance](https://en
 Translational invariance means that the similarity between two input points is not affected by shifting both points by the same amount but only determined by their distance. Example: with a translationally invariant kernel, the values 10 and 20 are equally similar to each other as the values 20 and 30, while with a polynomial kernel the latter pair has potentially higher similarity. Polynomial kernels are often suitable for high-dimensional inputs while for low-dimensional inputs an RBF or Matérn kernel is recommended.
 
 **Note:**
+
 - SingleTaskGPSurrogate with PolynomialKernel is equivalent to PolynomialSurrogate.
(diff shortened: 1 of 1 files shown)


jduerholt:
@R-M-Lee, are you reviewing this?

docs/fingerprint_bayesopt.ipynb
```diff
@@ -0,0 +1 @@
+../tutorials/benchmarks/009-Bayesian_optimization_over_molecules.ipynb
```
R-M-Lee, line 1: this and the other two ipynb should be symlinks. Right now it looks like these files are breaking the pipeline

R-M-Lee (changes requested):
Thanks for doing this Toby! I think we need to change
- docs/fingerprint_bayesopt.ipynb
- docs/multifidelity_bo.ipynb
- docs/reaction_optimization.ipynb
To be symbolic links so that the pipeline does not break. Can you do this? You can see in the web interface that the ipynb files in https://github.com/experimental-design/bofire/tree/main/docs are blue because they are recognized as links. Displaying the contents in github also shows a note. Right now I think that they are just normal files so the pipeline tries to interpret them as true notebooks

R-M-Lee:
Otherwise it looks nice @TobyBoyne . Thanks again. The symlink thing can be annoying depending on what OS you are developing in

TobyBoyne:
@R-M-Lee thanks for spotting that - I was working in Windows and hadn't realised they were symlinks. Should hopefully be working now.

R-M-Lee:
yep, it looks good now. The jobs in question have passed. I will merge when the whole pipeline is finished

R-M-Lee (approved):

PR #561, opened 2025-03-27
Feature/add ga optimizer

LukasHebing:
PR for adding a genetic algorithm (pymoo) to BoFire.

`GeneticAlgorithm` is the second option, besides `BotorchOptimizer`, but works similar:

- All `q` experiments are optimized simultaneously. That means, that the dimension of the problem is `d*q`. A sequential mode could be implemented later.

**UPDATE: Variable Types**
- `CategoricalInput` variables, which are treated as one-hot-encoded columns by the model and the acquisition functions, are turned into categorical variables for the GA optimization. In the objective function, these categorical variables are transformed to one-hot-encoded tensors. The object `BofireDomainMixedVars` handles this conversion.
- `CategoricalDescriptorInput` is also transformed in to a categorical pymoo variable, but transformed into the descriptor space
- `DiscreteInput` will be converted to an pymoo Integer.

All transformations are handled in the helper class `BofireDomainMixedVars` 

**Constraints**
The GA cannot handle equality constraints well. Constraints are therefor handled differently:

- Constraints of the type `LinearEqualityConstraint`, `LinearInequalityConstraint`, and `NChooseKConstraint` are handled in a "repair-function". This repair function is used by the GA to map all individuals from the population $x$ to the feasible space $x'$. In this case, I implemented a repair-function for an arbitrary mixture of linear equality and inequality constraints with a quadratic programming approach:

$$
\min_{x'} \left( ||x-x' ||_2^2 \right)
$$

s.t.

$$
A \cdot x' = b
$$

$$
G \cdot x' <= h
$$

$$
lb <= x' <= ub
$$ 

The `NChooseKConstraint` is also handled in the reapir function: For each experiment in the population, the smallest factors are set to 0, if the *max_features* constraint is violated, and the upper bound of the largest feactors is set to an offset (defaults to $1e-3$), if the *min_features* constraint is violated.

The repair functions are handled in the class `LinearProjection`.

- Other supported constraints are: `ProductInequalityConstraint` and `NonlinearInequalityConstraint`. `ProductInequalityConstraint` are evaluated by the torch-callable, provided by the `get_nonlinear_constraints` function. `NonlinearInequalityConstraint` are evaluated from the experiments data-frame, by the constraints `__call__` method.

These are handled by the optimizer.

`NonlinearEqualityConstraints` are not supported.


LukasHebing:
This PR is in Draft stage, I am still working on some details.

R-M-Lee:
Hi @LukasHebing , looks good. This is purely for acqf optimization, not for pareto front estimation, correct? That seems like it would only be a small extension. ie, we have a `MultiobjectiveStrategy` that has been trained and we just want to compute the Pareto front based on the surrogate(s). This could be an extension to `get_pareto_front` in `bofire/utils/multiobjective.py` so that it uses `pymoo` as an alternative to just working with the data that the user provides. What do you think?

And concerning the N-choose-k constraints; perhaps we can help to put them into the repair operator. Have done that before

LukasHebing:
> so that it uses `pymoo` as an alternative to just working with the data that the user provides. What do you think? Maybe in separate PR

I'll put the NChooseK constraint on the repair function. Also, I am working on making the optimizer see the categorical values  as true categorical values. Pymoo now supports this.

LukasHebing:
I could change the GA optimizer to use mixed variable types, so that categorical parameters are treated as such by the optimizer. They are converted into one-hot-encoded Tensors in the objective function, for the acquisition function

jduerholt:
Hi @LukasHebing,

thanks! Can you rebase/merge with main so that we see here what is new and do not need to re-review the stuff that is already merged?

Best,

Johannes

LukasHebing:
@jduerholt, @R-M-Lee: I am right now working on the NChooseK constraint repair function. I got some questions:

The constraint is about factors not equal to zero, which is checked with:
```python
sums = (np.abs(experiments[cols]) > tol).sum(axis=1)
```
Does it make sense, to have an `NChooseKConstraint` for a factor, which can be smaller than zero? Should we add this in the validator?
Or is there a case, where factors can be smaller, and greater zero, but the number of non-zero entries matters?

R-M-Lee:
> @jduerholt, @R-M-Lee: I am right now working on the NChooseK constraint repair function. I got some questions:
> 
> The constraint is about factors not equal to zero, which is checked with:
> 
> ```python
> sums = (np.abs(experiments[cols]) > tol).sum(axis=1)
> ```
> 
> Does it make sense, to have an `NChooseKConstraint` for a factor, which can be smaller than zero? Should we add this in the validator? Or is there a case, where factors can be smaller, and greater zero, but the number of non-zero entries matters?

I think that N-choose-k constraints only make sense for inputs that are $\geq0$ or, in some cases $>0$. As far as I'm aware, the N-choose-k constraints are only used on inputs that are amounts of material (eg., in formulations) and the goal is to avoid a product or process with too many ingredients. So I would not bother accounting for cases where factors can be $<0$, but I would be ok with it if someone else has a good reason

LukasHebing:
@jduerholt: Another question on the `stepsize` of 

> > @jduerholt, @R-M-Lee: I am right now working on the NChooseK constraint repair function. I got some questions:
> > The constraint is about factors not equal to zero, which is checked with:
> > ```python
> > sums = (np.abs(experiments[cols]) > tol).sum(axis=1)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Does it make sense, to have an `NChooseKConstraint` for a factor, which can be smaller than zero? Should we add this in the validator? Or is there a case, where factors can be smaller, and greater zero, but the number of non-zero entries matters?
> 
> I think that N-choose-k constraints only make sense for inputs that are ≥ 0 or, in some cases > 0 . As far as I'm aware, the N-choose-k constraints are only used on inputs that are amounts of material (eg., in formulations) and the goal is to avoid a product or process with too many ingredients. So I would not bother accounting for cases where factors can be < 0 , but I would be ok with it if someone else has a good reason

I would add a validator for this in the domain, ok?

jduerholt:
Hi @LukasHebing, 

thanks for merging it. 

I will perform a review on Monday morning.

Fine for you?

Best,

Johannes

LukasHebing:
> Hi @LukasHebing,
> 
> thanks for merging it.
> 
> I will perform a review on Monday morning.
> 
> Fine for you?
> 
> Best,
> 
> Johannes

Great!
There was one minor issue in the test. I hope, that's resolved now

R-M-Lee:
> @jduerholt: I realized, that optimization tests occasionally fail, because of the very strict way of checking n-choose-k constraints, mostly for the botorch optimizer. The narrow gaussian function might lead to constraint values of ~0.5 - 1.0, even if the largest feature is ~1e-12.
> 
> I was thinking of checking the constraint differently? Or with a configurable tolerance in the narrow gaussian?

which narrow gaussian do you mean? I see that we are using a tolerance of 1e-5 in the `test_optimizer.py` tests that are currently failing. (`assert (constr(proposals).values <= 1e-5).all()`)

Regardless, I have a question: We are using the default cvxpy solver, Clarabel, correct? Could the constraint violation perhaps be improved by taking a different one? Another option would be to do one final 'post-repair' on the candidates with a tighter tolerance

jduerholt:
@R-M-Lee and @LukasHebing:

I would in general not put too much efforts into the NChooseKs here. Our handling of them (also using the botorch optimizer) is in general not very good but there is a way how to handle them correctly and this is by using SEBO (https://arxiv.org/abs/2203.01900). I only need to find time to implement it in BoFire (it is already in botorch). So, I would not put too much time here into this solutions, but put it into SEBO. And SEBO would work for both ACQF optimizers.

LukasHebing:
> > @jduerholt: I realized, that optimization tests occasionally fail, because of the very strict way of checking n-choose-k constraints, mostly for the botorch optimizer. The narrow gaussian function might lead to constraint values of ~0.5 - 1.0, even if the largest feature is ~1e-12.
> > I was thinking of checking the constraint differently? Or with a configurable tolerance in the narrow gaussian?
> 
> which narrow gaussian do you mean? I see that we are using a tolerance of 1e-5 in the `test_optimizer.py` tests that are currently failing. (`assert (constr(proposals).values <= 1e-5).all()`)
> 
> Regardless, I have a question: We are using the default cvxpy solver, Clarabel, correct? Could the constraint violation perhaps be improved by taking a different one? Another option would be to do one final 'post-repair' on the candidates with a tighter tolerance

The failing test comes from the original botorch optimizer. The issue is only, at which factor value, the constraint is seen as fulfilled, and this is not $1e-5$, but something close to $1e-12$, due to the form of calculation.

My proposal would be, to make this a bit more loose, because we are too close to most optimizers tolerance.

bofire/strategies/predictives/acqf_optimization_utils/GA_utils.py
```diff
@@ -0,0 +1,709 @@
...
+            assert all(
+                c in (NonlinearInequalityConstraint,)
+                for c in nonlinear_pandas_constraints
+            )
+        self.nonlinear_pandas_constraints = domain.constraints.get(
+            includes=nonlinear_pandas_constraints
+        )
+
+        super().__init__(
+            n_obj=len(acqfs),
```
jduerholt, line 293: I think, we should not support here more than one ACQF, or? Originally, it is used only for QPAREGO, and there it is used in a sequential way and not a MOO way, or what was your intention here?

jduerholt (changes requested):
Wow, this is a massive PR! @LukasHebing: Thank you!

I now went over the implementation in a first pass, and marked some stuff. I especially have the question, if it would be helpful to implement a new transform for the `DiscreteInput` that transforms it to integers, but I think this would not help. Maybe you could comment ;)

Another question: If we would have in the future not anymore the restriction that categoricals has to be one-hotted, but could be also integer-encoded in the GPs directly. How complx would it be to adapt?

I will go a second time over it, after you have adressed the points, then I should also have a better understanding!

Thank you very much.

Best,

Johannes

LukasHebing:
> @R-M-Lee and @LukasHebing:
> 
> I would in general not put too much efforts into the NChooseKs here. Our handling of them (also using the botorch optimizer) is in general not very good but there is a way how to handle them correctly and this is by using SEBO (https://arxiv.org/abs/2203.01900). I only need to find time to implement it in BoFire (it is already in botorch). So, I would not put too much time here into this solutions, but put it into SEBO. And SEBO would work for both ACQF optimizers.

@R-M-Lee: I realized, that I used the wrong function for the constraint validation. In the function `is_fulfilled`, the constraint is tested, as you described. The more complex check with the narrow gaussian is in the `__call__` method.

LukasHebing:
@jduerholt; 
Yes, the GA makes the encoding more complex, because now we have three encoding domains: 1) The original domain, 2) The model and the acquisition functions, 3) the optimizer.
Here, the optimizer needs to check encoding 1+2, to generate the encoding 3. So that's really complex now.
But I don't know, if we can avoid that.
I try to clean up the GA encoding finder a bit.

jduerholt:
Hi @LukasHebing,

if not possible, than it is like it is ;) 

Best,

Johannes

LukasHebing:
@jduerholt: Tests work now. You can take another look now.

jduerholt:
Thanks. I will have a look tomorrow!

bofire/strategies/predictives/acqf_optimization_utils/GA_utils.py
```diff
@@ -0,0 +1,720 @@
...
+        """
+        vars = {}
+        for qi in range(self.q):
+            vars = {**vars, **{f"{key}_q{qi}": var for key, var in self.vars.items()}}
+        return vars
+
+    def _pymoo_specific_transform(self, experiments: pd.DataFrame) -> pd.DataFrame:
+        """handles the non-domain encodings (e.g. pymoo-type 'Integer' to discrete"""
+        for key, conversion in self.pymoo_conversion.items():
+            experiments[key] = [conversion[i] for i in experiments[key]]
```
jduerholt, line 149: Can you use pandas.DataFrame.replace here to get rid of the the second loop (comprehension)? https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.replace.html

bofire/strategies/predictives/acqf_optimization_utils/GA_utils.py
```diff
@@ -0,0 +1,720 @@
...
+    def transform_to_experiments(self, X: List[dict]) -> List[pd.DataFrame]:
+        """Transform to a list of "experiments" dataframes for each q-point"""
+        experiments = pd.DataFrame.from_records(X)
+        q_column = np.array([int(x.split("_q")[-1]) for x in list(experiments)])
+
+        experiments_out = []
+
+        for qi in range(self.q):
+            experiments_qi = experiments.iloc[:, q_column == qi]
+            experiments_qi.columns = [x.split("_q")[0] for x in experiments_qi.columns]
```
jduerholt, line 176: What happens if we have "_q" in a feature name?

bofire/strategies/predictives/acqf_optimization_utils/GA_utils.py
jduerholt, line 1: Does it make sense to have an own directory here?

bofire/utils/torch_tools.py
```diff
@@ -254,7 +256,23 @@ def get_nonlinear_constraints(
         as input and return a float value representing the constraint evaluation.
 
     """
-    return get_nchoosek_constraints(domain) + get_product_constraints(domain)
+    if includes is None:
```
jduerholt, line 259:
```suggestion
    includes = includes or [NChooseKConstraint, ProductInequalityConstraint]
```

jduerholt: This makes it a one-liner and you get rid of the if-else and just move the assert out of the else.

tests/bofire/strategies/test_acqf_optimizer_utils.py
```diff
@@ -0,0 +1,4 @@
+from bofire.benchmarks.api import Ackley
+
+
+benchmark = Ackley(categorical=False, dim=2, lower=-30.0, upper=30.0)
```
jduerholt, line 4: Here is something missing, or? You are just initializing the benchmark but not testing anything, or?

jduerholt (changes requested):
Hi @LukasHebing,

looks quite ready. I let a few comments! Thansk for the efforts. Are the tests also already ready for review? 

I was not sure, since there was a file in tests with almost nothing in, and then there are also still this `tmp_` files, so I was not sure, and have not yet properly looked at the tests and also not at the tmp files. When you say, I should look at them, I will directly do it ;)

Best,

Johannes

jduerholt:
@LukasHebing 

Driving home, I had the following idea to simplify the conversion stuff:

The `DiscreteInput` can be modelled as a `CategoricalDescriptiorInput` with a one-dimensional descriptor which are the actual discrete values. By doing it in this way, we could fully rely on the already implemented transforms and would save a lot of complexity.

This could be achieved in two ways:
1. Do not support the `DiscreteInput`, and require a `CategoricalDiscreteInput`
2. Transform the domain in the optimizer under the hood so that every DiscreteInput is a CategoricalDiscreteInput, generate the candidates then back to DiscreteInput. 

The second option could have some pitfalls. The first would work out of the box and you could have a pre and post processor outside of the actual strategy. Using BoFire tools, one could do it with the StepwiseStrategy and using only one step for example. I could generate an example for this tmr (wrapping the conversion in a stepwiseStrategy), if you are interested ;)

What do you think?

Other option is of course to keep it as is, I do not want to block you ;)

Best, 

Johannes

bofire/strategies/predictives/acqf_optimization_utils/GA_utils.py
jduerholt: See the comment also below, it could be that we could completly drop it.

bofire/strategies/predictives/acqf_optimization_utils/GA_utils.py
jduerholt: What do you think if we drop the support for the combined optimization and do it always in a sequential way? That would solve this issue and would reduce complexity. 

But we can also do it in a seperate PR after it is merged.

LukasHebing: I tested the sequential optimization, and the results are different. We have this approach in AIDET now as well. It occasionally leads to nearly identical sets of experiments.
The renaming can be made more clear (e.g. creating a lookup-table in the object)

bofire/utils/torch_tools.py
LukasHebing: nice trick :)

bofire/strategies/predictives/acqf_optimization_utils/GA_utils.py
LukasHebing: I thought predictives is already quite full, and the methods are only helper methods, but I don't mind if we put the file in predictives.

tests/bofire/strategies/test_acqf_optimizer_utils.py
LukasHebing: that was an artifact, I deleted the file.

LukasHebing:
> @LukasHebing
> 
> Driving home, I had the following idea to simplify the conversion stuff:
> 
> The `DiscreteInput` can be modelled as a `CategoricalDescriptiorInput` with a one-dimensional descriptor which are the actual discrete values. By doing it in this way, we could fully rely on the already implemented transforms and would save a lot of complexity.
> 
> This could be achieved in two ways:
> 
> 1. Do not support the `DiscreteInput`, and require a `CategoricalDiscreteInput`
> 2. Transform the domain in the optimizer under the hood so that every DiscreteInput is a CategoricalDiscreteInput, generate the candidates then back to DiscreteInput.
> 
> The second option could have some pitfalls. The first would work out of the box and you could have a pre and post processor outside of the actual strategy. Using BoFire tools, one could do it with the StepwiseStrategy and using only one step for example. I could generate an example for this tmr (wrapping the conversion in a stepwiseStrategy), if you are interested ;)
> 
> What do you think?
> 
> Other option is of course to keep it as is, I do not want to block you ;)
> 
> Best,
> 
> Johannes

That could make the transformation here a bit easier, but most of the class would remain, as we still need to transform the categorical values to the encodings, and sort the q-repeats, etc. 
The main reason why I took "Integer" as the pymoo type for Discrete is, that the crossover, mutation etc. will be closer to the numerical meaning (kind of like an ordinal-encoding of the sorted discrete values, e.g. "1" is close to "0", but far from "5"). We don't have that for the categorical Input.

So I would like to keep the Integer and Categorical encodings.

jduerholt:
Ok, then let us keep it! Tell me when I should rereview ;=

LukasHebing:
> Ok, then let us keep it! Tell me when I should rereview ;=

@jduerholt: You can check now :). Apart from pyright, everthing seems to pass the tests

jduerholt:
@LukasHebing: one other question regarding testing: So far we are only testing the whole optimization cycle. But it would also make sense to test the methods in `Ga_utils.py` on their own, or? One method after the other. What do you think?

bofire/strategies/predictives/acqf_optimization_utils/GA_utils.py
jduerholt: Let us just call it `utils.py` in the parent directory.

LukasHebing: ok, let's do that :)

LukasHebing:
> @LukasHebing: one other question regarding testing: So far we are only testing the whole optimization cycle. But it would also make sense to test the methods in `Ga_utils.py` on their own, or? One method after the other. What do you think?

That would be more clean. I try to make to make copilot write some tests...

jduerholt:
> That would be more clean. I try to make to make copilot write some tests...

If it is too much effort, just tell it, I do not want to block here ;)

jduerholt:
> Had a look into it. That's mostly not easy to do, since the input format to the functions are highly pymoo dependent, and preparation without the optimization is quite some work. If is's ok to skip that, I would be relieved...

Ok, then we skip it for now and discuss at the hackathon in two weeks how to better test it. But then you can already use it! Do you fix the typing stuff? Then we could merge it ;)

jduerholt:
If it is too complex to fix, you can also use ignore statements.

LukasHebing:
> > Had a look into it. That's mostly not easy to do, since the input format to the functions are highly pymoo dependent, and preparation without the optimization is quite some work. If is's ok to skip that, I would be relieved...
> 
> Ok, then we skip it for now and discuss at the hackathon in two weeks how to better test it. But then you can already use it! Do you fix the typing stuff? Then we could merge it ;)

I fixed the pyright issues. I had to add some ignore statements, because I didn't want to change type annotations all around the rest of the repo. Pyright is a bit picky e.g. with `Optional[...]` statements...

Looking forward, thanks for the help :)

jduerholt (approved):
Looks good to me. Thanks! Let us merge it for now, so that it can be used. We can discuss potential improvements in two weeks in person!

PR #565, opened 2025-04-01
naive random sampling pfront estimation

### LINKED ISSUE #563: [FEATURE REQUEST]: example showing a quick way of estimating Pareto fronts
### Motivation

I think an example showing explicitly how to get Pareto fronts from predictive MO strategies would be nice. It's almost what, e.g., tutorials/benchmarks/011-ZDT1.ipynb is doing, but not quite.

- Define/choose benchmark
- Get initial dataset
- Fit surrogates
- generate lots of random samples in the input space
- Get surrogate predictions for the random samples
- Retain only the non-dominated samples
- Visualize

The problem can be more elegantly solved with, eg, pymoo. That will be the subject of a separate PR

_No response_

### Pull Request

None

R-M-Lee:
## Motivation

I think it would be nice to have a very simple example showing how to estimate the Pareto front from a strategy containing trained surrogate models (that you don't want to change any more)

yes

## Test Plan

N/A


jduerholt:
@R-M-Lee : Looks good to me! Thanks. Would it make sense to also add pymoo to it? But this can also be done later. 

What would be your typical usecase for it?

You have run a bayesopt campaing and have distributed points around the front and now you want to estimate the front more accurately, in order to navigate it?

jduerholt:
@R-M-Lee: Any updates on this PR?

jduerholt:
@bertiqwerty @R-M-Lee @KappatC @LukasHebing 

Hi all,

we discussed pareto fron sampling on the hackathon on wednesday and I find it a super nice feature, especially, when we would combine it with more sophisticated sampling using NSGA-2 and pymoo (since pymoo is now also a dependency). For implementing this we could reuse a lot of the code from @LukasHebing that he wrote for optimizing an acqf with pymoo, since from setting up pymoo out of the domain, it makes no diference if we optimize an acqf or surrogates. This could all stay ;)

I also included @KappatC since in the VBLL paper, they showed an interesting acqf for multiobjective opt based on thompson sampling and NSGAII. Also there we could then reuse the code. But first would be to have the VBLL surrogate. 

Best,

Johannes

LukasHebing:
@R-M-Lee: Did you check out the GA (mostly in bofire/strategies/predictives/utils.py)? 
My idea is, that we could make most of the methods a bit more abstract. We only need to replace the acqf-function part with something more general

Jimbo994:
Hi guys, I was recently estimating pareto fronts with the GA from @LukasHebing, then stumbled upon this, and figured it was nice to revisit and wrap it up. 

Made a new notebook in our quarto documentation with random sampling, and the GA. Is that what you had in mind @R-M-Lee?

If yes, perhaps we can merge it.

R-M-Lee:
Hi @Jimbo994, yes this is pretty much what I had in mind! Thanks for bringing this back to life.

I added prediction uncertainty to the final plot because I thought that would be interesting too.

Otherwise I have a couple of minor points:

- the comparison of sobol vs ga is not really fair because they get different numbers of evaluations (8192 vs 20k). But I guess we don't see the notebook as a rigorous benchmark, but rather simply a demonstration of how to do these things
- would you like to add a few words about how you chose the reference point for the hv calculation?
- what do you think about the comment I added to explain why `optimization_direction="min"` is needed when calling the ga (even though the benchmark domain already defines the objectives with directions)?

Otherwise looks good to merge! Thanks again

Jimbo994:
Thanks @R-M-Lee! I’ve now given Sobol and the GA the same evaluation budget. I also added a note that this is meant as a showcase rather than a rigorous comparison, and that the GA is expected to scale better to higher-dimensional input spaces.
I added an explanation for the hypervolume reference point as well. For BNH, the worst attainable values over the input domain are (136, 50), so I used (138, 52) and kept it fixed for all fronts.
Your explanation of optimization_direction="min" makes sense to me, so I kept it. The uncertainty bars are a nice addition too—thanks for adding those!
I think this should be good to merge now.

Jimbo994:
@jduerholt can you review and merge?

jduerholt (approved):
Looks good to me. Thx!

PR #566, opened 2025-04-01
Use Seed Sequences for random number generation

TobyBoyne:
## Motivation

Currently, when generating random numbers, the `RandomStrategy` uses a `np.random.Generator` object to sample a random seed. The best practice for generating sequences of seeds is [`np.random.SeedSequence`](https://numpy.org/doc/2.2/reference/random/bit_generators/generated/numpy.random.SeedSequence.html). This PR now uses this in `Strategy`.

Yes

## Test Plan

All tests pass for `tests/bofire/strategies/test_random.py`.

### DIFF
--- a/bofire/strategies/strategy.py
+++ b/bofire/strategies/strategy.py
@@ -33,7 +33,7 @@ class Strategy(ABC):
             if data_model.seed is not None
             else np.random.default_rng().integers(1000)
         )
-        self.rng = np.random.default_rng(self.seed)
+        self.seed_seq = np.random.SeedSequence(self.seed)
         self._experiments = None
         self._candidates = None
 
@@ -45,7 +45,8 @@ class Strategy(ABC):
             int: random seed.
 
         """
-        return int(self.rng.integers(1, 100000))
+        (spawned_seed_seq,) = self.seed_seq.spawn(1)
+        return spawned_seed_seq.generate_state(1).item()
 
     @classmethod
     def from_spec(cls, data_model: DataModel) -> "Strategy":
--- a/bofire/strategies/strategy.py
+++ b/bofire/strategies/strategy.py
@@ -24,16 +24,12 @@ class Strategy(ABC):
     ):
         self.domain = data_model.domain
         # if data_model.seed is None (no explicit seed provided by the user),
-        # we draw a random seed from the default random number generator
-        # This is done to ensure reproducibility of the strategy,
+        # we use the randomly generated entropy from the seed sequence.
+        # This is done to ensure reproducibility of the strategy:
         # even if the user does not provide a seed one can extract the used seed
         # from the strategy object via `strategy.seed`.
-        self.seed = (
-            data_model.seed
-            if data_model.seed is not None
-            else np.random.default_rng().integers(1000)
-        )
-        self.seed_seq = np.random.SeedSequence(self.seed)
+        self.seed_seq = np.random.SeedSequence(data_model.seed)
+        self.seed = self.seed_seq.entropy
         self._experiments = None
         self._candidates = None
 
--- a/bofire/strategies/strategy.py
+++ b/bofire/strategies/strategy.py
@@ -24,12 +24,16 @@ class Strategy(ABC):
     ):
         self.domain = data_model.domain
         # if data_model.seed is None (no explicit seed provided by the user),
-        # we use the randomly generated entropy from the seed sequence.
+        # we use a randomly generated seed from the seed sequence.
         # This is done to ensure reproducibility of the strategy:
         # even if the user does not provide a seed one can extract the used seed
         # from the strategy object via `strategy.seed`.
-        self.seed_seq = np.random.SeedSequence(data_model.seed)
-        self.seed = self.seed_seq.entropy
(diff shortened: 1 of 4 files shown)


TobyBoyne:
Hi @jduerholt! This PR introduces Seed Sequences for RNG. 

While looking at the `RandomStrategy`, I couldn't work out the benefit of making `_sample_from_polytope` a static method? It leads to repetition, in passing `domain`, `fallback_sampling_method`, `n_burnin`, and `n_thinning` whenever the method is called. Plus, it is private, so it shouldn't really need to be called outside of the class. I see that it is used in testing, but those tests should be rewritten to test public methods eg. `ask`, no?

Thanks!

TobyBoyne:
Something worth noting: by changing the approach of random number generation, any code that relies on the previous behaviour will produce different random results after this PR for the same seed. Do you think this will be a problem?

bofire/strategies/strategy.py
```diff
@@ -24,12 +24,12 @@ def __init__(
...
-        # we draw a random seed from the default random number generator
-        # This is done to ensure reproducibility of the strategy,
+        # we use the randomly generated entropy from the seed sequence.
+        # This is done to ensure reproducibility of the strategy:
         # even if the user does not provide a seed one can extract the used seed
         # from the strategy object via `strategy.seed`.
-        self.seed = data_model.seed or np.random.default_rng().integers(1000)
-        self.rng = np.random.default_rng(self.seed)
+        self.seed_seq = np.random.SeedSequence(data_model.seed)
+        self.seed = self.seed_seq.entropy
```
jduerholt, line 32: Let us make `self.seed` to a property to prevent a user to overwrite it:

```
@property
def seed(self) --> int:
    return self.seed_seq.entropy
```

jduerholt (changes requested):
Looks good to me. Thanks @TobyBoyne!

I only let one comment. 

@bertiqwerty: also fine for you?

jduerholt:
> Something worth noting: by changing the approach of random number generation, any code that relies on the previous behaviour will produce different random results after this PR for the same seed. Do you think this will be a problem?

Should not be a problem, at least I do not see one.

> While looking at the RandomStrategy, I couldn't work out the benefit of making _sample_from_polytope a static method? It leads to repetition, in passing domain, fallback_sampling_method, n_burnin, and n_thinning whenever the method is called. Plus, it is private, so it shouldn't really need to be called outside of the class. I see that it is used in testing, but those tests should be rewritten to test public methods eg. ask, no?

The reason is, that the method is called by the strategy not using `self.domain` as argument but `domain` (see around line 153 in the random strategy). `domain` is a modification of `self.domain` to handle NChooseKs ... One could of course make it a method of the class and only pass the updated `domain` in. But, if we want to do it, I would do in in a seperate PR.

jduerholt:
There is also still a problem with the tests, I assume that has something todo with passing the seeds to botorch.

bertiqwerty:
> Looks good to me. Thanks @TobyBoyne!
> 
> I only let one comment.
> 
> @bertiqwerty: also fine for you?

Besides the failing tests, looks good to me. Thanks, @TobyBoyne!

> While looking at the RandomStrategy, I couldn't work out the benefit of making _sample_from_polytope a static method? It leads to repetition, in passing domain, fallback_sampling_method, n_burnin, and n_thinning whenever the method is called. Plus, it is private, so it shouldn't really need to be called outside of the class. I see that it is used in testing, but those tests should be rewritten to test public methods eg. ask, no?

I get your point. I see arguments for both ways:
- pro static: It does not need access to all of `self` but just to a subset. This way we have better encapsulation and for the reader it is clear what are the parameters actually used by this function. The number of parameters with 6 is not small but in my opinion still ok. This way it is actually a pure function without side effects.
- contra static: convenience to not re-type parameters. 

I personally prefer it the way it is. One could make it a free function outside the class.

jduerholt:
I have no idea, why the 3.10 test is failing ... @bertiqwerty @TobyBoyne any idea from you?

TobyBoyne:
No clue @jduerholt... are you able to rerun the pipeline?

jduerholt:
> No clue @jduerholt... are you able to rerun the pipeline?

I did it already several times. Seems to be a conda/numpy issue. @bertiqwerty what do you think?

bertiqwerty:
> > No clue @jduerholt... are you able to rerun the pipeline?
> 
> I did it already several times. Seems to be a conda/numpy issue. @bertiqwerty what do you think?

Yeah. I also don't get it. I will have a closer look until the end of the week.

jduerholt:
@TobyBoyne: can you try to raise here the python version number to 3.11 and push again so we see if it is an issue with 3.10 ...:https://github.com/experimental-design/bofire/blob/d75d1996fdb2b71338377587968b08112a6c04bf/.github/workflows/test.yaml#L71

TobyBoyne:
Not ideal! That test turned out to be tuned very sensitively - there is a threshold that was often exceeded by random chance. I've increased the threshold a bit to make the test more reasonable.

jduerholt (approved):
Looks good to me. Thx @TobyBoyne.

PR #567, opened 2025-04-04
bounds can be tuple

bertiqwerty:
## Motivation

Static type checkers will not complain if bounds are defined as tuples with two elements. With this change, lists and tuples can be used without complaints of static type checkers.

Y

## Test Plan

Should not have an impact on the runtime behaviour. Just a change of a type annotation.

### DIFF
--- a/bofire/data_models/types.py
+++ b/bofire/data_models/types.py
@@ -1,5 +1,5 @@
 from collections.abc import Sequence
-from typing import Annotated, Dict, List, Union
+from typing import Annotated, Dict, List, Tuple, Union
 
 from pydantic import AfterValidator, Field, PositiveInt
 
@@ -92,7 +92,7 @@ Descriptors = Annotated[
 ]
 
 Bounds = Annotated[
-    List[float],
+    Union[List[float], Tuple[float, float]],
     Field(min_length=2, max_length=2),
     AfterValidator(validate_monotonically_increasing),
 ]


jduerholt:
Hi Behrang,

thanks, it was already a tuple, then I rewrote it to a list with two mandatory items, since openAI cannot handle tuples in pydantic. We need to test here, if it can handle this behavior, or have you already done this?

Best,

Johannes

bertiqwerty:
No. I haven't tested this. I just realized that we use a lot of tuples across the code base. Especially in tests. What about the following: We merge this PR and if you want to use it with OpenAI you have to use lists in your domains.

jduerholt:
It is not about using lists, it is about the json schema (out of pydantic) which is supported or not supported by openAI. How urgent is it? If not too urgent, I could test it this week ;)

bertiqwerty:
Ah. Ok. It is about the schema not the json itself. Can you point me to the OpenAI documentation that explains this? I want to understand this better.

The schema is indeed different and does depend on the type hint (I did `ContinuousInput.model_json_schema()` to look at it). However, if you want to keep the list for OpenAI, you should also change all usages to lists then, I think. My VS Code shows me a lot of errors where tuples are used. Further, the docs, e.g., in `ContinuousInput` still say tuple.

jduerholt (approved):
Let us merge it in. If we run again into problems with openAI, we will find a solution.

jduerholt:
Hi @bertiqwerty, I found time to check it. It does not work. OpenAI is not happy with our json schemas as they are not accepting tuples. And nothing were tuples are optional used. Fine for you, if we roll it back?

Issue #515, opened 2025-02-07
Naming of data model and strategy classes


Osburg:
Hey :) 
Strategy and datamodel classes are often named the same (e.g. there exists a data model and a corresponding strategy with the exact same name `DoEStrategy`). When working with the code, this can sometimes be confusing for me, since I do not see at the first glance which class is meant. Is there a reason why the class names do not reflect whether they are a datamodel or a strategy (such as a suffix 'DataModel' for all data models)? 

Cheers
Aaron :)

bertiqwerty:
Hey Aaron. 

I completely get your point. However, I do not have a too strong on opinion on whether we should change the names of the data model classes. My workaround for the problem you described is usually something like
```python
from bofire.data_models.strategies.api import SoboStrategy as SoboDm
from bofire.strategies.api import SoboStrategy

def f(domain: Domain):
    sobo_dm = SoboDm(domain=domain)
    
    # note that I do not use the mapper-function when I know the strategy since that would
    # make my IDE loose type information for code completion
    sobo_strat = SoboStrategy(sobo_dm) 
    ...
```
Depending on the number of imports I might also do
```python
import bofire.data_models.strategies.api as dm
import bofire.strategies.api as strats

def f(domain: Domain):
    random_dm = dm.RandomStrategy(domain=domain)
    random_strat = strats.RandomStrategy(random_dm) 
    sobo_dm = dm.SoboStrategy(domain=domain)
    sobo_strat = strats.SoboStrategy(sobo_dm) 
    ...
```

Our [Getting Started](https://experimental-design.github.io/bofire/getting_started/#random-strategy) suggests the following where this mixup also does not appear

```python


import bofire.strategies.api as strategies
from bofire.data_models.strategies.api import RandomStrategy


strategy_data_model = RandomStrategy(domain=domain)

random_strategy = strategies.map(strategy_data_model)
random_candidates = random_strategy.ask(2)

random_candidates
```

However, I don't like using `.map` if I know the type due to code completion as mentioned before.

Osburg:
Hey @bertiqwerty,

thanks for the tips, that's helpful! :) Yes, I find it okay too. Just wanted to check with this issue if there are many other users/contributors who also find this confusing sometimes. If people are fine with the current status, it seems to be good as it is. @jduerholt Do you have an opinion on this? 

Best
Aaron :)

jduerholt:
[...]

PR #572, opened 2025-04-08
Create a `make` method in `Strategy`s

TobyBoyne:
## Motivation

Following from discussion on #515, this is a proposal for bridging the gap between `Strategy`s and their `DataModel`s. I have left this as a draft, since we should still discuss exactly how to implement this.

The motivation is for the following code to "just work":

```python
from bofire.strategies.predictives.sobo import SoboStrategy
strategy = SoboStrategy.make(domain=domain)
```

This can be pretty easily implemented as described in the linked issue. The main difficulty is in getting correct type hinting. We want the signature of `make` to match the signature of `SoboStrategyDataModel.__init__`. This PR includes a rough outline of how this can be achieved, but it would require the `copy_datamodel_kwargs` decorator to be used in every Strategy. It would be nicer if `make` just used `Strategy.data_model_cls`, and then subclasses only need to set this class attribute. However, the type hinting requires knowing which class is to be used.

Some relevant resources:
https://stackoverflow.com/questions/71968447/python-typing-copy-kwargs-from-one-function-to-another
https://github.com/python/mypy/issues/12673

Yes

## Test Plan

(Write your test plan here. If you changed any code, please provide us with clear instructions on how you verified that your changes work.)


bertiqwerty:
Regarding the parameters and type hints, I have changed the initial idea to use a `copy_kwargs`-function. Instead, I have duplicated the arguments into the `make`-functions. We have test to make sure that they coincide with the parameters of the data models. I have added doc-strings to the `make`-functions so that all parameters including a description can be found in one place and one does not need to search across the data model's inheritance hierarchy.

#### Test Plan

In unit tests,
- we check that the signatures of the data model constructors and make functions conincide including type annotations.
- we make sure that the data models after creation coinicde.

tests/bofire/strategies/test_make_strategy.py
```diff
@@ -0,0 +1,118 @@
...
+
+def remove_optional(anno):
+    origin = get_origin(anno)
+    if origin == typing.Union or origin is types.UnionType:
+        return sorted((a for a in anno.__args__ if a is not type(None)), key=hash)
+    else:
+        return [anno]
+
+
+def test_make():
```
jduerholt, line 35: I am wondering, could we use here for testing all **valid** strategy specs that are defined in https://github.com/experimental-design/bofire/blob/main/tests/bofire/data_models/specs/strategies.py ? As they are used in https://github.com/experimental-design/bofire/tree/main/tests/bofire/data_models/serialization and https://github.com/experimental-design/bofire/blob/main/tests/bofire/data_models/test_invalid.py. This would have the advantage, to not have to manually add new strategies also here, instead all strategy specs, that are registered for the serialization tests, would also be tested here. This would keep testing in the future easier.

bofire/strategies/strategy.py
```diff
@@ -18,10 +18,13 @@ class Strategy(ABC):
...
     """
 
+    data_model_cls = DataModel
+
     def __init__(
         self,
         data_model: DataModel,
     ):
+        self.data_model = data_model
         self.domain = data_model.domain
```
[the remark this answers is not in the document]
jduerholt: But maybe it is also too much ... What do you think?

jduerholt (commented):
Hi @bertiqwerty,

looks really good to me. This is a tremendous quality of life feature! I let some minor comments regarding documentation, code sanity and testing. Apart from this, this looks really great and I love it. It is a massive quality of life feature!

What do you think of doing the same for surrogates in a seperate PR?

Best,

Johannes

bofire/strategies/strategy.py
```diff
@@ -18,10 +18,13 @@ class Strategy(ABC):
 
     """
 
+    data_model_cls = DataModel
```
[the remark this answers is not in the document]
bertiqwerty: Good point. I think we don't need it in the base class. I will double check.

bofire/strategies/strategy.py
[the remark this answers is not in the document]
bertiqwerty: Agreed. We could also make `self.data_model` a property. Do we change `self.domain` or `self.data_model` after instantiation?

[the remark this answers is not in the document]
bertiqwerty: I made `domain` a property and `self.data_model` is now the private `self._data_model`.

LukasHebing (commented):
If we agree, that this is the way to create the "real" strategy object, should we then get rid of the "map" function in all tutorials?

LukasHebing:
Generally, very nice. Thanks for taking care to also change the README, and the basic tutorial!

bertiqwerty:
> If we agree, that this is the way to create the "real" strategy object, should we then get rid of the "map" function in all tutorials?

At least for all tutorials that don't talk about the separation of data models and functional components, I agree.

bertiqwerty:
> What do you think of doing the same for surrogates in a seperate PR?

Interesting. Do you often create surrogates without strategies just for modelling?

tests/bofire/strategies/test_make_strategy.py
bertiqwerty: done.

jduerholt:
> Interesting. I have no objection. Do you often create surrogates without strategies just for modelling?

Yes, we do ;)

tests/bofire/strategies/test_make_strategy.py
bertiqwerty: Btw., I by doing that I found some strategies that were missing. I will add them.

bertiqwerty:
I think, I have addressed all the points. Please have a look again.

tests/bofire/strategies/test_make_strategy.py
```diff
@@ -0,0 +1,91 @@
...
+        FactorialStrategy,  # this is deprecated in favor of F, hence we don't add make
+    ]
+    for spec in strat_specs.valids:
+        data_model = spec.obj()
+        strat = strats.map(data_model)
+        if type(strat) not in strats_without_make:
+            # test the make function
+            test(strat, data_model)
+        else:
+            print(f"Skipping {strat} because it does not have a make function")
```
jduerholt, line 87: This would be my last question: should we not enforce that we have make methods by making it an abstract method of `Strategy`?

jduerholt (approved):
Looks good t me. Could be merged from my side. I only have one question. Next, we should do it with surrogates and also go over the strategies and maybe introduce (where necessary) also other properties that point to the data model. I think this could make things even cleaner.

tests/bofire/strategies/test_make_strategy.py
bertiqwerty: I don't really see the benefit for putting it in the base class. It will just mess up the linting, I guess. We make sure that all strategies have make via the test. We need an individual make for each strategy anyway.

franktoffel:
I followed the official BoFire tutorial, but encountered the following error:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[18], line 4
      1 from bofire.strategies.api import SoboStrategy
      2 from bofire.data_models.acquisition_functions.api import qLogEI
----> 4 sobo_strategy = SoboStrategy.make(
      5     domain=domain, acquisition_function=qLogEI(), seed=19
      6 )

AttributeError: type object 'SoboStrategy' has no attribute 'make'
```

I'm using a clean environment with Python **3.10** on **macOS (Apple Silicon)**.  
All packages were installed fresh using `conda` and `pip`.

Here is my current environment (from `pip list`):

```
bofire==0.1.0
botorch==0.14.0
gpytorch==1.14
torch==2.5.1
```

Any ideas?

Thanks for the good work.

PR #577, opened 2025-04-11
Expose `sequential` to data model

jduerholt:
## Motivation

Following discussion https://github.com/pytorch/botorch/discussions/2810, this PR adds the possibility to explicitly toggle between sequential and parallel batch candidate generation for the botorch optimizer. It applies only to the case where `optimize_acqf` is used, all other cases are sequential in itself.

Y.

## Test Plan

Not necessary for this PR.


Jimbo994:
Seems like a very sensible decision to be able to toggle between sequential and parallel after reading the discussion. Code changes also look good to me.

PR #584, opened 2025-05-15
Fix missing documentation by including subfolders automatically

rlars:
Fix missing documentation by including subfolders automatically with show_submodules.

## Motivation

(Write your motivation here.)

(Write your answer here.)

## Test Plan

(Write your test plan here. If you changed any code, please provide us with clear instructions on how you verified that your changes work.)

### DIFF
--- a/mkdocs.yaml
+++ b/mkdocs.yaml
@@ -15,7 +15,6 @@ nav:
     - Constraints: ref-constraints.md
     - Objectives: ref-objectives.md
     - Features: ref-features.md
-    - Domain Util: ref-domain-util.md
     - Utils: ref-utils.md
 
 theme:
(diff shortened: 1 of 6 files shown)


mkdocs.yaml
```diff
@@ -49,6 +48,10 @@ plugins:
   - mkdocs-jupyter
   # https://mkdocstrings.github.io/
   - mkdocstrings:
+      handlers:
```
jduerholt, line 51: What is this for?

jduerholt (approved):
Hi @rlars,

looks good to me. Thanks. 

I only had one question regarding understanding from my side ;)

Best,

Johannes

jduerholt:
@bertiqwerty: we also need to decide, if we want to have more modules in the API documentation in the future.

mkdocs.yaml
rlars: It includes the submodules automatically, e. g. bofire.data_models.features.categorical, bofire.data_models.features.continuous etc.
It should not be used on the top level bofire or bofire.data_models though, as it will put too much on one page.
Instead we may want to add a few more ref.md files

PR #592, opened 2025-05-27
Shap 

jduerholt:
## Motivation

This PR adds Shap Analysis to BoFire surrogates.

Yes.

## Test Plan

Unit tests. @jkeupp and @e-dorigatti: could you have a look at the current implementation, and tell me if it looks ok for you. Then I will implement unit tests (so far it is tested only in a notebook). Note that this version is able to deal with categoricals etc.

### DIFF
--- a/bofire/surrogates/feature_importance.py
+++ b/bofire/surrogates/feature_importance.py
@@ -3,6 +3,7 @@ from typing import Dict, Optional
 
 import numpy as np
 import pandas as pd
+import shap
 
 from bofire.data_models.enum import RegressionMetricsEnum
 from bofire.surrogates.diagnostics import metrics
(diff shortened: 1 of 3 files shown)


bofire/surrogates/feature_importance.py
```diff
@@ -69,6 +70,115 @@ def combine_lengthscale_importances(importances: Sequence[pd.Series]) -> pd.Data
...
+        )
+
+    # to be able to handle categorical inputs, we run the shap analysis on the
+    # already preprocessed data and use surrogate._predict to get the predictions.
+
+    preprocessed_bg_experiments = surrogate._prepare_data_for_predict(
+        bg_experiments[surrogate.inputs.get_keys()].copy()
+    )
+
+    preproccesed_experiments = surrogate._prepare_data_for_predict(
```
e-dorigatti: As already mentioned by Julian I would not run shap on the transformed inputs, otherwise each level of categorical variables gets a different explanation, and that is confusing and harder to understand. The shap library already handles categoricals out of the box so I'd move the preprocessing in the predict function

jduerholt: Ok, then I will change it in a way that by default the categoricals are handled by SHAP, and via a flag one can request to run it on transformed inputs. This could be insightful for descriptors. Thanks for having a look!

jduerholt:
@jkeupp and @e-dorigatti : have you tried it with categoricals? For me it creates errors ...

bofire/surrogates/feature_importance.py
```diff
@@ -69,6 +70,115 @@ def combine_lengthscale_importances(importances: Sequence[pd.Series]) -> pd.Data
...
+) -> shap.Explanation:
+    """Combines a list of SHAP explanations into one SHAP Explanation object.
+
+    Args:
+        shap_values: List of SHAP Explanation objects to combine.
+
+    Returns:
+        Combined SHAP Explanation object.
+    """
+    return shap.Explanation(
```
e-dorigatti: I am wondering how future-proof this is, but I checked shap's code and did not find any better ways of doing it. I guess we can leave it like this.

PR #596, opened 2025-06-04
Feature/modularize ga

LukasHebing:
Make GA available for other purposes


bertiqwerty:
Thanks, Lukas! @R-M-Lee should also have a look.

R-M-Lee:
> 
> What would you expect for the GA?

I think it would be useful to have an interface for using pymoo to estimate the pareto front from a pretrained multiobjective strategy. Here is what I mean in pseudo code:

```python
def estimate_pareto_front(strategy: MultiobjectiveStrategy)

    def objective_function(expt: pd.DataFrame) -> torch.Tensor:
        return torch.vstack([ -domain.outputs[i].objective(strategy.predict(expt)[i]) for i in domain.outputs]

    problem, algorithm, termination = get_ga_problem_and_algorithm(
        optimizer_data_model, strategy.domain, input_preprocessing_specs, [objective_function],
        q=q, n_obj= len(surrogates.surrogates)*q, verbose=True)

    return pymoo_minimize(problem, algorithm, termination, verbose=True)
```

Something like that. Is it already possible with the current changes?

jduerholt:
Looks good from my side. I like it!

Especially when focusing for now on sampling pareto fronts. Regarding the objective functions without surrogates, I cannot really follow ;) 

The question is of course if the callable should be even more abstract to make use of the common interface that every predictive strategy in BoFire has a `predict` method that takes as input a data frame and spits out a dataframe with the predictions. For the torch based strategies, it transforms the input data frame to a tensor using the `input_preprocessing_specs` of the strategy, does the prediction with the trained torch surrogates and transforms the predictions back to a dataframe. 

In this case, we would only need to convert from the GA space to the domain space, the rest would be handled via the predict functionality. Thinking about it, I recognize that we could in principle to the same for acqf optimization, as the botorch based strategies have the `calc_acquistion` method that takes as `predict` a dataframe as input:

https://github.com/experimental-design/bofire/blob/5e3a19e127b37e9352a4f917367e292b182d0c0e/bofire/strategies/predictives/botorch.py#L136

Of course, for exploting this, one would need to change the architecture of the acqf optimizer a bit ... and the same could be achieved by just buidling up the objective function using tensors and `strategy.model` as in `predict`:

https://github.com/experimental-design/bofire/blob/5e3a19e127b37e9352a4f917367e292b182d0c0e/bofire/strategies/predictives/botorch.py#L114

Note that we would also need to evaluate the objectives of the outputs on the model predictions and take care for maximization/minimization for pareto front sampling. But we can discuss this when we actually implement the pareto sampler ;) 

So for me the open question would be: What should be the signature of the callable that is passed to the pymoo optimizer, and which transformations are needed before data is entering the callable. In principle, the new usecase of pareto sampling can be totally covered by this implementation, but the pseudocode of Robert would not work ;) 

This were a lot of thoughts, I hope you can make some use out of it ;) We could of course, also setup a call ;)

jduerholt:
Sounds reasonable, for my perspective q-points are only needed for acqf optimization. We should then have max synergies between both options. Another idea would be to make the preprocessing more configurable. Currently it always transformed from pymoo space to the torch space via the domain space. We could make it configurable and give the option to only go from pymoo space to domain space, then we could use `predict` and `calc_acqf` out of the box. Btw, `calc_acqf` also supports q-points. Maybe, we could also get rid of the torch sapce in total, for this we would need to slighly change the architecture of the acqf optimization.

LukasHebing:
@R-M-Lee, @jduerholt 
The GA should be usable now for more purposes, using the domain with inputs and constraints. See e.g. the example in the .ipynb:

```python
def objective_function_1(x: List[pd.DataFrame]) -> np.ndarray:
    """assume we want to maximize the mean variance of the experiments dataframe"""
    vars = [xi.var(numeric_only=True).mean() for xi in x]
    return -np.array(vars)
#%%
def objective_function_2(x: List[pd.DataFrame]) -> np.ndarray:
    """Maximize the sum of all inputs"""
    vars = [xi.sum().sum() for xi in x]
    return -np.array(vars)
#%%
x_opt, f_opt = run_ga(
    data_model=optimizer,
    domain=domain,
    objective_callables=[
        objective_function_1,
        objective_function_2,
    ],  # list of objective functions to optimize
    q=3,  # number of points to optimize
    callable_format="pandas",
)
```

bofire/strategies/utils.py
```diff
@@ -216,6 +236,26 @@ def transform_to_experiments(self, X: List[dict]) -> List[pd.DataFrame]:
 
         return experiments_out
 
+    def transform_to_experiments_per_individual(
```
jduerholt, line 241: What is the difference of this methond in comparison to`transform_to_experiments_per_q_point` ?

bofire/strategies/utils.py
```diff
@@ -351,20 +432,48 @@ def __init__(
         )
 
     def _evaluate(self, x_ga_encoded, out, *args, **kwargs):
```
jduerholt: Can you add type hints here?

bofire/strategies/utils.py
```diff
@@ -766,11 +921,14 @@ def get_problem_and_algorithm(
             eliminate_duplicates=MixedVariableDuplicateElimination(), repair=repair
         )  # see https://github.com/anyoptimization/pymoo/issues/575
 
+    if n_obj > 1:
+        algorithm_args["survival"] = RankAndCrowdingSurvival()
+
     algorithm = MixedVariableGA(**algorithm_args)
```
jduerholt, line 930: Would it make sense to switch under the hood to a different GA when we do not have discrete and categorical degrees of freedom?

tests/bofire/strategies/test_optimizer.py
```diff
@@ -28,4 +35,69 @@ def test_optimizer(optimizer_benchmark, optimizer_data_model):
...
+        [objective_function],
+        q=q,
+        callable_format="torch",
+        n_obj=len(surrogates.surrogates) * q,
+        verbose=True,
+    )
+
+    result = pymoo_minimize(problem, algorithm, termination, verbose=True)
+
+    print(result.X)
```
jduerholt: Remove this line ;)

jduerholt (approved):
Looks good to me, I only let a few minor comments. You can fix and merge then.

Just a few questions:

- Would it make sense to expose also the algorithm part to the user? Currently, always this mixed GA is used. But if we do not have discretes or categoricals this would not be necessary, or? Either we code it in, or we expose it to the user or? Ofc, this is something for a seperate PR.
- Let's assume we want to do pareto front sampling (which was the original use case for this PR) and has in addition to the features that should be maximized/minimized also an output constraint active for example that an additional output should be always below a certain thresshold. How would one do this with the GA?
- Do we have somewhere documented if the GA assumes maximization or minimization?

I think the next step would then be to finish the PR with the notebook on pareto front sampling ;)

bofire/strategies/utils.py
LukasHebing: Could be possible. That would require a deep dive into the method implementation in pymoo. I would keep this as it is for now, as checking comparability and performance of the alternatives could get quite tedious.

LukasHebing: We could add more fields to the data model with different algorithm options. But, the logic of how they can (or cannot) be combined is quite complicated.

LukasHebing:
> Looks good to me, I only let a few minor comments. You can fix and merge then.
> 
> Just a few questions:
> 
> * Would it make sense to expose also the algorithm part to the user? Currently, always this mixed GA is used. But if we do not have discretes or categoricals this would not be necessary, or? Either we code it in, or we expose it to the user or? Ofc, this is something for a seperate PR.
Due to the complexity, I would say we keep this out for now

> * Let's assume we want to do pareto front sampling (which was the original use case for this PR) and has in addition to the features that should be maximized/minimized also an output constraint active for example that an additional output should be always below a certain thresshold. How would one do this with the GA?
The objective functions can be formulated freely. I would implement this as a soft-constraint (probably taking the distribution of the model prediction into account?)

> * Do we have somewhere documented if the GA assumes maximization or minimization?
It's now added in the docstring

> I think the next step would then be to finish the PR with the notebook on pareto front sampling ;)
Did you check out the examples in the GA notebook? What would you add there?

jduerholt:
Hi @LukasHebing, for me this is ready to merge, can you resolve the conflict on the notebook? Some changes were made there in https://github.com/experimental-design/bofire/pull/605

bofire/strategies/utils.py
R-M-Lee: I agree that it might be wise, in future, to use a different GA (like NSGA-III) if the problem is purely continuous. My anecdotal experience is that this can work better than the `MixedVariableGA` But I also agree that we can do it later if we want

Issue #595, opened 2025-06-04
[Bug]: Categorical DoE unstable


dlinzner-bcs:
### What happened?

Current implementation often can create segmentation fault error

### Please provide a minimal, reproducible example of the unexpected behavior.

def one_cont_3_cat():
    np.random.seed(42)
    torch.manual_seed(42)
    # Test case: extra experiments
    domain = Domain(
        inputs=Inputs(
            features=[
                ContinuousInput(key="x1", bounds=(0, 1)),
                CategoricalInput(key="cat1", categories=["miau", "meow"]),
                CategoricalInput(key="cat2", categories=["oink", "oinki", "grunt"]),
                CategoricalInput(key="cat3", categories=["wuff", "wuffwuff", "ruff"]),
            ]
        ),
        outputs=Outputs(features=[ContinuousOutput(key="y")]),
    )

    obj = DOptimalDesign(
        domain=domain,
        model_type="linear",
        objective=DoEObjectiveEnum.D_OPTIMALITY,
        objective_extra_experiments=DoEObjectiveEnum.D_OPTIMALITY,
        ipopt_options={"max_iter": 1000},
    )

    data = obj.ask(n_proposals=20)
    print(data)
    assert np.shape(data.to_numpy()) == (20, 4)

### Please paste any relevant traceback/logs produced by the example provided.

```shell
Fatal Python error: Segmentation fault
```

### BoFire Version

0.2.0

PR #597, opened 2025-06-04
595 bug categorical doe unstable

dlinzner-bcs:
## Motivation

[edit Behrang]
The actual problem was a segmentation fault. This has been fixed by checking that smart rounding is only applied if discrete inputs exist. Further, this PR contains additional fixes that were found during debugging.

- A faulty n-choose-k constraint avoided categoricals to be optimized. This has been removed. However, this constraint made sure that the number of experiments was correct. The actual reason for the wrong number of experiments was a wrong formulaic formula. This formula has been fixed.
- Underscores and minus signs in formula strings made problems. Has been fixed by using unambiguous strings instead.
- In the creation of the formula string we have to separate numerical and categoricals because we cannot calculate with categorical.

[/edit]

[edit Behrang]
Yes
[/edit]

## Test Plan

[edit Behrang]
Unit tests have been extended.
[/edit]

### DIFF
--- a/bofire/strategies/doe/utils.py
+++ b/bofire/strategies/doe/utils.py
@@ -50,7 +50,7 @@ def represent_categories_as_by_their_states(
         )
 
         # enforce categoricals excluding each other
-        all_but_one_categoricals = categorical_one_hot_variabes
+        all_but_one_categoricals = categorical_one_hot_variabes[:-1]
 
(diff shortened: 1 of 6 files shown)


dlinzner-bcs:
link to issue in this PR explains it. This PR closes [#597 ](https://github.com/experimental-design/bofire/issues/595)

Issue #579, opened 2025-05-07
[FEATURE REQUEST]: Support alternating acqf optimization


TobyBoyne:
### Motivation

Currently, performing Bayesian optimization over mixed spaces in BoFire requires enumeration over all of the discrete combinations, and optimizing each subproblem. This is expensive, and becomes infeasible when the number of combinations explodes combinatorially. For example, the Cell Network problem in the [Prob Reparam](https://arxiv.org/pdf/2210.10199) paper has $6^{15}$ combinations.

BoFire should support the alternating approach for optimizing the acqf proposed in Wu et al. (2021).

Wu et al. 2021 [Think Global and Act Local: Bayesian Optimisation over High-Dimensional Categorical and Mixed Search Space](https://arxiv.org/pdf/2102.07188)

### Describe the solution you'd like to see implemented in BoFire.

In the `BoTorchOptimizer` class, we should allow the user to select the `optimize_acqf_mixed_alternating` [1] approach to optimize the acquisition function.  https://github.com/experimental-design/bofire/blob/03b1f320c62d0af990e66268d16f23007c34c35f/bofire/strategies/predictives/acqf_optimization.py#L433

[1] https://botorch.readthedocs.io/en/stable/optim.html#botorch.optim.optimize_mixed.optimize_acqf_mixed_alternating

### Describe any alternatives you've considered to the above solution.

More complicated methods, such as Probabilistic Reparameterization, would be great once they have been merged into BoTorch. Simpler methods, such as continuous relaxation, have been shown to not have great performance.

### Pull Request

Yes

TobyBoyne:
Hi @jduerholt,

You are right that the method doesn't support categoricals. I will check what Ax does and get back to you.

Even though this function only works for discrete, is it not still worth adding to BoFire in the case that all your inputs are continuous/discrete/binary? Or do you think we should only add the alternating approach to BoFire once it supports categoricals?

Thanks,
Toby

jduerholt:
Let us have a look how Ax is doing it and then decide. @LukasHebing also added an GA for highly categorical search spaces, did you already had a look at it?

LukasHebing:
I didn't do any studies on that topic. As far as I can say is, that GAs are usually quite robust in all search spaces. This is why we chose a GA for our app, where we don't see which search spaces the users provide. 

However, in comparison to gradient-based optimizations, the GA is really inefficient. So it might be good to take a look at other mixed-integer algorithms.

[...]

Issue #459, opened 2024-11-12
update kernels of mixed GPs


R-M-Lee:
Now we have THREESIX, HVARFNER and MBO kernels. Mixed GPs (hyperparam config, etc) needs updating

PR #603, opened 2025-06-13
Ordinal encoding for surrogates for mixed alternating optimization

TobyBoyne:
## Motivation

This PR removes the requirement that Botorch surrogates must have a one-hot encoding. This should make BoFire more flexible. This PR also introduces mixed alternating optimization (https://github.com/pytorch/botorch/pull/2866), which requires inputs to have an integer encoding.

See issue #579. I would also recommend addressing #459 in this PR, since we are changing the MixedSingleTaskGP model anyway.

Yes

## Test Plan

- Test the new mixed optimizer in line with `test_optimizer.py`
- Test the MixedSingleTaskGP model, compare that the predictions are equivalent regardless of input encoding.

### DIFF
--- a/bofire/data_models/domain/features.py
+++ b/bofire/data_models/domain/features.py
@@ -475,7 +475,7 @@ class Inputs(_BaseFeatures[AnyInput]):
 
         num_values = num_cats + num_discretes
 
-        return functools.reduce(operator.mul, num_values)
+        return functools.reduce(operator.mul, num_values, 1)
 
     # transformation related methods
     def _get_transform_info(
--- a/bofire/data_models/domain/features.py
+++ b/bofire/data_models/domain/features.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
+import functools
 import itertools
+import operator
 import re
 import warnings
 from collections.abc import Iterator, Sequence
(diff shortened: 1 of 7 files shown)


TobyBoyne:
cc @jduerholt, following from our discussion before.

I have converted this to draft as the PR hasn't yet been released in a

jduerholt:
Hi @TobyBoyne, thank you very much? One quick question without looking at the code ;) Did you get it working? Best, Johannes

TobyBoyne:
Yes, it seems to be working! I created a `SoboStrategy` with a `MixedSingleTaskGPSurrogate`, and was able to call `strategy.ask` which all works and calls the correct optimizer.

I'm now working on fixing the GP to support both onehot and ordinal encodings.

jduerholt:
> * ...which means there should be a way to compute the number of combinations without creating a list of all of them, which can cause OOM. Should this be in the `Input` class?

Hmm, what do you mean with be in the `Input` class? We could compute a rough estimate of the number of combinations by just taking the product of all discrete options (for DiscreteInput) and allowed categories (for Categorical Input)? Then based on this assesment, we could decide for either `optimize_acqf_mixed` or `optimize_acqf_mixed_alternating`. 

> * What happens if BotorchOptimizer.categorical_method != BotorchOptimizer.discrete_method? Exhaustive and Free use two different optimizers, and it isn't possible use them both simultaneously.

Honestly, I would prefer to get rid of the `categorical_method`, `discrete_method` and `molecular_method` options. Their intention was to be able to switch to a continuous relaxation for these features but I think nobody does it, and with the alternating optimizer and the GA we have a good set of options. And if somebody really want to run there over a relaxed space, one could setup the domain upfront like this. 

What do you think?

cc: @bertiqwerty @LukasHebing @dlinzner-bcs

TobyBoyne:
Sorry for not being clear. Currently, there is a method `Inputs.get_categorical_combinations`. I was simply suggesting another method, `Inputs.get_number_of_categorical_combinations`, which as you say would just take the product of all discrete/categorical options. Fairly trivial to do, I just wasn't sure where to put that logic - whether it is better in the `Inputs` class, or just where we would use it in the `BoTorchOptimizer`.

> get rid of `*_method` options

I agree with this - the alternating optimizer is fast enough, I don't think we need to support continuous relaxation, especially when it has been shown to not work very well.

jduerholt:
> Fairly trivial to do, I just wasn't sure where to put that logic - whether it is better in the `Inputs` class, or just where we would use it in the `BoTorchOptimizer`.

I would put it into to `Inputs` class.

LukasHebing:
> Honestly, I would prefer to get rid of the `categorical_method`, `discrete_method` and `molecular_method` options. So feel free to remove it. Their intention was to be able to switch to a continuous relaxation for these features but I think nobody does it, and with the alternating optimizer and the GA we have a good set of options. And if somebody really want to run there over a relaxed space, one could setup the domain upfront like this.
> 
> What do you think?
> 
> cc: @bertiqwerty @LukasHebing @dlinzner-bcs

Totally support this. I never looked into this deeply, but it was quite annoying to write around that. We don't use this as well.

jduerholt:
This PR was integrated into this one: https://github.com/experimental-design/bofire/pull/626

PR #604, opened 2025-06-17
Feature/exclude constraints in bo

jduerholt:
## Motivation

This PR introduces the use of `CategoricalExcludeConstraint`s in botorch based BO strategies. Currently it is only supported for fully discrete search spaces. Soon, it will also be available for mixed ones using the botorch acqf optimizer. For this reason, the code for the validator was duplicated, as it will be in the future independent validators, since the two acqf optimizers will support different things there.

Yes.

## Test Plan

Unit tests.

### DIFF
--- a/bofire/data_models/strategies/predictives/acqf_optimization.py
+++ b/bofire/data_models/strategies/predictives/acqf_optimization.py
@@ -266,6 +266,7 @@ class GeneticAlgorithmOptimizer(AcquisitionOptimizer):
             constraints.ProductInequalityConstraint,
             constraints.NonlinearInequalityConstraint,
             constraints.NChooseKConstraint,
+            constraints.CategoricalExcludeConstraint,
         ]
 
     def validate_domain(self, domain: Domain):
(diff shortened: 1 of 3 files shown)


LukasHebing:
So, currently this is only filtering the discrete choices in `_optimize_acqf_discrete`, or did I miss something? That is  fine to do so.

bofire/data_models/strategies/predictives/acqf_optimization.py
```diff
@@ -241,9 +266,30 @@ def is_constraint_implemented(self, my_type: Type[constraints.Constraint]) -> bo
             constraints.ProductInequalityConstraint,
             constraints.NonlinearInequalityConstraint,
             constraints.NChooseKConstraint,
+            constraints.CategoricalExcludeConstraint,
```
LukasHebing, line 269: It is a bit puzzling, that the constraint-type is added as supported constraint, but it is actually only possible to use this in the "all-categorical" mode. We should add the support into the real GA soon (or make a transformation to a NonlinearConstraint).

LukasHebing (approved):
All fine for me. The changes are also not big.

jduerholt:
> So, currently this is only filtering the discrete choices in `_optimize_acqf_discrete`, or did I miss something? That is fine to do so.

This is correct!

Issue #593, opened 2025-05-27
[FEATURE REQUEST]: `is_fulfilled` method for `Inputs`


jduerholt:
### Motivation

It would be helpful to have an `is_fulfilled` method for the `Inputs` object as for the `Constraints` object. With this method one can then quickly check which experiments match the current constraints just acting on the inputs, as for example lower and upper bounds for a continuous input. 

### Describe the solution you'd like to see implemented in BoFire.

Implement `is_fulfilled` on the feature level and call it then from a global is_fulfilled from the `Inputs` class. 

### Is this related to an existing issue in BoFire or another repository? If so please include links to those issues here.

Would be especially helpful for the `FeasibleExperimentCondition` in https://github.com/experimental-design/bofire/pull/580

### Pull Request

None

PR #605, opened 2025-06-23
`is_fulfilled` for `Domain` and `Inputs`

jduerholt:
## Motivation

This PR introduces `is_fulfilled` functionality for `Inputs` and `Domain`. It is especially needed for setting up acquisition functions in use cases where the constraints for the actual optimization differ from them used to generate the historical data. It could be that for example the constraints exclude the best experiments seen so far, but if they are then provided for the acqf setup, the (expected improvement based) acqf are not returning meaningful values during the optimization with the new constraints.

Closes https://github.com/experimental-design/bofire/issues/593.

Y.

## Test Plan

Unit tests.


bofire/strategies/predictives/botorch.py
```diff
@@ -219,12 +219,21 @@ def get_acqf_input_tensors(self):
...
             keep="first",
             inplace=False,
         )
+        # we should only provide those experiments to the acqf builder in which all
+        # input constraints are fulfilled, output constraints are handled directly
+        # in botorch
+        clean_experiments = clean_experiments[
+            self.domain.is_fulfilled(clean_experiments)
+        ].copy()
+        if len(clean_experiments) == 0:
```
jduerholt, line 233: @LukasHebing @bertiqwerty: I am not sure here, if it is really the best way to error out here, maybe it is better to raise a warning if no feasible experiments are found and then provide the experiments that are available. What do you think?

bofire/data_models/features/continuous.py
```diff
@@ -93,6 +93,21 @@ def round(self, values: pd.Series) -> pd.Series:
...
+        """Method to check if the values are within the bounds of the feature.
+
+        Args:
+            values: A series with values for the input feature.
+
+        Returns:
+            A series with boolean values indicating if the input feature is fulfilled.
+        """
+
+        noise = 10e-6
```
bertiqwerty: make noise a parameter with default value `10-e6`?

bofire/strategies/predictives/botorch.py
bertiqwerty: I am personally fine with raising an error.

bertiqwerty (approved):
Thanks, Johannes!

jduerholt:
> I am personally fine with raising an error.

Then we keep it ;)

PR #626, opened 2025-08-15
Refactor Input Preproccesing and Mixed Optimization

jduerholt:
## Motivation

This PR contains a major refactoring of BoFire. Here is a short summary:

- botorch introduced the `optimize_acqf_mixed_alternating` functionality, which thanks to @TobyBoyne can now also deal with categoricals. I added a few other features there so that it is now ready for use within BoFire. `optimize_acqf_mixed_alternating` performs a block gradient descent optimization of the acqf and is an alternative to the GA on large mixed domains.
- `optimize_acqf_mixed_alternating` expects the categoricals always in an ordinal encoding. So far, BoFire was transforming the categoricals upfront to a vector encoding (like one-hot) before entering the actual surrogate. This will now change, categoricals will be always ordinal encoded, and different vector encodings can be realized by using a botorch input transform namely the `NumericToCategoricalEncoding` (https://github.com/pytorch/botorch/blob/b1097c6c475f29f694532c3282393ce8a67a9d6c/botorch/models/transforms/input.py#L1628C7-L1628C35). For this reason, the meaning of the input_preproccesing_specs will change, it is now not a transformation that is applied before the data enters the actual botorch model, instead it is a transformation which is applied within the model.
- This change will make setting up the acqf optimization much cleaner as we do not have to account on this level for the different encodings.
- I am currently rewriting the `AcquisitionOptimizer`. @LukasHebing: why is the `domain` not an attribute of the acquisition optimizer? Would make it cleaner, or? I do not remember, why we decided to do it in the way we did it :D

Yes

## Test Plan

Unit tests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,7 +71,7 @@ all = [
     "cvxpy[CLARABEL,SCIP]",
     "scikit-learn>=1.0.0,<1.6",
     "entmoot>=2.0.6",
-    "pyomo<=6.9.4"
+    "pyomo<=6.9.4",
     "xgboost>=1.7.5",
     "rdkit>=2023.3.2",
     "mordred>=1.2.0",


bofire/data_models/priors/constraint.py
LukasHebing, line 1: Doesn't have botorch also a great library for priors, which we use anyway later?
Is it maybe possible to just store a reference to the botorch priors / constraints with parameters, instead of building this elaborated structure? That would be more flexible, otherwise you need to add the infrastructure for all the new priors here as well.

jduerholt:
@bertiqwerty @TobyBoyne @LukasHebing 

Anybody volunteering to review, or should I provide more info etc?

Best,

Johannes

bofire/data_models/surrogates/mixed_single_task_gp.py
```diff
@@ -100,27 +105,81 @@ def _update_hyperparameters(
 class MixedSingleTaskGPSurrogate(TrainableBotorchSurrogate):
     type: Literal["MixedSingleTaskGPSurrogate"] = "MixedSingleTaskGPSurrogate"
     continuous_kernel: AnyContinuousKernel = Field(
-        default_factory=lambda: MaternKernel(
-            ard=True, nu=2.5, lengthscale_prior=HVARFNER_LENGTHSCALE_PRIOR()
+        default_factory=lambda: RBFKernel(
```
e-dorigatti, line 108: Is there a reason why the default changed?

bofire/strategies/utils.py
```diff
@@ -49,13 +48,10 @@ def _default_input_preprocessing_specs(
     domain: Domain,
 ) -> InputTransformSpecs:
     """Default input preprocessing specs for the GA optimizer: If none given, will use OneHot encoding for all categorical inputs"""
```
e-dorigatti: doc needs to be updated (" ... will use ordinal encoding ...")

bofire/data_models/priors/constraint.py
```diff
@@ -7,3 +7,35 @@ class PriorConstraint(BaseModel):
     """Abstract Prior Constraint class."""
 
     type: Any
```
TobyBoyne, line 9: (Minor) Why is this not `type: str`?

bofire/data_models/surrogates/botorch.py
```diff
@@ -21,40 +30,55 @@ def validate_input_preprocessing_specs(cls, v, info):
...
         inputs = info.data["inputs"]
-        categorical_keys = inputs.get_keys(CategoricalInput, exact=True)
-        descriptor_keys = inputs.get_keys(CategoricalDescriptorInput, exact=True)
-        molecular_keys = inputs.get_keys(MolecularInput, exact=True)
+        categorical_keys = inputs.get_keys(CategoricalInput, exact=False)
         for key in categorical_keys:
             if (
-                v.get(key, CategoricalEncodingEnum.ONE_HOT)
-                != CategoricalEncodingEnum.ONE_HOT
+                v.get(key, CategoricalEncodingEnum.ORDINAL)
```
TobyBoyne, line 53: (MInor) Why is the default here `ORDINAL`, when the default value generated in `_default_categorical_encodings` below is `ONE_HOT`? I don't think think this actually makes a difference, but to me it isn't clear what the intended default behaviour should be.

bofire/data_models/surrogates/mixed_single_task_gp.py
```diff
@@ -100,27 +105,81 @@ def _update_hyperparameters(
...
+            ard=True,
+            lengthscale_prior=HVARFNER_LENGTHSCALE_PRIOR(),
+            lengthscale_constraint=GreaterThan(lower_bound=2.500e-02),
         )
     )
     categorical_kernel: AnyCategoricalKernel = Field(
-        default_factory=lambda: HammingDistanceKernel(ard=True),
+        default_factory=lambda: HammingDistanceKernel(
+            ard=True, lengthscale_constraint=GreaterThan(lower_bound=1.000e-06)
```
TobyBoyne, line 116: I wonder if it's worth considering placing a prior on this lengthscale as well? I think the Hvarfner paper only considers the purely continuous case, but surely their argument extends to mixed kernels?

bofire/data_models/surrogates/botorch.py
```diff
@@ -21,40 +30,55 @@ def validate_input_preprocessing_specs(cls, v, info):
...
-                )
-            v[key] = mol_encoding
+        return v
+
+    @classmethod
+    def _default_categorical_encodings(
+        cls,
+    ) -> dict[Type[CategoricalInput], CategoricalEncodingEnum | Fingerprints]:
+        return {
+            CategoricalInput: CategoricalEncodingEnum.ONE_HOT,
```
TobyBoyne, line 72: In `MixedSingleTaskGPSurrogate._default_categorical_encodings`, the defaults are `ORDINAL`. Why are they `ONE_HOT` here? I was under the impression that this was because you can have a purely continuous `BotorchSurrogate`, but then it looks like the check above in `validate_input_preprocessing_specs` means that you can't use `ONE_HOT` for any Botorch surrogate?

bofire/data_models/types.py
```diff
@@ -87,7 +87,7 @@ def validate_monotonically_increasing(sequence: Sequence[float]) -> Sequence[flo
 
 Descriptors = Annotated[
     List[str],
-    Field(min_length=1),
+    Field(min_length=2),
```
TobyBoyne, line 90: (Minor) Is there no case for having descriptors of length 1? I imagine it's rare but I don't see the benefit of preventing that behaviour?

bofire/strategies/predictives/acqf_optimization.py
```diff
@@ -141,27 +135,26 @@ def _optimize(
 
     @staticmethod
     def _features2idx(
-        domain: Domain, input_preprocessing_specs: InputTransformSpecs
+        domain: Domain,
     ) -> Dict[str, Tuple[int]]:
         features2idx, _ = domain.inputs._get_transform_info(
-            input_preprocessing_specs,
+            {},
```
TobyBoyne, line 149: Why does this not use `AcquisitionOptimizer._input_preprocessing_specs(domain)` here?

bofire/strategies/predictives/acqf_optimization.py
```diff
@@ -141,27 +135,26 @@ def _optimize(
...
 
     @staticmethod
-    def _features2names(
-        domain, input_preprocessing_specs: InputTransformSpecs
-    ) -> Dict[str, Tuple[str]]:
-        _, features2names = domain.inputs._get_transform_info(
-            input_preprocessing_specs,
-        )
-        return features2names
+    def _input_preprocessing_specs(
```
TobyBoyne, line 154: Why is this a private static method, and not just a private function in this file? It is also called inconsistently, sometimes as `self._input_preprocessing_specs`, and sometimes as `AcquisitionOptimizer._input_preprocessing_specs`.

bofire/strategies/predictives/acqf_optimization.py
```diff
@@ -285,10 +266,83 @@ def _optimize_acqf_discrete(
...
+
+
+class _OptimizeAcqfMixedInput(_OptimizeAcqfInputBase):
+    acq_function: Callable
+    bounds: Tensor
+    q: int
+    num_restarts: int
+    fixed_features_list: List[Dict[int, float]]  # it has to have more than two items
+    raw_samples: int
+    options: dict[str, bool | float | int | str] | None
```
TobyBoyne, line 309: (Minor) Consistency with PEP 585 and 604 style annoyations (`dict` vs `Dict`, `Optional` vs `| None`). There are a few of these elsewhere

bofire/strategies/predictives/acqf_optimization.py
```diff
@@ -285,10 +266,83 @@ def _optimize_acqf_discrete(
...
+    num_restarts: int
+    raw_samples: int
+    options: dict[str, bool | float | int | str] | None
+    inequality_constraints: list[tuple[Tensor, Tensor, float]] | None
+    equality_constraints: list[tuple[Tensor, Tensor, float]] | None
+    nonlinear_inequality_constraints: list[tuple[Callable, bool]] | None
+    fixed_features: dict[int, float] | None
+    sequential: bool
+    ic_generator: Callable | None
+    generator: Any
```
TobyBoyne, line 299: What is the difference between these?

bofire/strategies/predictives/acqf_optimization.py
```diff
@@ -495,37 +466,45 @@ def _get_optimizer_options(self, domain: Domain) -> Dict[str, int]:
...
+        n_categorical_combinations = (
+            domain.inputs.get_number_of_categorical_combinations()
+        )
+        if n_categorical_combinations == 1:
+            return "optimize_acqf"
+        if (
+            n_categorical_combinations <= ALTERNATING_OPTIMIZER_THRESHOLD
+            or len(get_nonlinear_constraints(domain)) > 0
+        ):
+            return "optimize_acqf_mixed"
```
TobyBoyne: If there are 10,000 categorical combinations, but there are some nonlinear constraints, what happens here? Will `optimize_acqf_mixed` be called, which will end up hanging because the optimization takes too long?

bofire/surrogates/botorch.py
```diff
@@ -75,3 +92,43 @@ def loads(self, data: str):
         """Loads the actual model from a base64 encoded pickle bytes object and writes it to the `model` attribute."""
         buffer = io.BytesIO(base64.b64decode(data.encode()))
         self.model = torch.load(buffer, weights_only=False)
+
+
+class TrainableBotorchSurrogate(BotorchSurrogate, TrainableSurrogate):
```
TobyBoyne, line 98: I'm a bit confused by this class: aren't all BotorchSurrogates trainable? Also, some surrogates (`MLPEnsemble`, `PiecewiseLinearGPSurrogate`, ...) inherit from `BotorchSurrogate` and `TrainableSurrogate`, which to me seems that they should just inherit `TrainableBotorchSurrogate`?

bofire/utils/torch_tools.py
```diff
@@ -1093,3 +1101,69 @@ def create_supervised_dataset(
...
+        for feat in inputs.get(CategoricalInput)
+        if transform_specs.get(feat.key, CategoricalEncodingEnum.ORDINAL)
+        is not CategoricalEncodingEnum.ORDINAL
+    }
+    if len(encoders) > 0:
+        return NumericToCategoricalEncoding(
+            dim=len(inputs.get()),
+            categorical_features={i: enc.dim for i, enc in encoders.items()},
+            encoders={
+                i: enc.__call__  # type: ignore
```
TobyBoyne: (Minor) Can you not just write this as `enc` instead of `enc.__call__`?

bofire/data_models/surrogates/mixed_single_task_gp.py
jduerholt: This is now the same default that also botorch has.

bofire/strategies/utils.py
jduerholt: Fixed.

bofire/data_models/priors/constraint.py
jduerholt: This is our new convention, since the non abstract classes have the `type` attribute always of type `Literal`, setting it to `str` in the abstract class will lead to errors from the type checker.

bofire/data_models/surrogates/botorch.py
jduerholt: These are two different things:

- `input_preproccessing_specs` are applied to all models, and if it is a botorch model it has to be `ORDINAL` for all categoricals (and derived) features.
- `categorical_encodings` is only available in botorch based models and defines how categoricals are treated within the botorch model via a  botorch `InputTransform`. Here the default is `ONE_HOT`. Via this mechanism, one can apply different categorical encodings while having them for the acqf optimizer always ordinally encoded.

Is this making it clear?

bofire/data_models/surrogates/mixed_single_task_gp.py
jduerholt: Have not yet thought about this, this implementation reflects the botorch defaults.

bofire/data_models/surrogates/botorch.py
jduerholt: This question goes in the same direction as the one from above, `input_preprocessing_specs` and `categorical_encodings` are two differnt things. For the MixedGP, we always want to have at least one ordinal encoded input, on which the categorical kernel operates. So it is the special case, in general it is one-hot.

bofire/data_models/types.py
jduerholt: If you want to work with a one-dimensional descriptor, you can/should use the `DiscreteInput` istead.

bofire/strategies/predictives/acqf_optimization.py
jduerholt: It is the same, if the cateogircal is encoded as ordinal or not encoded at all, it will not have any influence on the order of the features.

bofire/strategies/predictives/acqf_optimization.py
jduerholt: I am always puzzled here, which one should be used nowadys? dict or Dict? Optional or | None?

bofire/strategies/predictives/acqf_optimization.py
jduerholt: This stems from some problematic botorch syntac for `optimize_acqf`. The`ic_generator` is the function to generate starting conditions, we use `gen_batch_initial_conditions` from botorch, via `generator` specifies a keyword argument parsed to `ic_generator`. The botorch syntax is really puzzling there ...

bofire/strategies/predictives/acqf_optimization.py
jduerholt: Yes, we have to bring nonlinears also to `optimize_acqf_mixed_alternating`.

bofire/data_models/surrogates/botorch.py
TobyBoyne: Ah, I think I understand. I find the naming a little confusing:
- Since the values of `input_preprocessing_specs` are `CategoricalEncodingEnum`, but is completely separate from `categorical_encodings`
- The type of `categorical_encodings` is `InputTransformSpecs`
To me, it is not very obvious the difference that you have outlined here. Maybe a docstring/comment somewhere explaining what you have explained here?

bofire/strategies/predictives/acqf_optimization.py
TobyBoyne: I believe `dict` is preferred over `Dict`. And I believe `| None` is preferred, see [here](https://discuss.python.org/t/clarification-for-pep-604-is-foo-int-none-to-replace-all-use-of-foo-optional-int/26945/6) 

I don't think this is a big issue, and re-typing the whole codebase in either style is beyond the scope of this PR at the very least! Just wanted to note that the inconsistency within a single class (L299 vs L301) stuck out to me.

bofire/strategies/predictives/acqf_optimization.py
TobyBoyne: I think it might be worth at the very least raising a warning here until nonlinear constraints are available in the alternating optimizer, since this will just crash with OoM as it is.

bofire/surrogates/botorch.py
jduerholt: This is a good catch, I will update it.

bofire/data_models/surrogates/botorch.py
LukasHebing: Sorry, I also struggle to understand the scope of `input_preprocessing_specs`:
- When `ORDINAL` is the only possible option. Why do we have other options. 
- Where are the user-specified encodings (descriptor, etc.)?

bofire/surrogates/botorch.py
jduerholt: Fixed it for MLP ensemble, for PiecewieseLinearGPSurrogate, we let it as is for now, as this is just a very expert model for one specific internal use case on which we are currently doing a lot of tinkering, for this reason, I would let it as is for now.

bofire/utils/torch_tools.py
jduerholt: Done, good catch!

bofire/data_models/surrogates/botorch.py
jduerholt: `InputTransformSpecs` is just `Dict[str, CategoricalEncodingEnum]` and it is used for both `input_preprocessing_specs` and `categorical_encodings.` I will add a doc string.

jduerholt: I added a docstring.

jduerholt:
@TobyBoyne: thanks for your comments, I went over them. You can have a look again ;)

LukasHebing:
Hi @jduerholt:
I wanted to have a look at the flow of variables in the code, but (as in the tests), I still get an import error:
ImportError: cannot import name 'NumericToCategoricalEncoding' from 'botorch.models.transforms.input'
I would retry this, when that is fixed, ok?

bofire/strategies/utils.py
```diff
@@ -48,14 +47,11 @@
 def _default_input_preprocessing_specs(
     domain: Domain,
 ) -> InputTransformSpecs:
-    """Default input preprocessing specs for the GA optimizer: If none given, will use OneHot encoding for all categorical inputs"""
-    input_preprocessing_specs = {}
-    for input_ in domain.inputs.get():
-        if isinstance(input_, CategoricalDescriptorInput):
```
LukasHebing, line 54: I guess, the `CategoricalDescriptorInput` will be handled in the botorch transformation?

jduerholt:
@LukasHebing: the problem is that this input transform is not yet in the latest botorch release (it is in main and will be part of the next release), so one would need to install botorch from source for this. I can just copy the source code in here, should I? And as soon as it is in botorch release, we remove it again from bofire, maybe this is the smartest way of dealing with it ...

LukasHebing:
> @LukasHebing: the problem is that this input transform is not yet in the latest botorch release (it is in main and will be part of the next release), so one would need to install botorch from source for this. I can just copy the source code in here, should I? And as soon as it is in botorch release, we remove it again from bofire, maybe this is the smartest way of dealing with it ...

Ah, now I understand the problem. Is it possible to use the github branch as a source, instead of the pypi release? I know that is possible with poetry or uv. 

Otherwise, we need to wait for the release anyway before we merge this PR, or?

I'll try to add botorch manually and check that.

bofire/data_models/priors/constraint.py
jduerholt: Hmm, we are using the botorch priors, what I added was the option to use more of them also within our data models.

bofire/strategies/utils.py
jduerholt: Yes!

tests/bofire/strategies/test_acqf_optimization.py
```diff
@@ -0,0 +1,283 @@
...
+    BotorchOptimizer,
+    _OptimizeAcqfInput,
+    _OptimizeAcqfListInput,
+    _OptimizeAcqfMixedAlternatingInput,
+    _OptimizeAcqfMixedInput,
+)
+from bofire.utils.torch_tools import tkwargs
+
+
+def test_determine_optimizer():
```
TobyBoyne, line 32: (Very minor) do you need to do any kind of testing/handling for when the domain is mixed *and* `n_acqfs>1`?

tests/bofire/strategies/test_acqf_optimization.py
```diff
@@ -0,0 +1,283 @@
...
+    domain.inputs.features.append(DiscreteInput(key="x_discrete", values=[0, 1, 2, 3]))
+    optimizer_args = optimizer._get_arguments_for_optimizer(
+        optimizer="optimize_acqf_mixed_alternating",
+        domain=domain,
+        candidate_count=1,
+        acqfs=[simple_acqf],
+        bounds=get_bounds(domain),
+    )
+    assert isinstance(optimizer_args, _OptimizeAcqfMixedAlternatingInput)
+    assert optimizer_args.discrete_dims == {6: [0, 1, 2, 3]}
```
TobyBoyne: (Minor) Could you please change the values passed into `DiscreteInput` to not be ordinal? It isn't clear to me from this test whether the expected behaviour is to map the input values to an ordinal encoding (as in categorical features), or to retain the original values.

TobyBoyne (approved):
Finally had the chance to look through everything again, thank you for waiting :)

Everything looks good, only remaining unresolved comments are pretty minor. Approved!

bofire/utils/torch_tools.py
```diff
@@ -1093,3 +1101,218 @@ def create_supervised_dataset(
...
+        r"""Check if another input transform is equivalent.
+
+        Args:
+            other: Another input transform.
+
+        Returns:
+            A boolean indicating if the other transform is equivalent.
+        """
+        return (
+            type(self) is type(other)
```
bertiqwerty: put all lines in parentheses (or none but i think it is good to be explicit about operator precedence).

bofire/surrogates/random_forest.py
```diff
@@ -115,30 +112,32 @@ def __init__(
...
-            ty = torch.from_numpy(Y.values).to(**tkwargs)
-            ty = output_scaler(ty)[0].numpy()
+        if outcome_transform is not None:
+            ntY = outcome_transform(tY)[0].numpy()
         else:
-            output_scaler = None
-            ty = Y.values
+            ntY = tY.numpy()
+
+        # if self.output_scaler == ScalerEnum.STANDARDIZE:
```
bertiqwerty: remove uncommented code

bofire/strategies/predictives/botorch.py
```diff
@@ -187,7 +187,7 @@ def _ask(self, candidate_count: int) -> pd.DataFrame:  # type: ignore
             candidate_count,
             acqfs,
             self.domain,
-            self.input_preprocessing_specs,
+            # self.input_preprocessing_specs,
```
bertiqwerty: remove uncommented code

bofire/strategies/predictives/acqf_optimization.py
```diff
@@ -495,37 +466,45 @@ def _get_optimizer_options(self, domain: Domain) -> Dict[str, int]:
             "maxiter": self.maxiter,
         }
 
-    def _setup_ask(
+    def _determine_optimizer(self, domain: Domain, n_acqfs) -> str:
```
bertiqwerty: the returned optimizer should be an enum.

bofire/strategies/predictives/acqf_optimization.py
```diff
@@ -537,151 +516,112 @@ def _setup_ask(
...
 
     def get_categorical_combinations(
         self,
         domain: Domain,
-        input_preprocessing_specs: InputTransformSpecs,
     ) -> List[Dict[int, float]]:
         """Provides all possible combinations of fixed values
 
         Returns:
             list_of_fixed_features List[dict]: Each dict contains a combination of fixed values
```
bertiqwerty: no need to duplicate type info from functin signature, prefer `list` over `List` and `dict` over `Dict`.

bofire/strategies/predictives/acqf_optimization.py
```diff
@@ -690,35 +630,14 @@ def get_categorical_combinations(
...
-                    for j, idx in enumerate(features2idx[feat]):
-                        fixed_features[idx] = transformed.values[0, j]
-                elif isinstance(feature, CategoricalInput):
-                    # it has to be onehot in this case
-                    transformed = feature.to_onehot_encoding(pd.Series([val]))
-                    for j, idx in enumerate(features2idx[feat]):
-                        fixed_features[idx] = transformed.values[0, j]
-
-                elif isinstance(feature, DiscreteInput):
+                # assert isinstance(feature, CategoricalInput)
```
bertiqwerty: remove uncommented code

bofire/utils/torch_tools.py
```diff
@@ -1093,3 +1101,218 @@ def create_supervised_dataset(
...
+        encodings = torch.from_numpy(
+            feature.to_descriptor_encoding(pd.Series(feature.categories)).values
+        ).to(**tkwargs)
+    else:
+        raise ValueError(
+            f"No categorical transformer found for feature with key: {feature.key} "
+            f"and transform: {transform}"
+        )
+    return Encoder(encodings)
+    # return lambda x: encodings[x]
```
bertiqwerty: remove uncommented code

tests/bofire/data_models/specs/surrogates.py
```diff
@@ -329,10 +339,103 @@
...
+                ard=True, features=["x_cont"]
+            ).model_dump(),
+        },
+        error=ValueError,
+        message="The features defined in the categorical",
+    ),
+)
+
+
+# "continuous_kernel": MaternKernel(
```
bertiqwerty: remove uncommented code

bofire/surrogates/random_forest.py
jduerholt: Done.

bofire/strategies/predictives/botorch.py
jduerholt: done

bofire/strategies/predictives/acqf_optimization.py
jduerholt: done

bofire/utils/torch_tools.py
jduerholt: done

tests/bofire/data_models/specs/surrogates.py
jduerholt: done

bofire/strategies/predictives/acqf_optimization.py
jduerholt: Done.

bofire/strategies/predictives/acqf_optimization.py
jduerholt: Done

bofire/utils/torch_tools.py
jduerholt: Done.

tests/bofire/strategies/test_acqf_optimization.py
jduerholt: `optimize_acqf_list` is able to deal with mixed domains via enumerating them.

tests/bofire/strategies/test_acqf_optimization.py
jduerholt: Done.

jduerholt:
Ok, everything is now settled, tests are failing due to a new pydantic version which was released yesterday. Here is the fix: https://github.com/experimental-design/bofire/pull/638

But we have to wait until we merge this for a new botorch version, as there is one bug in the current release version which will lead to trouble. It is fixed now in botorch, but they have to file a release, I pinged Max regarding this. Let us see what they say.

Issue #628, opened 2025-08-22
Sorting out experiments for botorch acqusition function: needed or bug?


LukasHebing:
I found, that in the `.ask()` function of the botorch optimizer, the acquision functions are generated by botorch. For this, experimental tensors `X_train` need to be provided.

In the function `get_acqf_input_tensors` from `/bofire/strategies/predictives/botorch.py` these are generated.

However, here experiments are filtered:
```python
# we should only provide those experiments to the acqf builder in which all
# input constraints are fulfilled, output constraints are handled directly
# in botorch
clean_experiments = clean_experiments[
            self.domain.is_fulfilled(clean_experiments)
        ].copy()
```

This leads to a bug in our application when users change the specifications and all previous experiments are sorted out. 
When we remove this filter, the BO algorithm runs fine

jduerholt:
Hi @LukasHebing,

i added this "feature" on purpose, as we had internally the following situation, there was a dataset with some historic experiment in which certain ingeredients were used which yield to very good performance of the material. But for some reasons, these ingredients should not be used anymore, so there bounds for the optimization were set to zero. If you would then include these experiments with the "illegal" ingredients into acqf setup for EI (also expected hypervolume), this can yield to severe problems, because your best solutions are out of reach for any future solution, and the optimizer is just coming up with super low acqf values (resulting in almost random candidates) because it could not beat the old but forbidden experiments.  For this reason, I added that only allowed experiments were used for acqf setup.

Can you elaborate a bit more which problems it creates on your side?

Maybe we need to introduce some flag to toggle the behavior on a per input/per constraint basis.

Best,

Johannes

LukasHebing:
Thanks @jduerholt . 
The basic situation is the same (changing the design space during the optimization, due to other business reasons).

It will then depend, which acquisition function we take. For EHVI, EI etc. I understand that this would be a problem. In our application, we majorly use UCI, where this wouldn't matter.

I will make a PR, in which I make this behavior optional, If that's ok for you..

jduerholt:
Sounds good to me! Best, Johannes

jduerholt:
Hmm, but I do not understand why it is at all a problem for UCB? Initializing UCB does not need the experiments. At botorch level, they are not passed to the acqf, I think ...

LukasHebing:
[...]

PR #630, opened 2025-08-25
Fix/628

LukasHebing:
This PR addresses #628.

It adds a field to the data-model of the `BotorchStrategy`: `include_infeasible_exps_in_acqf_calc`

- This flag changes the behavior of the `get_acqf_input_tensors` function: Filtering out infeasible experiments (violation of constraints or bounds) will only be done, if the field is set to False.

- The default value is False, so the function will work as before, if nothing is changed.

- All `make` methods from the strategy classes, inheriting from `BotorchStrategy` are adapted to show this option in the kwargs

### DIFF
--- a/bofire/data_models/strategies/predictives/botorch.py
+++ b/bofire/data_models/strategies/predictives/botorch.py
@@ -39,7 +39,7 @@ class BotorchStrategy(PredictiveStrategy):
     include_infeasible_exps_in_acqf_calc: bool = Field(
         default=False,
         description="Whether infeasible experiments should be included in the set "
-                    "of experiments used to compute the acquisition function.",
+        "of experiments used to compute the acquisition function.",
     )
 
     @model_validator(mode="after")
--- a/bofire/data_models/strategies/predictives/botorch.py
+++ b/bofire/data_models/strategies/predictives/botorch.py
@@ -36,6 +36,11 @@ class BotorchStrategy(PredictiveStrategy):
     # hyperopt params
     frequency_hyperopt: Annotated[int, Field(ge=0)] = 0  # 0 indicates no hyperopt
     folds: int = 5
+    include_infeasible_exps_in_acqf_calc: bool = Field(
+        default=False,
+        description="Whether infeasible experiments should be included in the set "
+                    "of experiments used to compute the acquisition function.",
+    )
 
     @model_validator(mode="after")
     def validate_domain_for_optimizer(self):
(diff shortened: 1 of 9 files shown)


bofire/data_models/strategies/predictives/botorch.py
```diff
@@ -36,6 +36,11 @@ class BotorchStrategy(PredictiveStrategy):
     # hyperopt params
     frequency_hyperopt: Annotated[int, Field(ge=0)] = 0  # 0 indicates no hyperopt
     folds: int = 5
+    include_infeasible_exps_in_acqf_calc: bool = Field(
```
bertiqwerty, line 39: This is not a hyperopt parameter but the hyperopt comment is somehow close.

bofire/strategies/predictives/botorch.py
```diff
@@ -214,6 +217,13 @@ def has_sufficient_experiments(
         return False
 
     def get_acqf_input_tensors(self):
+        """
+
+        Returns:
+            X_train (Tensor): Tensor of shape (n, d) with n training points and d input dimensions.
```
bertiqwerty, line 223: Thanks for adding the docstring. I would prefer to have the type information as type annotation instead of in the docstring. I.e.,
```python
def get_acqf_input_tensors(self) -> tuple[Tensor, Tensor | None]:
    """
    Returns:
            X_train: Tensor of shape (n, d) with n training points and d input dimensions
            X_pending: Tensor of shape (m, d) with m pending points
    """
```

bofire/strategies/predictives/botorch.py
```diff
@@ -224,16 +234,19 @@ def get_acqf_input_tensors(self):
...
-        # in botorch
-        clean_experiments = clean_experiments[
-            self.domain.is_fulfilled(clean_experiments)
-        ].copy()
-        if len(clean_experiments) == 0:
-            raise ValueError(
-                "No valid and feasible experiments are available for setting up the acquisition function. Check your constraints.",
-            )
+        if not self.include_infeasible_exps_in_acqf_calc:
+            # we should only provide those experiments to the acqf builder in which all
```
bertiqwerty, line 238: This comment is now somewhat outdated.

jduerholt:
Hi @LukasHebing,

I will have a look this evening, the latest tmr.

Best,

Johannes

jduerholt (approved):
Looks good to me!

PR #634, opened 2025-09-22
Active Learning Strategy: fix missing input processing for acqf samples.

jkeupp:
## Motivation

The active learning strategy crashes for mixed domains (some input features are categorical) when 

```
mc_samples = sampler.ask(candidate_count=self.acquisition_function.n_mc_samples)
# followed by 
torch.tensor(mc_samples)
``` 
gets categorical features instances as strings and complains.
While looking at the fix i think the input processing is missing in the active learning strategy. I think this was not caught during development as the utilized benchmark only has numerical features in 0-1 range. 

## Steps Taken

The missing transform was added. 

## Test Plan

Additional tests are now present. The original test passes, and the additional tests make sure the strategy works for mixed as well as purely categorical domains.  A slow test also tests for candidates actually having comparatively large uncertainties compared to average uncertainties.

### DIFF
--- a/tests/bofire/strategies/test_active_learning.py
+++ b/tests/bofire/strategies/test_active_learning.py
@@ -1,11 +1,19 @@
 import pandas as pd
+import pytest
 
 import bofire.strategies.api as strategies
 from bofire.benchmarks.multi import DTLZ2
 from bofire.data_models.acquisition_functions.api import qNegIntPosVar
-from bofire.data_models.domain.api import Outputs
+from bofire.data_models.domain.api import Domain, Inputs, Outputs
+from bofire.data_models.features.api import (
+    CategoricalInput,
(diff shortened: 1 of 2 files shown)


tests/bofire/strategies/test_active_learning.py
```diff
@@ -55,3 +67,416 @@ def test_active_learning():
...
+                    + (flow_diff / 4.9) ** 2
+                )
+                min_distance = min(min_distance, distance)
+
+        # Recommendations should have some distance from existing points
+        assert (
+            min_distance > 0.001
+        ), "Recommendations should explore new regions, not repeat existing experiments"
+
+    print(f"\n{'='*80}")
```
jduerholt: Can you remove the print statements?

Issue #643, opened 2025-10-28
[Bug]: Wrong __call__ method for categorical exclude constraint


LukasHebing:
### What happened?

The `__call__` method for constraints should return a numerical value ($<=0$, if fulfilled). 

However, the `CategoricalExcludeConstraint` method returns $1.0$ for fulfilled constraints, and $0.0$ for not fulfilled constraints. The descriptions seems to be wrong.

```python
def __call__(self, experiments: pd.DataFrame) -> pd.Series:
    """Numerically evaluates the constraint.

    Returns the distance to the constraint fulfillment. Here 1 for not fulfilled
    and 0 for fulfilled.

    Args:
        experiments (pd.DataFrame): Dataframe to evaluate the constraint on.

    Returns:
        pd.Series: Distance to reach constraint fulfillment.

    """
    return self.is_fulfilled(experiments).astype(float)
```

proposed fix:

```python
return 1 - self.is_fulfilled(experiments).astype(float)
```

### Please provide a minimal, reproducible example of the unexpected behavior.

import pandas as pd
import numpy as np
from itertools import product
#%%
from bofire.data_models.domain import api as domain_api
from bofire.data_models.features import api as features_api
from bofire.data_models.constraints import api as constraints_api
from bofire.data_models.objectives import api as objectives_api
#%% md
# Domain definition
#%%

exclude_constraint = constraints_api.CategoricalExcludeConstraint(
    features=["solvent", "temperature"],
    conditions=[
        constraints_api.SelectionCondition(selection=["ethanol"]),
        constraints_api.ThresholdCondition(threshold=70.0, operator=">="),
    ]
)

domain = domain_api.Domain(
    inputs=domain_api.Inputs(features=[
        features_api.ContinuousInput(key="temperature", bounds=(30.0, 100.0)),
        features_api.CategoricalInput(key="solvent", categories=["water", "ethanol"]),
    ]),
    outputs=domain_api.Outputs(features=[
        features_api.ContinuousOutput(key="yield", objective=objectives_api.MaximizeObjective()),
    ]),
    constraints=domain_api.Constraints(constraints=[
        exclude_constraint,
    ])
)
#%%
print(domain)
#%% md
## Generate candidates and check constraint fulfillment
#%%
candidates = pd.DataFrame(
    list(
        product(
            [30.0, 50.0, 70.0, 90.0],
            ["water", "ethanol"],
        )
    ),
    columns=["temperature", "solvent"],
)
#%%
[...]

PR #644, opened 2025-10-28
Feature/ Exclude constraints usable in GA

LukasHebing:
This PR makes the `CategoricalExcludeConstraint` available for solving with GA in mixed domains:

The constraint is evaluated in the objective function, using the `__call__` method. No gradients are required. 

The fix for #643 is included in this PR

### DIFF
--- a/bofire/data_models/constraints/categorical.py
+++ b/bofire/data_models/constraints/categorical.py
@@ -163,7 +163,7 @@ class CategoricalExcludeConstraint(Constraint):
             experiments (pd.DataFrame): Dataframe to evaluate the constraint on.
 
         Returns:
-            pd.Series: Distance to reach constraint fulfillment.
+            pd.Series: Constraint fulfillment: Will be 1.0 for not fulfilled and 0.0 for fulfilled.
 
         """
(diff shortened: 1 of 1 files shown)


bofire/data_models/constraints/categorical.py
```diff
@@ -166,7 +166,7 @@ def __call__(self, experiments: pd.DataFrame) -> pd.Series:
             pd.Series: Distance to reach constraint fulfillment.
 
         """
-        return self.is_fulfilled(experiments).astype(float)
+        return 1 - self.is_fulfilled(experiments).astype(float)
```
jduerholt, line 169: It is fine for me to change it.

bofire/data_models/constraints/categorical.py
```diff
@@ -166,7 +166,7 @@ def __call__(self, experiments: pd.DataFrame) -> pd.Series:
             pd.Series: Distance to reach constraint fulfillment.
```
jduerholt: Maybe we should mention here that it is not really a distance but just 0/1?

jduerholt: I mean in the docstring

PR #645, opened 2025-10-28
Conditional input features

TobyBoyne:
## Motivation

This PR introduces conditional features to BoFire. These are features that only impact the underlying surrogate if some other "indicator" feature is active.

When optimizing chemical processes, we often have some inputs that are dependent on others.
For example, the value of a `catalyst_concentration` input feature is only relevant depending
on another feature `use_catalyst==True`. Whilst it may seem that `use_catalyst==False` is
equivalent to just setting `catalyst_concentration==0`, there may be some limitations to 
this approach:
- If a catalyst is used, there may be some minimum amount required. It is difficult to model
 the disjoint bounds of a continuous feature.
- It may be the case that some tiny presence of catalyst enables a side reaction that completely
changes the reaction. We therefore have a step change at 0, with smoother behaviour everywhere
else in the domain, which GP surrogates cannot model well.

To solve these problems, we can use conditional kernels (eg. [1, 2]). These project a 1D input into a 2D space, which contains the original 1D interval as well as an "inactive" location, to which all points where the indicator is 0 are projected.

I have also included a jupyter notebook that sets up the example described above.

[1] Swersky et al. 2014, "Raiders of the Lost Architecture: Kernels for Bayesian Optimization in Conditional Parameter Spaces" arXiv 
[2] Horn et al. "Surrogates for hierarchical search spaces: the wedge-kernel and an automated analysis", GECCO

Yes

## Test Plan

I will test all of the new methods that I have created (once I have approval for the current design structure). I will ensure the data models work for a variety of different input spaces; check that errors are appropriately raised if the indicator features are missing; check that the integration with GPs works as expected; check some simple examples (such as those in the notebook) to confirm the kernel is well behaved.

### DIFF
--- a/bofire/kernels/conditional.py
+++ b/bofire/kernels/conditional.py
@@ -60,6 +60,9 @@ def build_indicator_func(
     conditions: Sequence[tuple[str, str, Condition]],
     features_to_idx_mapper: Callable[[list[str]], list[int]] | None,
 ) -> IndicatorFunction:
+    if not conditions:
+        return lambda X: torch.ones_like(X, dtype=torch.bool)
+
     if features_to_idx_mapper is None:
         raise RuntimeError(
             "features_to_idx_mapper must be defined when using a conditional kernel"
@@ -92,15 +95,21 @@ def compute_base_kernel_active_dims(
     appended but not
 
     This also removes any variables that were used as indicators."""
-    if features_to_idx_mapper is None:
-        raise ValueError(
-            "features_to_idx_mapper must be defined when using conditional kernels."
-        )
     if not active_dims:
         # since we append the embedded features on the right of the input features,
         # we need to know the number of features passed to the kernel
         raise ValueError(
             "`active_dims` must not be empty when using conditional kernels."
+        )
+    if not data_model.conditions:
+        # if there are no conditions, then no embedding takes place - simply return
+        # the active dimensions; this check prevents raising an error if there are
+        # no conditions, and `feature_to_idx_mapper` is None
+        return active_dims
+
+    if features_to_idx_mapper is None:
+        raise ValueError(
+            "features_to_idx_mapper must be defined when using conditional kernels."
         )
 
     base_kernel_data_model = data_model.base_kernel


bofire/data_models/constraints/constraint.py
```diff
@@ -12,7 +12,7 @@
 class Constraint(BaseModel):
     """Abstract base class to define constraints on the optimization space."""
 
-    type: Any
+    type: str
```
jduerholt: It has to by `Any`, in the abstact base class. Since it is always `Literal` in the derived class.

jduerholt:
Hey @TobyBoyne,

many thanks for this. This is definitely a very nice feature. I am still wrapping my head around how to do it best. As it also touches other related ideas that I have in my head. One question regarding the current setup:

> To me, that didn't really make sense since we don't actually need to enforce that, for example, `catalyst_concentration==0` if `use_catalyst==False`. Rather, we actually want to capture a semantic relationship between the inputs, which is why I decided to go with a new feature.

I think we have to seperate two things:

1. Model related stuff and specific kernels, to have the Wedge kernel there is no need for the new feature or? One could do this completely independ from it, or?
2. Disjoint bounds for continuous inputs: How do you actually want to perform the acqf optimization for disjoint bounds? Disjoint bounds would be really really great and is something which would really benefit the package. The most common disjoint bound is actually that the feature is either zero or within a certain range. I think with this, one could already solve a lot of usecases. What do you think about just adding a flag to `ContinuousInput` which reads `allow_zero: bool`? If one needs more in the future, one could transform the `bounds` attribute into a list of disjoint bounds.

Bottom line:

I really like what we want to achieve here, question for me is if we need a new feature for it or could just modify the Continuous one and then especially how to optimize over it? Of course one can enumerate it ...

bertiqwerty:
cc @R-M-Lee

jduerholt:
Hmm, I am not yet 100% convinced if it is the best to handle it via a new feature and coupling to a categorical, but I really would like to have it in BoFire. How about to first integrate just the kernel and then decide on how to do the rest? Or would has this implications?

TobyBoyne:
> auxilliary variables

I agree that for this problem, the auxilliary variable seems unnecessary. I think `optimize_acqf` would require the auxilliary variable if we, for example, wanted to formulate this as an alternating optimization so that we have a discrete parameter we can change. But in this example, it isn't necessary for the user to see/define it.

Maybe if the feature key is `"x"`, you can provide `"x"` as the `indicator_feature` field to indicator that it doesn't depend on anything else, and maybe that way you can represent disjoint bounds?

I also worry that just adding a new parameter to `ContinuousInput` obscures the fact that the kernel would change under the hood. It may be better to be explicit and require a `ConditionalContinuousInput` to use a `ConditionalKernel`?

I chose to add a new feature so that we can also support more complicated problem spaces, like those below. Is there a different approach to represent these feature spaces without introducing a new feature?

**Architecture optimization of neural networks**
Consider the domain below, where the conditions all depend on the same auxilliary:
```python
nn_domain = Domain.from_lists(
    inputs=[
        DiscreteInput(key="num_layers", values=[1, 2, 3]),
        DiscreteInput(key="num_nodes_1", bounds=(1, 64)),
        ConditionalDiscreteInput(key="num_nodes_2", bounds=(1, 64), indicator_feat="num_layers", indicator_condition=Threshold(threshold=2, operator=">=")),
        ConditionalDiscreteInput(key="num_nodes_3", bounds=(1, 64), indicator_feat="num_layers", indicator_condition=Threshold(threshold=3, operator=">=")),
    ], ... )
```

**Unstable catalyst at high temperatures**
Consider this reaction optimization, where we want to use a catalyst but it becomes unstable at high temperatures. This would be a use case for `drop_indicators_in_base_kernel=False`, since we want to include temperature in the surrogate.

Edited to add: in hindsight, maybe this is actually better represented by a constraint...

```python
rxn_domain = Domain.from_lists(
    inputs=[
        ContinuousInput(key="temperature", bounds=(0, 200)),
        ConditionalContinuousInput(key="solvent_concentration", bounds=(0, 1), indicator_feat="temperature", indicator_condition=Threshold(threshold=100, operator="<"))
    ], ...)
```

TobyBoyne:
Sounds good! I will start working on the optimization next week :) For now, all conditionality must be encoded in the `ConditionalKernel`

jduerholt:
Good point, for this I would like to include the `allow_zero` into the ContinuousInput Feature. We defnititely need it!

jduerholt:
Hi @TobyBoyne,

sorry for my late response, I have just a crazy week. 

`AcceptingFeatures`: I would go for your propal of only allowing it for the base kernel. 

`RandomStrategy`: Let us do this in a seperate PR in which we also clarify how we do it in acqf optimization. This is also very much related to the handling of NChooseKs, as we turn features of. Do you have time this week or next one on a discussion via MS Teams on how to do this in the best way?

Tell me when you are ready with the PR, then I will review. 

Best,

Johannes

bofire/data_models/features/continuous.py
```diff
@@ -19,6 +19,9 @@ class ContinuousInput(NumericalInput):
         stepsize (PositiveFloat, optional): Float indicating the allowed stepsize between lower and upper. Defaults to None.
         local_relative_bounds (Tuple[float, float], optional): A tuple that stores the lower and upper bounds relative to a reference value.
             Defaults to None.
+        allow_zero (bool): A boolean indicating if the input feature can take inactive values. If True,
+            use a conditional encoding kernel for this feature. Useful for features that take values between `bounds`, but can
```
jduerholt: I would write that one can use a conditional kernel for this, but that it is not mandatory.

bofire/data_models/kernels/conditional.py
```diff
@@ -0,0 +1,53 @@
...
+        # ScaleKernel,
+    ]
+
+    conditions: Sequence[tuple[str, str, Condition]]
+
+
+class WedgeKernel(ConditionalEmbeddingKernel):
+    type: Literal["WedgeKernel"] = "WedgeKernel"
+    ard: bool = True
+    lengthscale_prior: Optional[AnyPrior] = None
```
jduerholt, line 55: Do we have priors for this hyperparams that make sense?

jduerholt: And why do we have it here? Is the lenghtscale here including the indicator feature? And for this reason, we always have a lenghtscale and also need to specify it here?

bofire/data_models/kernels/conditional.py
```diff
@@ -0,0 +1,53 @@
...
+        RBFKernel,
+        MaternKernel,
+        LinearKernel,
+        HammingDistanceKernel,
+        # AdditiveKernel,
+        # MultiplicativeKernel,
+        # ScaleKernel,
+    ]
+
+    conditions: Sequence[tuple[str, str, Condition]]
```
jduerholt, line 49: Can you provide an example here for a condition, not yet 100% clear for me what will be specified here.

bofire/kernels/conditional.py
```diff
@@ -0,0 +1,280 @@
...
+    ThresholdCondition,
+)
+from bofire.data_models.kernels.conditional import (
+    ConditionalEmbeddingKernel as ConditionalEmbeddingKernelDataModel,
+)
+
+
+IndicatorFunction = Callable[[Tensor], Tensor]
+
+_threshold_operators: dict[str, Callable] = {
```
jduerholt: Can we not import this from constraints/conditions?

bofire/kernels/conditional.py
```diff
@@ -0,0 +1,280 @@
...
+        return self.raw_radius_constraint.transform(self.raw_radius)  # type: ignore
+
+    @radius.setter
+    def radius(self, value):
+        if not isinstance(value, Tensor):
+            value = torch.as_tensor(value).to(self.raw_radius)  # type: ignore
+        self.initialize(raw_radius=self.raw_radius_constraint.inverse_transform(value))  # type: ignore
+
+    def embedding(self, x):
+        # this assumes that x has been normalized to [0, 1]
```
jduerholt, line 238: Should we check that it is in the range between zero and one and raise a warning if not the case?

bofire/data_models/features/continuous.py
jduerholt, line 1: `get_bounds` of the ContinuousInput feature is used to setup up the bounds for the botorch input transform that scales the input between zero and one. We have to get the allow_zero in there: https://github.com/experimental-design/bofire/blob/779e4f557b04893859d004f3741b66decc4a48b0/bofire/surrogates/utils.py#L150

bofire/strategies/predictives/acqf_optimization.py
```diff
@@ -624,7 +629,8 @@ def get_categorical_combinations(
         fixed_basis = self.get_fixed_features(domain=domain)
 
         combos = domain.inputs.get_categorical_combinations(
-            include=[CategoricalInput, DiscreteInput],
+            # TODO: is commented line below necessary? If not, remove
```
jduerholt: Should be fine.

tests/bofire/data_models/features/test_continuous.py
```diff
@@ -186,6 +190,10 @@ def test_continuous_input_feature_validate_candidental_valid(input_feature, valu
             specs.features.valid(ContinuousInput).obj(bounds=(3, 3)),
             pd.Series([3.1, 3.2, 3.4]),
         ),
+        (
+            specs.features.valid(ContinuousInput).obj(allow_zero=False),
```
jduerholt: I am not sure what are the bounds of the default ContinuousInput of the test suite, can you define it explicitly? The ContinuosInput Feature here.

bofire/kernels/conditional.py
TobyBoyne: There were some issues with circular imports that I quickly resolved by copy/pasting this here. I can move the dictionary to `conditions.py` which should help avoid this issue.

bofire/kernels/conditional.py
TobyBoyne: We can do! This raises the issue of false negatives, where this check makes us implicitly think that everything is fine if, for example, the input range is [0, 0.001]. However happy to add this check as an extra backstop.

bofire/data_models/features/continuous.py
TobyBoyne: It's not clear to me the correct way to scale in this case. If you imagine a feature that has `bounds==[100, 105]`, with `allow_zero==True`, I think you should scale according to `bounds`. Otherwise, you will squash all of the input space to a very small part of the [0, 1] interval, and most of the interval will not be reachable. Would likely lead to poor performance with the lengthscale priors that we use.

If we don't include zero in the scaling bounds, then the point at zero will likely always be very far away from all of the points in the [100, 105] interval which leads to entirely different issues. We don't have this problem for conditional kernels, but I'm not sure the best approach for regular continuous kernels.

bofire/data_models/kernels/conditional.py
TobyBoyne: The kernel embeds the space according to the image below. "Active points" (with $\delta=1$) are not moved - they stay at their original position on the numberline. "Inactive points" are mapped to some point above the line.

image

> Priors for hyperparams

The prior on the lengthscale, $\ell$, should be exactly the same as the prior used for other conditional kernels, ie. the Hvarfner prior. The reason why it needs to be in the conditional kernel and not the base kernel is that we want to only apply the lengthscale along the x direction. The height of the inactive point shouldn't depend on the lengthscale, since this may interfere with the model fitting when there is some entangling of the kernel parameters.

I haven't seen much discussion on the prior on the angle, $\theta$. Anything symmetric about 0.5 makes sense. This parameter encodes "which extreme value is the inactive point more similar to?". I think leaving it as uniform should generally be fine. If the inactive point represents 0, then your prior may be that it is closest to the end of the active interval that is closest to zero, but I don't think that prior is necessary.

I similarly haven't seen much discussion on the prior on the radius, $\rho$. This parameter encodes "how similar is the inactive point to all of the active points?". I think another Hvarfner lengthscale prior could be a good default here?

Also as a note: the literature doesn't really discuss these priors at all, and the priors have been able to find in implementations (eg. of the ArcKernel in gpytorch) still use gamma priors which aren't really used since the Hvarfner paper. You'll also notice that the `rotated_embedding` is different from the `embedding` (the latter more closely reflects the Wedge Kernel paper), since I think the rotated embedding enables more meaningful priors on parameters.

TobyBoyne: > Why do we have it here?

The lengthscale is exactly the same as the lengthscale for any other continuous kernel. This kernel always has a lengthscale, in the same way RBFKernel always has a lengthscale.

The lengthscale shouldn't be a part of the base kernel, for the reason noted above: we only want the lengthscale to transform the x-direction of the embedding (by which I mean the horizontal direction of the image above). For this reason, in `WedgeKernel.__init__`, we fix `base_kernel.lengthscale` to `1.0`.

It does also impact the indicator feature, although this doesn't actually matter if that feature is dropped by the base kernel.

TobyBoyne:
Hi @jduerholt,

I have made all the changes you've requested (except for those that I have replied to, awaiting your thoughts).

> Acqf optimization

Yes, this currently only works with `optimize_acqf_mixed`. If we want this to work with `optimize_acqf_mixed_alternating`, then we would need to have an indicator feature. There is no way with the current implementation of `alternating` to have a feature that is both discrete and continuous, but with an indicator that is discrete this will work.

Note that we can always do this transformation under-the-hood - a user defines one feature with `allow_zero=True`, and we append an extra indicator dimension if using alternating optimization. But this could get a bit messy, especially when deciding where to do this transformation.

> NChooseK

Am I correct in saying that NChooseK constraints in BoTorch are just implemented as smooth nonlinear constraints? If so, I think this feature should still work well with them! If you give me a very rough example of a problem `Domain` that has both conditional features and NChooseK constraints, I will test it out and see if it works.

bofire/data_models/kernels/conditional.py
jduerholt: Many thanks for your explanations!

bofire/kernels/conditional.py
jduerholt: Not really necessary. I was just wondering!

bofire/kernels/conditional.py
jduerholt: I think it is also not super important, do it as you prefer!

bofire/data_models/features/continuous.py
jduerholt: We let it as is and will also remove the explicit setting of bounds and rely of learning them from data, see this discussion: https://github.com/meta-pytorch/botorch/discussions/3077#discussioncomment-14925485

jduerholt:
Hey @TobyBoyne, looks good to me! But one test with respect to your implementation is faling:

FAILED tests/bofire/kernels/test_mapper.py::test_map[kernel_spec12] - RuntimeError: features_to_idx_mapper must be defined when using a conditional kernel

Can you have a look what is going on? Then we could merge it in.

PR #668, opened 2025-11-17
Feature/isolate repair

LukasHebing:
## Motivation

The current repair function from the GA uses QP-optimization to find feasible input space points (considering linear and N-choose-k constraints).
This would be beneficial to use also in other areas. This PR isolates the linear projection (QP-optimization) from the GA code. 

## Test Plan

Mostly, the old tests can be used.


LukasHebing:
Remaining to-do's

- Solve the QP with normalized matrices
- Some more documentation

Jimbo994:
Hi Lukas,

I cannot see the images unfortunately. But good that you can reproduce and that scaling makes it more stable.

I was just checking the CVXPY and OSQP documentation again on a big screen and learned the following:
- For QP problems, CVXPY will default to the OSQP solver.
- The OSQP solver has this eps_abs, and eps_rel we discussed, and the termination of the solver is defined here: https://osqp.org/docs/solver/index.html#convergence. Where you can see that eps_abs, and eps_rel both determine the convergence criteria of eps_prim and eps_dual. The defaults when I was running the repair function with verbose were 1e-5 and 1e-5 for eps_abs and eps_rel.
- In BoFire we then validate candidates here: https://github.com/experimental-design/bofire/blob/e719dd68cc56eeaaead45c922bcab40eec4a4bbe/bofire/data_models/domain/domain.py#L418, which in the end checks the constraints using a `np.isclose` like over here with a tolerance of 1e-5: https://github.com/experimental-design/bofire/blob/e719dd68cc56eeaaead45c922bcab40eec4a4bbe/bofire/data_models/constraints/constraint.py#L80
- This then causes issues for us when the eps_rel term pushes the eps_prim and eps_dual over 1e-5. 

So I think scaling will help, but still does not guarantee in all cases that no constraint violation can occur with our current settings, right? However, I checked, and we are allowed to set eps_rel to 0. So we could also set something like:
`problem.solve(verbose=self.verbose, solver=cp.OSQP, eps_abs=1e-5, eps_rel=0)` 

What do you think?

bofire/utils/domain_repair.py
```diff
@@ -0,0 +1,471 @@
...
+            "h": h,
+            "A": A,
+            "b": b,
+            "initvals": x,
+        }
+
+    def solve_numeric(self, X: np.ndarray) -> np.ndarray:
+        """
+        Converting a set of experiments in a domain, into the feasible space, considering the inputs in the numerical
+        space. inputs must be in the right orde (according to the order in domain.inputs.get_keys())
```
Jimbo994: orde=order.

Jimbo994:
Nice, just went over the code, very cool.

Jimbo994 (approved):

jduerholt:
> @LukasHebing Should we merge it in?

An updates here?

LukasHebing:
> > @LukasHebing Should we merge it in?
> 
> An updates here?
Sorry, didn't see your comment.

The failing examples issue was solved (?) https://github.com/experimental-design/bofire/discussions/670#discussioncomment-15025578

When I got time, I'll merge and let the tests run again

Issue #663, opened 2025-11-10
[FEATURE REQUEST]: Make method for surrogates


jduerholt:
### Motivation

We want to have a `Surrogate.make` method in analogy to the `Strategy.make` method to be able to instantiate directly from the functional model and do not have the need for a mapping of the surrogate data model to the functional model.

### Describe the solution you'd like to see implemented in BoFire.

see above.

### Pull Request

None

PR #671, opened 2025-11-19
Make method for surrogates

rlars:
## Motivation

This PR creates make methods for surrogates, see #663 

Yes.

## Test Plan

For now, created a test method analogous to the one for make methods for strategies.

### DIFF
--- a/bofire/data_models/surrogates/fully_bayesian.py
+++ b/bofire/data_models/surrogates/fully_bayesian.py
@@ -1,4 +1,4 @@
-from typing import Annotated, List, Literal, Type
+from typing import Annotated, List, Literal, Optional, Type
 
 from pydantic import AfterValidator, Field, field_validator, model_validator
 
@@ -8,7 +8,7 @@ from bofire.data_models.surrogates.trainable_botorch import TrainableBotorchSurr
 from bofire.data_models.types import make_unique_validator
 
 
-class FullyBayesianSingleTaskGPSurrogate(TrainableBotorchSurrogate[Hyperconfig]):
+class FullyBayesianSingleTaskGPSurrogate(TrainableBotorchSurrogate):
     type: Literal["FullyBayesianSingleTaskGPSurrogate"] = (
         "FullyBayesianSingleTaskGPSurrogate"
     )
@@ -19,6 +19,12 @@ class FullyBayesianSingleTaskGPSurrogate(TrainableBotorchSurrogate[Hyperconfig])
     features_to_warp: Annotated[
         List[str], AfterValidator(make_unique_validator("Features"))
     ] = []
+
+    hyperconfig: Optional[Hyperconfig] = None
+
+    @property
+    def hyperconfig_access(self) -> Optional[Hyperconfig]:
+        return self.hyperconfig
 
     @model_validator(mode="after")
     def validate_features_to_warp(self):
--- a/bofire/data_models/surrogates/mixed_single_task_gp.py
+++ b/bofire/data_models/surrogates/mixed_single_task_gp.py
@@ -55,9 +55,7 @@ class MixedSingleTaskGPHyperconfig(Hyperconfig):
     ] = "FractionalFactorialStrategy"
 
 
-class MixedSingleTaskGPSurrogate(
-    TrainableBotorchSurrogate[MixedSingleTaskGPHyperconfig]
-):
+class MixedSingleTaskGPSurrogate(TrainableBotorchSurrogate):
     type: Literal["MixedSingleTaskGPSurrogate"] = "MixedSingleTaskGPSurrogate"
     continuous_kernel: AnyContinuousKernel = Field(
         default_factory=lambda: RBFKernel(
@@ -75,6 +73,10 @@ class MixedSingleTaskGPSurrogate(
     hyperconfig: Optional[MixedSingleTaskGPHyperconfig] = Field(
         default_factory=lambda: MixedSingleTaskGPHyperconfig(),
     )
+
+    @property
+    def hyperconfig_access(self) -> Optional[Hyperconfig]:
+        return self.hyperconfig
 
     @classmethod
     def _default_categorical_encodings(
@@ -145,8 +147,7 @@ class MixedSingleTaskGPSurrogate(
             bool: True if the output type is valid for the surrogate chosen, False otherwise
         """
         return isinstance(my_type, type(ContinuousOutput))
-    
-    
+
     def update_hyperparameters(
         self,
         hyperparameters: pd.Series,
@@ -194,4 +195,3 @@ class MixedSingleTaskGPSurrogate(
 
         else:
             raise ValueError(f"Kernel {hyperparameters.kernel} not known.")
-
--- a/bofire/data_models/surrogates/tanimoto_gp.py
+++ b/bofire/data_models/surrogates/tanimoto_gp.py
@@ -1,4 +1,4 @@
-from typing import Literal, Type
+from typing import Literal, Optional, Type
 
 from pydantic import Field, model_validator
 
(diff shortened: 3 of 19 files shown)


bofire/data_models/surrogates/trainable_botorch.py
```diff
@@ -1,11 +1,16 @@
...
 from bofire.data_models.surrogates.scaler import ScalerEnum
-from bofire.data_models.surrogates.trainable import TrainableSurrogate
+from bofire.data_models.surrogates.trainable import Hyperconfig, TrainableSurrogate
+
+
+T = TypeVar("T", bound=Hyperconfig)
 
 
-class TrainableBotorchSurrogate(BotorchSurrogate, TrainableSurrogate):
+class TrainableBotorchSurrogate(BotorchSurrogate, TrainableSurrogate[T], Generic[T]):
```
jduerholt: For what is this needed?

bofire/data_models/surrogates/tanimoto_gp.py
```diff
@@ -16,10 +16,11 @@
     AnyPrior,
 )
 from bofire.data_models.surrogates.scaler import ScalerEnum
+from bofire.data_models.surrogates.single_task_gp import SingleTaskGPHyperconfig
 from bofire.data_models.surrogates.trainable_botorch import TrainableBotorchSurrogate
 
 
-class TanimotoGPSurrogate(TrainableBotorchSurrogate):
+class TanimotoGPSurrogate(TrainableBotorchSurrogate[SingleTaskGPHyperconfig]):
```
jduerholt:
```suggestion
class TanimotoGPSurrogate(TrainableBotorchSurrogate[Hyperconfig]):
```

This should have an empty hyperconfig.

tests/bofire/surrogates/test_make_surrogate.py
```diff
@@ -0,0 +1,124 @@
...
+                    made_surrogate, k
+                ), f"{type(surrogate_type).__name__}. {k} missing in made_surrogate"
+
+    for spec in surrogate_specs.valids:
+        data_model = spec.obj()
+        surrogate_type = surrogates.map(data_model)
+        test(surrogate_type, data_model)
+
+
+if __name__ == "__main__":
```
jduerholt: This should be removed, one can always execute a single test file by just calling it with `pytest path/to/file`.

tests/bofire/surrogates/test_make_surrogate.py
```diff
@@ -0,0 +1,124 @@
...
+
+def test_make():
+    class SurrogateWithMake(Protocol):
+        @classmethod
+        def make(cls, *args, **kwargs) -> Surrogate: ...
+
+    def test(surrogate_type: type[SurrogateWithMake], data_model):
+        if (
+            type(data_model) in surrogates_skip_all
+            or type(surrogate_type) in surrogates_skip_all
```
jduerholt, line 53: Why this skip things?

bofire/data_models/surrogates/trainable_botorch.py
rlars: That fixes two things:
- The VS code warning complaining about different types for the hyperconfig in base class and child classes.
- The test_make function failing because it sees that types are different (i. e. SingleTaskGPHyperconfig vs Hyperconfig).

However, there may be another way to fix this: it is uncommon for data models / dtos to have such a hierarchy, and as hyperconfig seems to be not accessed on the base, a more straightforward solution would be to specify each data model explicitly, without inherited properties.
@jduerholt What do you think?

bofire/data_models/surrogates/tanimoto_gp.py
rlars: Good catch. This supports the argument to have data models mostly separate without hierarchical properties. Otherwise, the inheritance would get even more complicated.

tests/bofire/surrogates/test_make_surrogate.py
rlars: Seems I missed to implement make for some surrogates (which at first glance I thought to be used internally only). And there are some more typing issues for certain classes. I will try to fix them.

bofire/data_models/surrogates/trainable_botorch.py
rlars: I have seen that there is access to hyperconfig in _update_hyperparameters.
So, I added another generic, which now makes things more coherent but also more complicated...

jduerholt:
Hi @rlars,

just tell me when you are ready, then I rereview.

Best,

Johannes

bofire/data_models/surrogates/trainable_botorch.py
rlars: Reworked this completely by adding abstract properties to the base so we do not need generics.

tests/bofire/surrogates/test_make_surrogate.py
rlars: This is a shortcoming of mapping multiple surrogate data models to the same surrogate. Eventually, the test function could be adapted, but this should go along with improvements to the modelling.

bofire/data_models/surrogates/fully_bayesian.py
```diff
@@ -19,6 +20,12 @@ class FullyBayesianSingleTaskGPSurrogate(TrainableBotorchSurrogate):
         List[str], AfterValidator(make_unique_validator("Features"))
     ] = []
 
+    hyperconfig: Optional[Hyperconfig] = None
+
+    @property
```
jduerholt, line 25: Why are we setting this not over inheritance?

bofire/data_models/surrogates/mixed_single_task_gp.py
```diff
@@ -190,3 +147,51 @@ def is_output_implemented(cls, my_type: Type[AnyOutput]) -> bool:
             bool: True if the output type is valid for the surrogate chosen, False otherwise
         """
         return isinstance(my_type, type(ContinuousOutput))
+
+    def update_hyperparameters(
```
jduerholt, line 151: Hmm, why are you moving it outside of the Hyperconfig? Is this necessary?

bofire/data_models/surrogates/trainable.py
```diff
@@ -85,15 +86,20 @@ def domain(self) -> Domain:
             ),
         )
 
-    @staticmethod
-    def _update_hyperparameters(surrogate_data, hyperparameters: pd.Series):
-        raise NotImplementedError(
-            "Ideally this would be an abstract method, but this causes problems in pydantic.",
+    def update_hyperparameters(self, hyperparameters: pd.Series):
```
jduerholt, line 89: Hmm, now I am puzzled, the update is now part of both the config and the actual model?

bofire/data_models/surrogates/fully_bayesian.py
rlars: With this, TrainableSurrogate only specifies that any (non-abstract) implementation provides access to the generic hyperconfig. Therefore, TrainableSurrogate does not need to know what specific subtype of hyperconfig a surrogate implementation uses which gets rid of the generics.
hyperconfig_access is used in runners/trainable.py and other places.

bofire/data_models/surrogates/mixed_single_task_gp.py
rlars: I think this way the relationship surrogate <-> hyperconfig becomes clearer. Before, there was an inversion of responsibility where the hyperconfig would modify its surrogate in a manner which I did not anticipate.

bofire/data_models/surrogates/trainable.py
rlars: Until now there were somehow reversed responsibilities, where the config would mess with the model. Now, this method (of the Hyperconfig) could likely be renamed as it is doing the validation (the code was copied multiple times before and is now at a single place satisfying DRY).

PR #672, opened 2025-11-20
Tests/reduce disk space

bertiqwerty:
Our tests in the pipeline run out of disc space. Check what packages are not necessary. Torch with CUDA is a candiate to be replaced by CPU-Torch.

Further, I cleaned a little up by removing uv from the workflow which is not really needed there and by removing the xgb option from workflow installations that does not exist anymore.

Y

## Test Plan

not applicable

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -29,13 +29,8 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - name: Set up uv
-        uses: astral-sh/setup-uv@v2
-        with:
-          enable-cache: true
-
       - name: Install Dependencies
-        run: uv pip install "." pytest --system
+        run: pip install "." pytest
 
       - name: Run data model tests
         run: pytest tests/bofire/data_models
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -29,13 +29,8 @@ jobs:
...
           python-version: ${{ matrix.python-version }}
 
-      - name: Set up uv
-        uses: astral-sh/setup-uv@v2
-        with:
-          enable-cache: true
-
       - name: Install Dependencies
-        run: uv pip install "." pytest --system
+        run: pip install "." pytest
```
jduerholt, line 33: Why did you remove uv, here?

bertiqwerty: It is just pip installing. We don't need it. Maybe, it is a little more efficient in dependency resolution, but this way is simpler.

bertiqwerty: We can also discuss uv in a separate PR and re-introduce it more consistently. Because currently sometimes it is used and sometimes not.

PR #678, opened 2025-11-28
Feature/aggregations2

jduerholt:
## Motivation

This PR adds the possibility to perform custom feature aggregations in surrogates and use them for optimization.

An example application could be ContinuousDescriptorFeature or ContrinuousMolecularFeatures and the featurization of mixtures as discussed in Toby's recent paper: https://arxiv.org/pdf/2506.07619 

Y.

## Test Plan

Unit tests.

### DIFF
--- a/docs/tutorials/benchmarks/012-engineered_features.qmd
+++ b/docs/tutorials/benchmarks/012-engineered_features.qmd
@@ -1,0 +1,176 @@
+---
+title: Optimization with `ContinuousDescriptorInput`s and engineered Features
+jupyter: python3
+---
+
+
+Continuous descriptor inputs can be used in combination with engineered features to incorporate more domain knowledge into an optimization. Example could be a solvent of formulation optimization where per continuous input feature additional information about its properties is available.
+
+For examples in the literature have a look at this [paper](https://arxiv.org/abs/2506.07619).
+
+
+## Imports
+
+```{python}
+import pandas as pd
+import os
+import matplotlib.pyplot as plt
+import numpy as np
+
+from bofire.data_models.features.api import ContinuousInput, ContinuousDescriptorInput, SumFeature, WeightedSumFeature, ContinuousOutput
+from bofire.data_models.domain.api import Inputs, EngineeredFeatures, Outputs, Domain, Constraints
+from bofire.benchmarks.api import Himmelblau, FormulationWrapper
+from bofire.runners.api import run
+from bofire.strategies.api import RandomStrategy, SoboStrategy
+from bofire.data_models.surrogates.api import SingleTaskGPSurrogate
+import bofire.surrogates.api as surrogates
+from bofire.data_models.surrogates.api import BotorchSurrogates
+from typing import Optional
+
+SMOKE_TEST = os.environ.get("SMOKE_TEST")
+```
+
+## Problem Setup
+
+For demo purposes, the formulation wrapper is used around the two dimensional `Himmelblau` Benchmark to wrap it as a formulation problem with 7 features in total. Per original feature, three new features are introduced and one filler feature. The sum of each feature group is used as (normalized) input for the evaluation of the original `Himmelblau` function. A seventh feature is introduced as filler to always reach a total of 1 to fulfill the formulation/equality constraint. For more information on this setup have a look at the docstring of `FormulationWrapper`.
+
+```{python}
+bench = FormulationWrapper(benchmark=Himmelblau(), n_features_per_original_feature=3, n_filler_features=1)
+```
+
+The continuos descriptor features have the following descriptor values for the descriptors `x_1` and `x_2`:
+
+```{python}
+pd.concat([feat.to_df() for feat in bench.domain.inputs.get(ContinuousDescriptorInput)])
+```
+
+## Benchmarks
+
+### Random Strategy
+
+As baseline, we run the random strategy on the problem.
+
+```{python}
+def sample(domain):
+    strategy = RandomStrategy.make(domain=domain)
+    return strategy.ask(10)
+
+
+def best(domain: Domain, experiments: pd.DataFrame) -> float:
+    return experiments.y.min()
+
+random_results = []
+for _ in range(15 if not SMOKE_TEST else 1):
+    results = run(
+        benchmark=bench,
+        strategy_factory=lambda domain: RandomStrategy.make(domain=domain),
+        n_iterations=40 if not SMOKE_TEST else 2,
+        metric=best,
+        initial_sampler=sample,
+        n_runs=1,
+        n_procs=1,
+    )
+    random_results.append(results[0][0])
+```
+
+### Sobo Strategy
+
+Next we run the `SoboStrategy` on the problem.
+
+```{python}
+def sample(domain):
+    return initial_experiments[domain.inputs.get_keys()].copy()
+
+def best(domain: Domain, experiments: pd.DataFrame) -> float:
+    return experiments.y.min()
+
+sobo_results = []
+for i in range(15 if not SMOKE_TEST else 1):
+    initial_experiments = random_results[i][:10].copy()
+    results = run(
+        benchmark=bench,
+        strategy_factory=lambda domain: SoboStrategy.make(domain=domain),
+        n_iterations=40 if not SMOKE_TEST else 2,
+        metric=best,
+        initial_sampler=sample,
+        n_runs=1,
+        n_procs=1,
+    )
+    sobo_results.append(results[0][0])
+```
+
+Now we run the `SoboStrategy` again, but with an engineered feature as additional input. The `WeightedSumFeature` computes the sum over the specified descriptors weighted by the values of the involved original features.
+
+```{python}
+def create_strategy(domain):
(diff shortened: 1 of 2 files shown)


jduerholt:
Note to myself: fix it for the multitask model and then also fit the standardization ...

bofire/data_models/domain/features.py
```diff
@@ -241,6 +252,40 @@ class Features(_BaseFeatures[AnyFeature]):
...
 
+class EngineeredFeatures(_BaseFeatures[AnyEngineeredFeature]):
+    type: Literal["EngineeredFeatures"] = "EngineeredFeatures"  # type: ignore
+
+    def get_features2idx(self, offset: int = 0):
+        features2idx = {}
+        counter = offset
+        for feat in self.get():
+            features2idx[feat.key] = tuple(
+                (np.array(range(feat.n_outputs)) + counter).tolist(),
```
TobyBoyne: I think this could be neater as `tuple(out_idx + counter for out_idx in range(feat.n_outputs))`

bofire/data_models/features/engineered_feature.py
```diff
@@ -0,0 +1,74 @@
...
+from bofire.data_models.features.api import ContinuousDescriptorInput, ContinuousInput
+from bofire.data_models.features.feature import Feature
+from bofire.data_models.types import FeatureKeys
+
+
+class EngineeredFeature(Feature):
+    features: FeatureKeys
+    keep_features: bool = True
+
+    def validate_features(self, inputs: "Inputs"):  # noqa: F821 # type: ignore
```
TobyBoyne: Can you not import `Inputs` wrapped in a `if TYPE_CHECKING` block to avoid needing these comments?

bofire/data_models/features/engineered_feature.py
```diff
@@ -0,0 +1,74 @@
...
+                f"The following features are missing in inputs: {missing_features}"
+            )
+        self._validate_features(inputs)
+
+    def _validate_features(self, inputs: "Inputs"):  # noqa: F821 # type: ignore
+        pass
+
+    @property
+    @abstractmethod
+    def n_outputs(self) -> int:
```
TobyBoyne: I'm not sure I like `n_outputs` as a name - these aren't really outputs, just transformed inputs, right?

bofire/data_models/features/engineered_feature.py
```diff
@@ -0,0 +1,74 @@
...
+    order_id: ClassVar[int] = 1
+
+    @property
+    def n_outputs(self) -> int:
+        return 1
+
+
+class WeightedSumFeature(EngineeredFeature):
+    type: Literal["WeightedSumFeature"] = "WeightedSumFeature"
+    descriptors: Annotated[List[str], Field(min_length=2)]
```
TobyBoyne: Can you use `data_models.types.Descriptors` here?

bofire/data_models/features/engineered_feature.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def validate_features(self, inputs: "Inputs"):  # noqa: F821 # type: ignore
+        super().validate_features(inputs)
+        for feature_key in self.features:
+            feature = inputs.get_by_key(feature_key)
+            if not isinstance(feature, ContinuousDescriptorInput):
+                raise ValueError(
+                    f"Feature '{feature_key}' is not a ContinuousDescriptorInput",
+                )
+            if len(set(self.descriptors) - set(feature.descriptors)) > 0:  # type: ignore
```
TobyBoyne: Should this not be `set(self.descriptors) == set(feature.descriptors)`? Is it okay if there are some descriptors present in the EngineeredFeature, which are not present in the Feature?

bofire/surrogates/engineered_features.py
```diff
@@ -0,0 +1,101 @@
...
+from bofire.data_models.features.api import (
+    EngineeredFeature,
+    MeanFeature,
+    SumFeature,
+    WeightedSumFeature,
+)
+from bofire.data_models.types import InputTransformSpecs
+
+
+def map_sum_feature(
```
TobyBoyne, line 14: Could these three mapping functions be rewritten as a single function? The only thing that changes between them is the definition of `f`.

bofire/surrogates/utils.py
```diff
@@ -107,86 +108,145 @@ def get_categorical_feature_keys(
...
-                tf1=categorical_transform, tf2=scaler
-            )
-        elif categorical_transform is not None:
-            input_transform = categorical_transform
-        elif scaler is not None:
-            input_transform = scaler
-    else:
-        input_transform = scaler
-    return input_transform
+        if feature.keep_features is False:
```
TobyBoyne: `if not feature.keep_features`

tests/bofire/surrogates/test_engineered_features.py
```diff
@@ -0,0 +1,116 @@
...
+        inputs=inputs, transform_specs={}, feature=aggregation
+    )
+
+    orig = torch.tensor([[0.1, 0.2, 0.7], [0.4, 0.1, 0.5]]).to(**tkwargs)
+    result = aggregator(orig)
+
+    assert result.shape[0] == 2
+    assert result.shape[1] == 5
+
+    assert torch.allclose(result[:, :-2], orig)
```
TobyBoyne: Please add the expected return values

TobyBoyne (commented):
Looks good, most of my comments are pretty minor nitpicks.

Since you mentioned the Catechol paper, I'm not sure this PR has enough features to replicate our model. We have two categorical inputs (solvent A and solvent B) with descriptors, then used a weighted sum of those two solvents. However, it looks to me like this PR only supports ContinuousDescriptors, which have fixed descriptors. Have I understood this correctly, and will you be adding this feature?

bofire/data_models/features/engineered_feature.py
jduerholt: Hmm, problem are here that I get circular imports ..., do you think this would help?

bofire/data_models/domain/features.py
jduerholt: I agree!

bofire/data_models/features/engineered_feature.py
jduerholt: It solved it, thanks for the hint!

bofire/data_models/features/engineered_feature.py
jduerholt: good point, I renamed it to `n_transformed_inputs` .

bofire/data_models/features/engineered_feature.py
jduerholt: Good catch!

bofire/surrogates/engineered_features.py
jduerholt: The `map_weighted_sum_feature` is different, for this reason, I would keep it as is.

tests/bofire/surrogates/test_engineered_features.py
jduerholt: Done.

bofire/surrogates/utils.py
jduerholt: Done.

bofire/data_models/features/engineered_feature.py
jduerholt: Done.

bofire/data_models/features/engineered_feature.py
jduerholt: Done.

bertiqwerty:
I had a first look at your changes. I think, I need a little more context. With this PR I can use additional features computed from the input features for the surrogate model. These computations can only be 

- sum, 
- addition, 
- mean, and 
- weighted sum

correct? Now If I optimize the acquisition function, I optimize over the original inputs and have these as additional transforms. Did I get this right? And this only makes sense to be applied to continuous descriptors? Why?

jduerholt:
@TobyBoyne:

 > I haven't been able to closely look at the changes to benchmark.py but it all seems okay from a brief glance.

This is also only slightly related, it gives some benchmarks to use the new machinery.

> You're right about the solvent mixtures. You also need to constrain the weighted sum to be == 1 (which I think you have done but I'm not sure), and in the case of only using two solvents must have an NChooseK constraint. 

Yeah, fully agree, but this is existing machinery. We have to work on improved handling of NChooseKs next. But I have an idea for this already.

> If I were able to do a full review now, I would double check that when you use the dimension-scaled prior, this doesn't scale when adding more ContinuousDescriptorInputs.

I am not sure, If I fully agree here, I give an few example on the current behavior:

If you have 10 solvents and each solvent is defined by three descriptors, and you add a WeightedSumFeature over all descriptors, you add three dimensions to the GP surrogate, in total you have then 13 dimensions in the GP, and the dimensionality scaled prior should also act on this 13. If you specify in the kernel of the surrogate, that only the three weighted descriptors should be considered, than the dimensionality scaled priors should act over dim 3.  Agree? This is how it currently works.

jduerholt:
@bertiqwerty fine for you to go ahead with this? Or do you need more explanation?

bertiqwerty (approved):
Thanks Johannes.

Issue #677, opened 2025-11-27
[FEATURE REQUEST]:  removal of dead candidate attributes in DoE strategy or fix/refactor + tutorial on them


rosonaeldred:
### Motivation

According to a quick claude analysis, we have a bunch of attributes of DoEStrategy that are unused. [I gave it the bofire repo to look through] 

@dlinzner-bcs  could you explain what they were supposed to do, or what I can remove without guilt? 

Attribute | Type | Where Defined | Where Used | Purpose | Status
-- | -- | -- | -- | -- | --
self._partially_fixed_candidates | None | DoEStrategy.__init__ (line 33) | Never used | Intended for experiments with some fixed values |  https://github.com/experimental-design/bofire/blob/152b8ac7c42293b7b2a525e894382892de0c1c0b/tests/bofire/strategies/test_doe.py#L270
self._fixed_candidates | None | DoEStrategy.__init__ (line 34) | Never used | Intended for completely fixed experiments | https://github.com/experimental-design/bofire/blob/152b8ac7c42293b7b2a525e894382892de0c1c0b/tests/bofire/strategies/doe/test_design.py#L347
self._sampling | DataFrame or None | DoEStrategy.__init__ (lines 35-39) | Never used | Stores sampling data from data model | https://github.com/experimental-design/bofire/blob/152b8ac7c42293b7b2a525e894382892de0c1c0b/tests/bofire/strategies/test_doe.py#L491
self._return_fixed_candidates | bool | DoEStrategy.__init__ (line 40) | DoEStrategy._ask() (line 113) | Controls whether to return fixed candidates in output | ✅ Active
self._candidates | DataFrame | DoEStrategy.set_candidates() (line 58) | Storage only | Private storage set by set_candidates() method | ✅ Active
self.candidates | Property | Inherited from Strategy | DoEStrategy._ask() (lines 62, 69, 70) | Public interface to access candidates | ✅ Active

Key Findings
🚨 Dead Code Identified
Three attributes are initialized but never used:
_partially_fixed_candidates
_fixed_candidates
_sampling

✅ Active Functionality
The working candidates system uses:

_candidates : Private storage
self.candidates: Public property (inherited from parent Strategy class
_return_fixed_candidates: Controls output behavior

Code Flow in _ask() Method

```
# Check if candidates exist
if self.candidates is not None:
    # Transform candidates to new domain
    adapted_partially_fixed_candidates = self._transform_candidates_to_new_domain(...)
    
[...]

Issue #631, opened 2025-09-09
[FEATURE REQUEST]: new function in doe/utils to unify our calls for model matrix


rosonaeldred:
### Motivation

Proposed feature: 
small helper function to be consistent about how we create the model matrix.  

Background: 
Sometimes, we use formulaic to get the model matrix, 
sometimes we use formulaic to get model terms, then build a tensor. 

See: 
Formulaic to get model matrix: 

- https://github.com/experimental-design/bofire/blob/8e70e331cbcd7b756ca9cd1b6fa24b7ea73bd6b6/bofire/strategies/doe/utils.py#L261
- https://github.com/experimental-design/bofire/blob/8e70e331cbcd7b756ca9cd1b6fa24b7ea73bd6b6/bofire/strategies/doe/objective.py#L246

formulaic for the terms probably, then tensor stuff 
https://github.com/experimental-design/bofire/blob/8e70e331cbcd7b756ca9cd1b6fa24b7ea73bd6b6/bofire/strategies/doe/objective.py#L107

### Describe the solution you'd like to see implemented in BoFire.

small helper function to be consistent about how we create the model matrix, living in doe/utils

@dlinzner-bcs 's opinion is to do it more with torch and less with formulaic 

### Pull Request

None

PR #684, opened 2025-12-10
675 fim matrix rank etc (replacing other pull request) 

rosonaeldred:
## Motivation

Use case: you have an existing DoE or experiments, and would like to get a sense of how 'good' they are for a planned new design with e.g. the same domain.  

There are various ways to do this, but the cheapest/fastest is to take the rank of the fisher information matrix (XTX) associated to the design matrix (X) built from your existing experiments and the (larger) design you're aiming for.  I acknowledge there are more sensitive approaches and this yields at best an overly optimistic estimate, but this is the one our group at BASF aligned on as the most pragmatic (we can talk about Kappa/condition number for the curious). 

In adding this functionality, I also tackled the issues with naming and some duplicate code in  #677  . No breaking changes, just internal variables renamed. Also added comments. 

### Functionality description: 
the quick and dirty way to say how many re-usable experiments are there would be to make the design matrix for the desired extended design, transform your experiments into that space and take the rank of that matrix.

Formulaic can handle this for continuous variables.
It can not handle it for categorical and discrete due to how we dummy-encode them. Thanks to https://github.com/experimental-design/bofire/issues/631 and https://github.com/experimental-design/bofire/pull/673 , we have the matrix we need already exposed, and then need to apply torch.rank (or whatever the appropriate rank function is).

I'll additionally altered logic around the get_number_of_experiments to have an optional handling of candidates/data that would then use this. That is in DoEStrategy, get_additional_experiments_needed  (no breaking change, just a new function). 
Open to re-naming. 

#### ModelBasedObjective 

recall: from #631 now has a matrix function https://github.com/experimental-design/bofire/blob/db23c00914c9c545f3f065457ac3e5271ece285a/bofire/strategies/doe/objective.py#L103

(Added a test for that in test_objective. )

#### DoE Strategy

Added function to get candidate rank https://github.com/experimental-design/bofire/blob/db23c00914c9c545f3f065457ac3e5271ece285a/bofire/strategies/doe/objective.py#L109
It calculates the rank of the candidates in the relevant model matrix, using the function from ModelBasedObjective to get the model matrix tensor. 

**CORE FUNCTIONALITY: get_additional_experiments_needed**

Basically: experiments for the design + domain minus the goodness of the candidates. Mildly confusing, because the required number of experiments is often (min needed) + 3. 
 
https://github.com/experimental-design/bofire/blob/db23c00914c9c545f3f065457ac3e5271ece285a/bofire/strategies/doe_strategy.py#L240-253

####  utils_categorical_discrete 

Fair to warn others: 
The _transform_candidates_to_new_domain method sets missing (aux) columns to None, which creates NaN values. This happens at line 279:

https://github.com/experimental-design/bofire/blob/a5c092735a8ad4176c8f7a8cd3ebedcd5b9e6f3a/bofire/strategies/doe_strategy.py#L273-L279

Created a function that actually transforms things. 

https://github.com/experimental-design/bofire/blob/4a2758f97ae84263e004f07f82e8122d296a4240/bofire/strategies/doe/utils_categorical_discrete.py#L420-L425

Yes

## Test Plan

in /tests/bofire/strategies/test_doe.py

- def test_get_candidate_rank():
- def test_get_candidate_rank_categorical_discrete():
- def test_get_additional_experiments_needed()

### DIFF
--- a/bofire/strategies/doe_strategy.py
+++ b/bofire/strategies/doe_strategy.py
@@ -18,6 +18,7 @@ from bofire.strategies.doe.objective import ModelBasedObjective, get_objective_f
 from bofire.strategies.doe.utils import get_formula_from_string, n_zero_eigvals
 from bofire.strategies.doe.utils_categorical_discrete import (
     create_continuous_domain,
+    encode_candidates_to_relaxed_domain,
     filter_out_categorical_and_categorical_auxilliary_vars,
     filter_out_discrete_auxilliary_vars,
     project_candidates_into_domain,
@@ -178,19 +179,28 @@ class DoEStrategy(Strategy):
         if self.candidates is None:
             return 0
 
-        # Only works for DoEOptimalityCriterion (model-based criteria), not SpaceFilling
+        # Only works for DoEOptimalityCriterion (model-based criteria)
         if not isinstance(self._data_model.criterion, DoEOptimalityCriterion):
             raise ValueError(
-                "get_candidate_rank() only works with DoEOptimalityCriterion, not SpaceFillingCriterion"
+                "get_candidate_rank() only works with DoEOptimalityCriterion"
             )
 
         # Step 1: get_relaxed_domain(original_domain)
-        relaxed_domain, *_ = create_continuous_domain(domain=self.domain)
-
-        # Step 2: get_relaxed_candidates(candidates)
-        relaxed_candidates = self._transform_candidates_to_new_domain(
+        (
             relaxed_domain,
-            self.candidates,
+            mappings_categorical_var_key_to_aux_var_key_state_pairs,
+            mapping_discrete_input_to_discrete_aux,
+            aux_vars_for_discrete,
+            mapped_aux_categorical_inputs,
+            mapped_continous_inputs,
+        ) = create_continuous_domain(domain=self.domain)
+
+        # Step 2: Properly encode candidates to relaxed domain
+        relaxed_candidates = encode_candidates_to_relaxed_domain(
+            candidates=self.candidates,
+            mappings_categorical_var_key_to_aux_var_key_state_pairs=mappings_categorical_var_key_to_aux_var_key_state_pairs,
+            mapping_discrete_input_to_discrete_aux=mapping_discrete_input_to_discrete_aux,
+            domain=self.domain,
         )
 
         # Step 3: get_objective_function (combines model + objective)
@@ -198,42 +208,26 @@ class DoEStrategy(Strategy):
         objective_function = get_objective_function(
             criterion=self._data_model.criterion,
             domain=relaxed_domain,
-            n_experiments=n_candidates,  # not actually used in this context, so effectively a dummy value
+            n_experiments=n_candidates,
             inputs_for_formula=self.domain.inputs,
         )
 
         # Step 4 & 5: Combined tensor_to_model_matrix + rank calculation
         if isinstance(objective_function, ModelBasedObjective):
-            # Handle relaxed candidates which may have NaN values for auxiliary variables
-            # Fill NaN with 0.0 and ensure all data is numeric
-            relaxed_candidates_clean = relaxed_candidates.copy()
-
-            # Convert object columns to numeric where possible, coerce errors to NaN
-            for col in relaxed_candidates_clean.columns:
-                if relaxed_candidates_clean[col].dtype == "object":
-                    relaxed_candidates_clean[col] = pd.to_numeric(
-                        relaxed_candidates_clean[col], errors="coerce"
-                    )
-
-            # Fill all NaN values with 0.0
-            relaxed_candidates_clean = relaxed_candidates_clean.fillna(0.0)
-
             # Ensure we only use columns that match the relaxed domain inputs
             expected_columns = relaxed_domain.inputs.get_keys()
-            relaxed_candidates_clean = relaxed_candidates_clean[expected_columns]
+            relaxed_candidates_clean = relaxed_candidates[expected_columns]
 
             # Convert to tensor
             candidates_tensor = torch.tensor(
                 relaxed_candidates_clean.to_numpy(), dtype=torch.float64
             )
 
-            # get candidate model matrix using objective:
+            # Get candidate model matrix using objective
             candidates_model_matrix = objective_function.tensor_to_model_matrix(
                 candidates_tensor
             )
 
-            # Convert tensor to model matrix
-            # X = tensor_to_model_matrix(candidates_tensor)
             model_matrix_rank = torch.linalg.matrix_rank(candidates_model_matrix).item()
 
             return model_matrix_rank
--- a/bofire/strategies/doe/utils_categorical_discrete.py
+++ b/bofire/strategies/doe/utils_categorical_discrete.py
@@ -415,3 +415,55 @@ def project_candidates_into_domain(
         ),
         columns=columns,
     )
+
+
+def encode_candidates_to_relaxed_domain(
+    candidates: pd.DataFrame,
+    mappings_categorical_var_key_to_aux_var_key_state_pairs: Dict[str, Dict[str, str]],
+    mapping_discrete_input_to_discrete_aux: Dict[str, List[str]],
+    domain: Domain,
+) -> pd.DataFrame:
+    """
+    Encode candidates from the original domain to the relaxed domain with proper auxiliary variable values.
+
+    For categorical inputs, creates one-hot encoding with aux_{key}_{category} naming.
+    For discrete inputs, creates auxiliary variables with values based on the discrete input value.
+    Continuous inputs are kept as-is.
+
+    Args:
+        candidates: DataFrame with candidates in original domain (e.g., x1=5, cat='A')
+        mappings_categorical_var_key_to_aux_var_key_state_pairs: Mapping from categorical keys to
+            their auxiliary variable names and category values
+        mapping_discrete_input_to_discrete_aux: Mapping from discrete input keys to their
+            auxiliary variable names
+        domain: The original (non-relaxed) domain
+
+    Returns:
+        DataFrame with candidates in relaxed domain (e.g., x1=5, aux_cat_A=1, aux_cat_B=0)
+    """
+    encoded = candidates.copy()
+
+    # Encode categorical inputs
+    for (
+        cat_key,
+        aux_mapping,
+    ) in mappings_categorical_var_key_to_aux_var_key_state_pairs.items():
+        if cat_key in encoded.columns:
+            # For each category, create a column with 1 if that category is active, 0 otherwise
+            for aux_key, category_value in aux_mapping.items():
+                encoded[aux_key] = (encoded[cat_key] == category_value).astype(float)
+
+    # Encode discrete inputs
+    for discrete_key, aux_keys in mapping_discrete_input_to_discrete_aux.items():
+        if discrete_key in encoded.columns:
+            discrete_input = domain.inputs.get_by_key(discrete_key)
+            assert isinstance(discrete_input, DiscreteInput)
+
+            # For each allowed discrete value, create an auxiliary variable
+            for aux_key, discrete_value in zip(aux_keys, discrete_input.values):
+                # Set to 1 if the discrete input equals this value, 0 otherwise
+                encoded[aux_key] = (encoded[discrete_key] == discrete_value).astype(
+                    float
+                )
+
+    return encoded
--- a/tests/bofire/strategies/doe/test_utils_categorical_discrete.py
+++ b/tests/bofire/strategies/doe/test_utils_categorical_discrete.py
@@ -14,6 +14,7 @@ from bofire.data_models.features.api import (
 )
 from bofire.strategies.doe.utils_categorical_discrete import (
     create_continuous_domain,
+    encode_candidates_to_relaxed_domain,
     filter_out_categorical_and_categorical_auxilliary_vars,
     filter_out_discrete_auxilliary_vars,
     project_candidates_into_domain,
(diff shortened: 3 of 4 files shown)


bofire/strategies/doe/objective.py
```diff
@@ -106,6 +106,19 @@ def tensor_to_model_matrix(self, D: Tensor) -> Tensor:
         var_dict["torch"] = torch  # type: ignore
         return eval(str(self.model_terms_string_expression), {}, var_dict)
 
+    def get_fisher_information_matrix_rank(self, D: Tensor) -> int:
```
Osburg: Hi rosona :) As described in #631 I think we should work the formulaic implementation here and spontaneously I would say we should consider implementing this as a utility function in bofire/strategies/doe/utils.py (but not as part of the objective. Is there a reason why this should be part of the objective? <-- at least I feel in the objectives should be optimization-related things. The model rank is independent of the (model-based) objective type and is not used during optimization, but rather for evaluation, right? Why do we need it to operator on tensors? We do not want to make use of any pytorch/tensor properties that would not be supported by numpy arrays as well right? What do you think?

bofire/strategies/doe/objective.py
```diff
@@ -106,6 +106,19 @@ def tensor_to_model_matrix(self, D: Tensor) -> Tensor:
...
+        """Get the rank of the Fisher Information Matrix (X.T @ X) from the design matrix tensor.
+        Standard practice in experimental design is to include the intercept when calculating
+        FIM rank, as it represents a real model parameter that needs to be estimated.
+        Args:
+            D (Tensor): Design matrix tensor.
+        Returns:
+            int: The rank of the Fisher Information Matrix (includes intercept if present).
+        """
+        X = self.tensor_to_model_matrix(D)
+        XTX = X.T @ X
```
Osburg: Tiny and unimportant thing, but I think we don't need to compute X.T @ X, the rank of X.T @ X should be the same as the rank of X, right?

uv.lock
Osburg, line 1: Is this file needed for anything or was it added accidentally? :D

rosonaeldred: i added it to gitignore, it shouldn't have been added. :/

bofire/strategies/doe/objective.py
rosonaeldred: @Osburg the conclusion of 631 was @dlinzner-bcs  exposing the model matrix tensor, which lives in objective. 
https://github.com/experimental-design/bofire/blob/6a4d3175c015031f8ce39e616e89c46ab0b4edc7/bofire/strategies/doe/objective.py#L103

Why I want that: because it handles how we handle the categoricals and discretes properly and cleanly.

bofire/strategies/doe/objective.py
rosonaeldred: Fair. This was originally step one of a larger computation involve Kappa and conditioning, which uses the FIM.

rosonaeldred:
Looks like tests/bofire/strategies/test_doe.py::test_categorical_discrete_doe is flaky. 

Analysis: 

This is a flaky test. It fails intermittently due to a numerical tolerance mismatch.

Root Cause
The test configures the SCIP solver with an extremely strict feasibility tolerance:

However, the validation in domain.validate_candidates() uses a default tolerance of tol=1e-5. The SCIP solver produces solutions with tiny numerical errors (like -0.000000, 0.000005, 14.999966) that:

- Are acceptable to the solver (< 1e-8 feasibility tolerance)
- Sometimes accumulate to exceed the validation tolerance (≥ 1e-5)
- This creates intermittent failures when numerical errors happen to be just above 1e-5.


The constraint violations shown are all well within acceptable numerical precision for optimization solvers - they're just triggering a validation check that's slightly too strict for the solver's numerical behavior.


Examples:

If you have a constraint like x + y + z = 15

With tolerance 1e-4 (0.0001):

14.9999 would be considered "equal to 15" ✓
15.0002 would be considered "equal to 15" ✓

With tolerance 1e-8 (0.00000001):
14.99999999 would be considered "equal to 15" ✓
14.9999 would be considered "NOT equal to 15" ✗ (too far off)


The validation flow is:

- Test calls strategy.ask() at line 235 of the test file
- Which calls at line 159 of strategy.py (without passing tol
- Which uses the default tol=1e-5 at line 422 of domain.py
- Which passes that to constraints.is_fulfilled(candidates, tol=tol) at line 451
- The constraint checking then uses that 1e-5 tolerance to determine if the numerical solutions from SCIP are "good enough"

bofire/strategies/doe_strategy.py
```diff
@@ -162,6 +173,92 @@ def get_required_number_of_experiments(self) -> Optional[int]:
...
+            self.candidates,
+        )
+
+        # Step 3: get_objective_function (combines model + objective)
+        n_candidates = len(self.candidates)
+        objective_function = get_objective_function(
+            criterion=self._data_model.criterion,
+            domain=relaxed_domain,
+            n_experiments=n_candidates,  # not actually used in this context, so effectively a dummy value
+            inputs_for_formula=self.domain.inputs,
```
dlinzner-bcs, line 212: do you want to have the domain.inputs here? If so, then discrete and categorical variables will be treated as continuous variables.

tests/bofire/strategies/doe/test_objective.py
```diff
@@ -586,3 +588,55 @@ def test_IOptimality_instantiation():
     )
 
     assert np.allclose(domain.constraints(i_optimality.Y), 0.0)
+
+
+def test_tensor_to_model_matrix():
```
dlinzner-bcs, line 599: It would be good to see such a test for categorical and discrete variables as well. Then, we will know if they are handled correctly.

bofire/strategies/doe_strategy.py
rosonaeldred: ah! it should be relaxed_domain.inputs, ne?

rosonaeldred: wait. This is exactly as written in _ask (looking at the main branch) 
https://github.com/experimental-design/bofire/blob/4e2e34759cd15ca052777dfe1c114455e2de2730/bofire/strategies/doe_strategy.py#L95-L99

dlinzner-bcs: yes. I just checked again. you did it correctly. the handling of categoricals is within the [get_formula_from_string() ](https://github.com/experimental-design/bofire/blob/4e2e34759cd15ca052777dfe1c114455e2de2730/bofire/strategies/doe/utils.py#L62) function. And discretes should be handled as continuous variables for the formula.

dlinzner-bcs: here is the line with the categorical handling https://github.com/experimental-design/bofire/blob/4e2e34759cd15ca052777dfe1c114455e2de2730/bofire/strategies/doe/utils.py#L98

tests/bofire/strategies/doe/test_objective.py
rosonaeldred: can you be a little clearer about the behavior you wanted to see tested here?

dlinzner-bcs: a test that generates the model matrix for a domain and formula with categorical inputs. Does the model matrix contain the indicator variables for the categoricals?

bofire/strategies/doe_strategy.py
```diff
@@ -162,6 +173,92 @@ def get_required_number_of_experiments(self) -> Optional[int]:
...
 
+    def get_candidate_rank(self) -> int:
+        """Get the rank of the model matrix with the current candidates."""
+        if self.candidates is None:
+            return 0
+
+        # Only works for DoEOptimalityCriterion (model-based criteria), not SpaceFilling
+        if not isinstance(self._data_model.criterion, DoEOptimalityCriterion):
+            raise ValueError(
+                "get_candidate_rank() only works with DoEOptimalityCriterion, not SpaceFillingCriterion"
```
dlinzner-bcs: minor: I recommend to remove ".., not SpaceFillingCriterion". This might become outdated at one point.

dlinzner-bcs: .. that it is the only criterion that does not work, I mean

bofire/strategies/doe_strategy.py
```diff
@@ -162,6 +173,92 @@ def get_required_number_of_experiments(self) -> Optional[int]:
...
+        )
+
+        # Step 4 & 5: Combined tensor_to_model_matrix + rank calculation
+        if isinstance(objective_function, ModelBasedObjective):
+            # Handle relaxed candidates which may have NaN values for auxiliary variables
+            # Fill NaN with 0.0 and ensure all data is numeric
+            relaxed_candidates_clean = relaxed_candidates.copy()
+
+            # Convert object columns to numeric where possible, coerce errors to NaN
+            for col in relaxed_candidates_clean.columns:
```
dlinzner-bcs: this is weird. Why is a data transformation/cleaning needed here? If it is really necessary, it should not be handled here imo.

bofire/strategies/doe_strategy.py
```diff
@@ -162,6 +173,92 @@ def get_required_number_of_experiments(self) -> Optional[int]:
...
+                if relaxed_candidates_clean[col].dtype == "object":
+                    relaxed_candidates_clean[col] = pd.to_numeric(
+                        relaxed_candidates_clean[col], errors="coerce"
+                    )
+
+            # Fill all NaN values with 0.0
+            relaxed_candidates_clean = relaxed_candidates_clean.fillna(0.0)
```
dlinzner-bcs: same here. And NaN != 0.0! That is blasphemy!!!!

bofire/strategies/doe_strategy.py
```diff
@@ -162,6 +173,92 @@ def get_required_number_of_experiments(self) -> Optional[int]:
...
+
+            # Ensure we only use columns that match the relaxed domain inputs
+            expected_columns = relaxed_domain.inputs.get_keys()
+            relaxed_candidates_clean = relaxed_candidates_clean[expected_columns]
```
dlinzner-bcs: This is somewhat all handled by the set_candidates() function. Would the workflow: 1) self.set_candidates(candidates) 2) self.get_candidate_rank() be more code efficient?

bofire/strategies/doe/objective.py
rosonaeldred: I adjusted the code, dropping the rank from objective and pulling the functionality into doe_strategy, making use of the model_matrix exposed in objective

dlinzner-bcs (changes requested):

bofire/strategies/doe_strategy.py
rosonaeldred: 'Handle relaxed candidates which may have NaN values for auxiliary variables' this is apparently the source confusion. I don't think they can ever be NaN, can they?

rosonaeldred: The _transform_candidates_to_new_domain method sets missing (aux) columns to None, which creates NaN values. This happens at line 279:

https://github.com/experimental-design/bofire/blob/a5c092735a8ad4176c8f7a8cd3ebedcd5b9e6f3a/bofire/strategies/doe_strategy.py#L273-L279

rosonaeldred: @dlinzner-bcs  the transform isn't the one i need, do you know if one exists?

rosonaeldred: off to build a new function...

rosonaeldred: Replaced with 

https://github.com/experimental-design/bofire/blob/4a2758f97ae84263e004f07f82e8122d296a4240/bofire/strategies/doe/utils_categorical_discrete.py#L420-L425

bofire/strategies/doe_strategy.py
rosonaeldred: the relaxed domain does not contain the original names of the discrete and categorical variables. 
I think re-setting candidates might mess some things up.  
Also, you would need a doe_strategy with the relaxed domain as its domain now, since set_candidates uses self.domain.  

=============
Saucy claude analysis: 

The column filtering on lines 217-219 (local numbering, same as the ones you picked) is necessary and not redundant because:

1. encode_candidates_to_relaxed_domain() preserves original columns (e.g., cat1, x1) AND adds auxiliary columns (e.g., aux_cat1_A, aux_x1_0__decpt__3)

2. The relaxed domain only contains:


- Auxiliary variables for categoricals/discretes
- Original continuous variables
- NOT the original categorical/discrete columns

3. So we need to filter to only the columns the relaxed domain expects before converting to tensor

Example:

Original: cat1='A', x1=5.0, x2=2.5
After encoding: cat1='A', x1=5.0, x2=2.5, aux_cat1_A=1.0, aux_cat1_B=0.0, aux_x1_5=1.0, aux_x1_10=0.0
Relaxed domain expects: x2, aux_cat1_A, aux_cat1_B, aux_x1_5, aux_x1_10 (no cat1, no x1)

tests/bofire/strategies/doe/test_objective.py
rosonaeldred: # Column 5: x1:aux_x3_A
...

dlinzner-bcs (approved):
Looks good :) Thank you @rosonaeldred!

tests/bofire/surrogates/test_multitask_gps.py
```diff
@@ -109,7 +110,14 @@ def test_MultiTaskGPModel(kernel, scaler, output_scaler, task_prior):
     benchmark = MultiTaskHimmelblau()
     inputs = benchmark.domain.inputs
     outputs = benchmark.domain.outputs
-    experiments = benchmark.f(inputs.sample(10), return_complete=True)
+    # Sample both tasks to ensure both are present in training data
```
dlinzner-bcs, line 113: what happened here?

rosonaeldred: Flaky test. It was spitting out an ugly error message. 
In the pipeline, it failed 

> FAILED tests/bofire/surrogates/test_multitask_gps.py::test_MultiTaskGPModel[kernel0-NORMALIZE-STANDARDIZE-None] - ValueError: Received invalid raw task values. Expected raw value to be in {1}, but got unexpected task values: {0}.

Then I looked into it. 

inputs.sample(10) randomly samples tasks, and it might only sample one task (e.g., all "task_1" or all "task_2"). 
Then when predicting with inputs.sample(5), it might sample the other task, causing the encoding mismatch error.
Which is the error message above. 

The fix is to ensure both tasks are present in the training data. 
Now the test explicitly samples 5 experiments from each task

PR #688, opened 2025-12-18
Index kernel and Positive Index kernel

KislayaRavi:
## Motivation
There are two major modifications:
- Added two new kernels to handle categorical and discrete variables:
  - [Index Kernel](https://docs.gpytorch.ai/en/latest/_modules/gpytorch/kernels/index_kernel.html)
  - [Positive Index Kernel](https://botorch.readthedocs.io/en/stable/_modules/botorch/models/kernels/positive_index.html) 
- Added feature in MixedSingleTaskGPSurrogate that allows user to allocate different types of kernels according to the feature name using a kernel dictionary. Earlier all categorical variables and all continuous variables had same type of kernel.

Yes

## Test Plan

- Added both unit tests and integration tests,
- Added PyDantic variable checks.


bofire/data_models/surrogates/mixed_single_task_gp.py
```diff
@@ -103,6 +105,31 @@ def _update_hyperparameters(
...
+    """
+    A Gaussian Process surrogate model for mixed (continuous and categorical) single-task optimization problems.
+    This surrogate combines continuous and categorical kernels to handle optimization problems with
+    both continuous variables and categorical features. It requires at least one categorical input
+    and supports ordinal encoding for categorical variables.
+    Attributes:
+        type: Literal identifier for the surrogate type.
+        continuous_kernel: Kernel for continuous features, defaults to RBF kernel with ARD.
+        categorical_kernel: Kernel for categorical features, defaults to Hamming distance kernel.
+        kernel_dict: Optional dictionary mapping feature keys to specific kernels. Useful for custom kernel assignments like different kernels for different different features.
```
jduerholt: This is already implemented at global level, as long as you have a kernel which is derived from `FeatureSpecificKernel`, you can explicitly set the `features` attribute and provide a list of the feature keys on which the kernel should operate. Note that exactly this behavior is exploited in the MixedGP, for this have a look here: https://github.com/experimental-design/bofire/blob/c8cd6723867728d6367aa74acd0182cb39b242b8/bofire/data_models/surrogates/mixed_single_task_gp.py#L136 and here: https://github.com/experimental-design/bofire/blob/c8cd6723867728d6367aa74acd0182cb39b242b8/bofire/surrogates/mapper.py#L25

bofire/data_models/kernels/categorical.py
```diff
@@ -13,3 +16,36 @@ class HammingDistanceKernel(CategoricalKernel):
     ard: bool = True
     lengthscale_prior: Optional[AnyPrior] = None
     lengthscale_constraint: Optional[AnyPriorConstraint] = None
+
+
+class IndexKernel(CategoricalKernel):
```
jduerholt, line 21: Can you provide docstrings that explain what this kernels are doing? Also in comparison to the hamming distance kernel?

jduerholt (commented):
Hi @KislayaRavi,

thanks for the PR, the two new kernels look good to me. Maybe, you can elaborate a bit in the docstring about what they are doing and what are there advantages etc. 

Regarding the kernel_dict, I do not think that this is need as it is alrady possible to provide feature specific kernels. Or am I overlooking something?

Best,

Johannes

bofire/data_models/surrogates/mixed_single_task_gp.py
KislayaRavi: Let us consider a case where we have three categorical variables and we want to use `HammingDistance` kernel for the first, `IndexKernel` for the second and `PositiveIndexKernel` for the third variable. In the existing implementation, user can define only one type of categorical kernel which will eventually be used for all the categorical variables. I am not sure if we can assign different kinds of categorical kernels to different categorical variable with existing implementation.

jduerholt: How should the kernels be combined? In the mixed GP it is categorical*continuous + continuous+categorical (a little bit simplified). Do you want to have for the categorical part a product of the HammingDistanceKernel, IndexKernel and PositiveIndexKernel, or a sum? I think, this could be also done, but first I need to understand it ;)

KislayaRavi: Suppose we have kernel1 for var1, kernel2 for for var2, .........., kernelN for varN. Then the combination kernel is
(kernel1 + kernel2 + .... + kernelN) + (kernel1 * kernel2 * ....* kernelN)
Each of the kernels can be either categorical or continuous depending upon the variable.

jduerholt: In general this is out of the box possible with the `SingleTaskGPSurrogate`, by using `AdditiveKernel` and `ProductKernel`. Note that the `MixedSingleTaskGPSurrogate` is also just a proxy to a `SingleTaskSurrogate` which simplifies kernel setup for the user. I would prefer to keep this option in the `SingleTaskGPSurrogate` and to make it not too complex here. The other option would be to make the validator for the categorical kernel less strict or check for more options there: https://github.com/experimental-design/bofire/blob/63e61229c6d2025771108e78935325d14ba536b7/bofire/data_models/surrogates/mixed_single_task_gp.py#L175

What do you think?

KislayaRavi: I agree with you. So, I reverted ```MixedSingleTaskGPSurrogate``` to original condition. I am now directly using ```SingleTaskGPSurrogate```.

bofire/data_models/kernels/categorical.py
KislayaRavi: I added the docstring explaining both the kernels. Additionally, I also added a tutorial on Index Kernels. There I have compared Hamming Distance and Index Kernel, and explained the purpose of Positive Index Kernel.  I pushed the changes.

bofire/data_models/kernels/categorical.py
```diff
@@ -13,3 +16,94 @@ class HammingDistanceKernel(CategoricalKernel):
...
+    Mathematically, the look up table is represented as:
+
+    $$
+    k(i, j) = \left(BB^\top + \text{diag}(\mathbf v) \right)_{i, j} # type: ignore
+    $$
+
+    where $B$ is a low-rank matrix, and $\mathbf v$ is a  non-negative vector.
+
+    Attributes:
+        type (Literal["IndexKernel"]): Identifier for the kernel type.
```
jduerholt: Remove the type attribute, we never show it in the doc. it is purly internal and needed for serialization/deserilaization

bofire/data_models/kernels/categorical.py
```diff
@@ -13,3 +16,94 @@ class HammingDistanceKernel(CategoricalKernel):
...
+
+    $$
+        k(i, j) = \frac{(LL^T)_{i,j}}{(LL^T)_{t,t}}
+    $$
+
+    where $L$ is a lower triangular matrix with positive elements and $t$ is the
+    target_task_index.
+
+    NOTE: This kernel should only be used when the correlation between different categories
+    is expected to be positive.
```
jduerholt, line 74: Can you add a white line in between?

bofire/data_models/kernels/categorical.py
```diff
@@ -13,3 +16,94 @@ class HammingDistanceKernel(CategoricalKernel):
...
+    Attributes:
+        type (Literal["PositiveIndexKernel"]): Identifier for the kernel type.
```
jduerholt: Remove the type attribute.

tests/bofire/kernels/test_mapper.py
```diff
@@ -406,3 +445,70 @@ def test_map_WedgeKernel():
 
     assert isinstance(k, conditionalKernels.WedgeKernel)
     assert k.base_kernel.active_dims.tolist() == [0, 1, 2, 4]
+
+
+"""Test suite for IndexKernel validation."""
```
jduerholt: Maybe put the ones that are invalid to the `add_invalid` functionality in the specs. Would be a bit cleaner and more in line to the overall test suite.

bofire/data_models/kernels/categorical.py
KislayaRavi: Done, I will take note of this thing for the next tasks from now.

bofire/data_models/kernels/categorical.py
KislayaRavi: Done

bofire/data_models/kernels/categorical.py
KislayaRavi: Done

tests/bofire/kernels/test_mapper.py
KislayaRavi: I moved the specs check tests in format of ```add_invalid``` or ```add_valid```.

Issue #680, opened 2025-12-05
[FEATURE REQUEST]: SphericalLinearKernel


jduerholt:
### Motivation

We should add the `SphericalLinearKernel` from this paper:

https://arxiv.org/pdf/2512.00170

Code is here: https://github.com/colmont/linear-bo/blob/main/src/kernels/spherical_linear.py

The paper is also a super interesting read!

### Describe the solution you'd like to see implemented in BoFire.

See above.

### Pull Request

None

PR #689, opened 2025-12-30
Issue: 680 spherical linear kernel

KislayaRavi:
## Motivation
This pull request is in response the Issue #680 
Added Spherical Linear Kernel.

Yes

## Test Plan
1. PyDantic validation in data model of SphericalLinearKernel.
2. Added test verifying the specs after creation of the object.
3. Testing the mapping to the correct object.
4. Testing if the surrogate is created using SphericalLinearKernel.

### DIFF
--- a/bofire/benchmarks/svm.py
+++ b/bofire/benchmarks/svm.py
@@ -7,6 +7,7 @@ Adapted from https://github.com/LeoIV/BenchSuite/blob/master/benchsuite/svm.py
 import gzip
 import logging
 import urllib
+from typing import Tuple
 
 import numpy as np
 import pandas as pd
@@ -68,7 +69,7 @@ class SVM(Benchmark):
         self._y_train = y[idxs[:half]]
         self._y_test = y[idxs[half:]]
 
-    def get_data(self):
+    def get_data(self) -> Tuple[np.ndarray, np.ndarray]:
         url_X = (
             "https://github.com/LeoIV/BenchSuite/raw/"
             "73de8c581aacf2dc99120d9cf65b79cbfe2aaf4e/data/svm/CT_slice_X.npy.gz"
@@ -91,7 +92,7 @@ class SVM(Benchmark):
             logging.error("Error downloading or loading data: %s", e)
             raise e
 
-    def _evaluate_single(self, hypers) -> float:
+    def _evaluate_single(self, hypers: np.ndarray) -> float:
         """
         Evaluate SVM error for one set of hyperparameters.
 
--- a/tests/bofire/kernels/test_mapper.py
+++ b/tests/bofire/kernels/test_mapper.py
@@ -431,3 +431,26 @@ def test_map_spherical_linear_kernel():
         features_to_idx_mapper=None,
     )
     assert hasattr(k_with_prior, "lengthscale_prior")
+
+
+def test_spherical_linear_kernel_bounds():
+    """Test SphericalLinearKernel with bounds (-1, 1) to verify buffer values."""
+    input_dim = 5
+    kernel = SphericalLinearKernel(ard=True, bounds=(-1.0, 1.0))
+    k = kernels.map(
+        kernel,
+        batch_shape=torch.Size(),
+        active_dims=list(range(input_dim)),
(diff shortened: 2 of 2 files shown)


bofire/data_models/kernels/continuous.py
```diff
@@ -46,3 +51,21 @@ class InfiniteWidthBNNKernel(ContinuousKernel):
...
+
+
+class SphericalLinearKernel(ContinuousKernel):
+    type: Literal["SphericalLinearKernel"] = "SphericalLinearKernel"
+    ard: bool = True
+    lengthscale_prior: Optional[AnyPrior] = Field(
+        default_factory=HVARFNER_LENGTHSCALE_PRIOR
+    )
+    lengthscale_constraint: Optional[AnyPriorConstraint] = None
+    bounds: Union[tuple[float, float], List[tuple[float, float]]] = (0.0, 1.0)
```
jduerholt, line 69: For what are these bounds needed?

KislayaRavi: One of the steps is to transform the data on a sphere. We assume that originally the data is inside a cube. The variable 'bounds' contains information about the dimension of the cube. We specifically need bound to center. By default, BOFIRE normalizes the continuous input data to range (0, 1). So, there is no need for user to worry about it.

I will add a tutorial using SVM (and other) benchmarks inside the /tutorials/advanced_examples folder. 

I have one concern which I wanted to discuss about the implementation too. The [global_ls](https://github.com/KislayaRavi/bofire/blob/a61b60e01dcd6cdcf6841bdd7d4bad5012484d4b/bofire/kernels/spherical_kernels.py#L161) variable does the square root of dimension scaling. So, we don't need to use HVARFNER_LENGTHSCALE_PRIOR as lengthscale prior. Am I correct or missing something?

jduerholt: I also understand it like this, have you had a look at their implementation?

KislayaRavi: The implementation sets the lengthscale prior to```None```. I did the same. I also added docstring. Moreover, I added a new benchmark SVM and tutorial for spherical linear kernel. I am using the benchmark only in the tutorial. I did not want to use the new benchmark for testing because it is computationally expensive. 

I tested spherical linear kernel on SVM benchmark on my local system. I am getting the following result.
spherical_vs_rbf_loss
I did not go till 1000 iterations as discussed in the paper because of time. But, the plot follows the expected behavior.

bofire/benchmarks/svm.py
```diff
@@ -0,0 +1,146 @@
...
+            with urllib.request.urlopen(url_y) as response:  # type: ignore
+                with gzip.GzipFile(fileobj=response) as f:
+                    y_np = np.load(f)
+            print("Download complete.")
+            return X_np, y_np
+        except Exception as e:
+            logging.error("Error downloading or loading data: %s", e)
+            raise e
+
+    def _evaluate_single(self, hypers) -> float:
```
jduerholt: Can you typehints here?

bofire/data_models/kernels/continuous.py
jduerholt: Very nice. Thanks!

tests/bofire/kernels/test_mapper.py
```diff
@@ -406,3 +408,26 @@ def test_map_WedgeKernel():
 
     assert isinstance(k, conditionalKernels.WedgeKernel)
     assert k.base_kernel.active_dims.tolist() == [0, 1, 2, 4]
+
+
+def test_map_spherical_linear_kernel():
```
jduerholt, line 413: Can you also add a mapping test case where you use kernel specific features and see then if the indices are correct?

KislayaRavi: I added a test to check if the kernel specific features are evaluated and assigned correctly in ```test_mapper.py```.

bofire/benchmarks/svm.py
KislayaRavi: Done

PR #693, opened 2026-01-12
MCTS based ACQF Optimization for Handling of NChooseKs.

jduerholt:
## Motivation

This PR implements an MCTS algorithm for optimitzing over conditional search spaces comprised of NChooseK constraints and/or conditional condinuous input features (`allow_zero`).

It uses the following features:
- canonical ordering
- progressive widening
- global rave stats
- caching of optimized values

This is still WIP and most of the code is (as a test) generated by a wild combo of LLMs.

@TobyBoyne  

Yes.

## Test Plan

Unit tests, not yet done.


jduerholt:
@TobyBoyne: I adjusted the random sampler for also using the MCTS Tree just using random rollouts. In principle the conditional input features should also work there now. But I need to test it. I will also write tmr a bit more about my thoughts, because, I think we can really do super crazy stuff with this new formulation.

bofire/strategies/random.py
```diff
@@ -124,30 +126,77 @@ def _sample_with_nchooseks(
...
+            for constraint in self.domain.constraints.get(NChooseKConstraint):
+                assert isinstance(constraint, NChooseKConstraint)
+                groups.append(
+                    NChooseK(
+                        features=[features2idx[key][0] for key in constraint.features],
+                        min_count=constraint.min_count,
+                        max_count=constraint.max_count,
+                    )
+                )
+            for feat in self.domain.inputs.get(ContinuousInput):
```
jduerholt, line 152: Hmm, is this correct, we should only create a single feature group if the feature is not in an nchoosek, or? Else it is already taken care in the first group, or?

bofire/strategies/random.py
```diff
@@ -124,30 +126,77 @@ def _sample_with_nchooseks(
...
-                num_samples_per_it = 1
-            else:
-                sampled_combinations = unused
-                num_samples_per_it = math.ceil(candidate_count / len(unused))
+
+            mcts = MCTS(
+                groups=Groups(groups=groups),
+                seed=self._get_seed(),
+                reward_fn=lambda x,
+                y: 0.0,  # dummy reward function as we are only using the
```
jduerholt: This looks very strangely formatted, should the y: 0.0 not appear on the same line as the lambda expression?

bofire/strategies/random.py
```diff
@@ -124,30 +126,77 @@ def _sample_with_nchooseks(
...
+            for _ in range(min(self.max_combinations, candidate_count)):
+                combo, _, _trajectory = mcts._rollout(mcts.root)
+                # combo is a tuple of feature indices, we convert it to feature keys
+                combo = tuple(idx2features[idx] for idx in combo)
+                if combo in combinations:
+                    combinations[combo] += 1
+                else:
+                    combinations[combo] = 1
+
+            nchoosek_features = {
```
jduerholt: Is this correct, here we would then neglect single groups, or not? so feature with allow_zero=True, which do not show in the nchoosek constraint, or?

jduerholt: We should also have a test for random strategy for this case.

bofire/strategies/predictives/optimize_mcts.py
```diff
@@ -0,0 +1,1124 @@
...
+from collections.abc import Mapping, Sequence
+from dataclasses import dataclass, field
+from typing import Callable, Optional
+
+import torch
+from botorch.optim import optimize_acqf
+from torch import Tensor
+
+
+# Ideas to improve MCTS performance / robustness in tree-like NChooseK spaces:
```
jduerholt: This comment block can be removed.

bofire/strategies/predictives/optimize_mcts.py
```diff
@@ -0,0 +1,1124 @@
...
+
+        # Initialize root node
+        n_groups = len(groups)
+        self.root = Node(
+            partial_by_group=tuple(() for _ in range(n_groups)),
+            stopped_by_group=tuple(False for _ in range(n_groups)),
+            group_idx=0,
+        )
+
+        # Best found so far
```
jduerholt, line 387: Provide in the comment how this data structure actually looks with an example.

bofire/strategies/predictives/optimize_mcts.py
```diff
@@ -0,0 +1,1124 @@
...
+            self.group_rollout_counts[g] += 1
+
+    def _compute_adaptive_p_stop(
+        self, group_idx: int, current_cardinality: int
+    ) -> float:
+        """Compute adaptive stop probability for a group at a given cardinality.
+
+        Uses learned cardinality statistics to decide whether stopping is better
+        than continuing. Falls back to fixed p_stop_rollout when disabled,
+        no data is available, reward range is zero, or during warmup.
```
jduerholt: Is it really falling back to fixed p_stop_rollout during warmup? It is just bledning with it during warmup, we should make this clear, or?

bofire/strategies/predictives/optimize_mcts.py
```diff
@@ -0,0 +1,1124 @@
...
+            else 1.0
+        )
+        return (1.0 - alpha) * self.p_stop_rollout + alpha * p_learned
+
+    def _normalize_reward(self, reward: float) -> float:
+        """Normalize reward to [0, 1] using running min-max.
+
+        Returns 0.5 when reward range is zero (all rewards identical).
+        """
+        reward_range = self.reward_max - self.reward_min
```
jduerholt: What happens in the first two rollouts? There by definition reward_max and reward:_min are not properly defined, or?

bofire/strategies/predictives/optimize_mcts.py
```diff
@@ -0,0 +1,1124 @@
...
+        if reward_range <= 0:
+            return 0.5
+        return (reward - self.reward_min) / reward_range
+
+    def _score_rollout_actions(
+        self, group_idx: int, legal_actions: list[int]
+    ) -> dict[int, float]:
+        """Score legal rollout actions using learned statistics.
+
+        For each action, computes mean_reward + novelty_weight / sqrt(visits + 1).
```
jduerholt: What is the ratio for the squred term ? Please add to docstring.

bofire/strategies/predictives/optimize_mcts.py
```diff
@@ -0,0 +1,1124 @@
...
+            group_idx: Index of the current group
+            legal_actions: List of legal action indices (may include STOP)
+
+        Returns:
+            Selected action index
+        """
+        scores = self._score_rollout_actions(group_idx, legal_actions)
+        n = len(legal_actions)
+
+        # Compute softmax probabilities
```
jduerholt: We have a torch tependency, can we not just use the softmax implementation from torch? Not sure regarding the temperature parameter. Could be easier to read.

bofire/strategies/predictives/optimize_mcts.py
```diff
@@ -0,0 +1,1124 @@
...
+
+        # Blend with uniform
+        eps = self.rollout_epsilon
+        uniform = 1.0 / n
+        probs = []
+        for i, _a in enumerate(legal_actions):
+            p_policy = exp_scores[i] / z
+            probs.append((1.0 - eps) * p_policy + eps * uniform)
+
+        # Sample
```
jduerholt: Can we not just sample here by the help of numpy choices and passing weights in? https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.choice.html

bofire/strategies/predictives/optimize_mcts.py
```diff
@@ -0,0 +1,1124 @@
...
+
+        partials = list(node.partial_by_group)
+        stoppeds = list(node.stopped_by_group)
+
+        if action == STOP:
+            stoppeds[g] = True
+            next_g = g + 1
+        else:
+            new_partial = partials[g] + (action,)
+            partials[g] = new_partial
```
jduerholt: Could this not be a one liner: `partials[g]`+=(action,)

mcts-report/REPORT.md
```diff
@@ -0,0 +1,2395 @@
...
+
+---
+
+## 3. Results
+
+### 3.1 Summary Tables
+
+#### multigroup_interaction (search space ~4.25M, optimum = 150.0)
+
+| Config | Mean Best | ±Std | Opt Rate | Unique Evals |
```
TobyBoyne, line 150: Nit: I can't actually tell what the bolded lines mean (eg., why are` MCTS (no RAVE+adpt)` and `MCTS (+rpol ε=0.1) ` not highlighted?)

mcts-report/REPORT.md
```diff
@@ -0,0 +1,2395 @@
...
+| MCTS (no RAVE+adpt+norm) | 108.9 | 25.1 | 23% | 455 |
+| **MCTS (+rpol)** | **111.4** | 23.6 | **23%** | 516 |
+| MCTS (+rpol ε=0.1) | 114.1 | 24.0 | 27% | 511 |
+| MCTS (+rpol τ=0.5) | 109.9 | 22.7 | 20% | 510 |
+| MCTS (+rpol τ=2) | 112.5 | 23.1 | 23% | 514 |
+| MCTS (+crave k=100) | 105.3 | 21.1 | 13% | 445 |
+| MCTS (+crave k=300) | 103.5 | 17.3 | 7% | 370 |
+| MCTS (+crave k=500) | 100.2 | 19.5 | 7% | 310 |
+
+#### needle_in_haystack (search space ~4,928, optimum = 100.0)
```
TobyBoyne, line 173: I think this benchmark is probably too easy, given most of the algorithms discussed later in the report reach close enough to 100%. Either that, or the iteration budget for this benchmark should be reduced.

Claude focused far too much on this benchmark, which I think ended up with somewhat meaningless distinctions being made (eg. "A is better than B because 97% success is better than 100% success).

mcts-report/REPORT.md
```diff
@@ -0,0 +1,2395 @@
...
+
+1. **Scale-invariant c_uct**: With raw rewards, `c_uct=1.0` gives different effective exploration pressure on each problem — under-exploring on large_sparse (range 272) and over-exploring on graduated (range 60). With normalization, `c_uct=0.01` gives consistent behavior across all problems.
+
+2. **Improved virtual loss**: Virtual loss on cache hit adds zero reward. With raw rewards centered around, say, 50, this dilutes toward 0 — far below the actual reward range. With normalized rewards in [0, 1], zero is exactly the minimum, making virtual loss dilute toward the worst case rather than an arbitrary anchor.
+
+3. **More exploration on harder problems**: The unique evaluation counts show normalization adds ~20% more exploration (380→455, 550→689) while maintaining focus. The raw configs under-explore large_sparse relative to its enormous search space; normalization partially corrects this.
+
+#### Recommended usage
+
+Normalization should be enabled together with `c_uct=0.01` (or more generally, `c_uct ≈ 1/typical_reward_range`). This combination produces the best overall results and removes the need to tune `c_uct` per problem.
```
TobyBoyne, line 453: We should be standardizing rewards, not normalizing. Think `logEI`, which can get very negative values that would skew the normalization bounds.

mcts-report/REPORT.md
```diff
@@ -0,0 +1,2395 @@
...
+
+The default rollout strategy selects actions uniformly at random (with adaptive p_stop for STOP decisions). This wastes search budget on poor actions even after the tree has accumulated evidence about which actions are good. The **blended softmax rollout policy** replaces uniform rollouts with a learned policy:
+
+1. For each `(group_idx, action)` pair, track `(visit_count, total_reward)` across all rollouts
+2. Score each action: `mean_reward + novelty_weight / sqrt(visits + 1)`
+3. Apply softmax with temperature τ: `p_policy[a] = exp(score[a] / τ) / Z`
+4. Blend with uniform: `p[a] = (1 - ε) * p_policy[a] + ε / |legal_actions|`
+5. STOP is treated as a regular action with its own learned statistics — no special handling needed
+
+#### `MCTS (no RAVE+adpt+norm)` vs `MCTS (+rpol)` — the key comparison
```
TobyBoyne, line 465: It's unclear to me if the second option includes RAVE, adapt, and norm.

mcts-report/REPORT.md
```diff
@@ -0,0 +1,2395 @@
...
+
+**Key result (NIG posterior)**: The Normal-Inverse-Gamma posterior is a transformative improvement over Normal-TS. The best NIG config (**NIG + TS(g,a) + vi + apv**: variance inflation + adaptive prior variance) achieves **80% on multigroup_interaction** (vs UCT's 23%), **100% on needle and mixed** (vs UCT's 100% and 77%), and **47% on large_sparse** (vs UCT's 50% — essentially tied). A single NIG config now matches or exceeds UCT on 5 of 6 problems, with the remaining gap on large_sparse within statistical noise (3pp). See Section 11.13 for full analysis.
+
+**Key result (Adaptive pessimistic strength)**: Scaling the pessimistic offset by node exhaustion did not resolve the vi-vs-comb tradeoff on interaction problems (vi+apv remains best at 80%). However, the no-APV adaptive modes (**NIG + TS(g,a) + apess**) achieved **53% on large_sparse — the first NIG configs to surpass UCT's 50%** on this problem. This revealed that adaptive prior variance (APV) hurts on massive search spaces by over-shrinking the prior too early. NIG now matches or exceeds UCT on all 6 problems when using problem-appropriate configs. See Section 11.14 for full analysis.
+
+---
+
+## 1. Experimental Setup
```
TobyBoyne, line 32: Each benchmark reports `Opt Rate`, which is the percentage of runs for which `abs(mcts.best_value - problem.optimal_value) < 1e-6`. To me, this threshold is too low - an optimizer that gets within 5% of the optimal value is good enough for acqf optimization in my opinion, since your model is going to be wrong anyway. We should be selecting for a consistently good algorithm, not a sometimes perfect algorithm

mcts-report/REPORT.md
```diff
@@ -0,0 +1,2395 @@
...
+
+The current UCT selection rule is:
+
+```
+score = w_total/n_visits + c_uct * sqrt(log(parent_visits) / child_visits)
+```
+
+The exploitation term (mean reward) is in the scale of the rewards; the exploration term is in the scale of `c_uct`. For the balance to work, `c_uct` must be matched to the reward scale. This is why reward normalization exists — it compresses rewards to [0, 1] so `c_uct=0.01` has a consistent meaning across problems (§4.6).
+
+But min-max normalization introduces its own problem: early iterations have a poor estimate of the reward range, so their normalized values are distorted relative to what they'd be with the final bounds. The current best config (+rpol) runs ~100-500 iterations, and the range typically stabilizes within the first ~10-20 iterations, meaning ~10-20% of all backpropagated values in the tree carry normalization error that never gets corrected.
```
TobyBoyne, line 614: I might not be understanding this, but: Claude is suggesting that normalization is bad because you don't know the normalization bounds, and so you might save a highly distorted reward if you have the wrong bounds at first. 

Why not just save the raw reward value, and normalize that value every time you access it? That way, if the bounds of normalization change, the reward can be renormalized? That seems much more sensible to me. It would require a bit more computation (since you would need to renormalize quite often, and each node would have to store the bounds with which it was previously normalized). But it still is the thing that makes mathematical sense to me. 

If you don't want to do that, you can just get a good idea of the range by sampling random points in the search space and evaluating the acqf. You might even skip that, and reason that most of the interesting `logEI` values will be bounded between `[-2, 2]`.

I just thinking that normalization is a really good thing to have, whatever algorithm you use (see my later comment about normalization in TS).

mcts-report/REPORT.md
```diff
@@ -0,0 +1,2395 @@
...
+- **multigroup_interaction** (currently 23%): TS's broader early exploration might help discover the cross-group interaction bonuses, but the search space (~4.25M) is large enough that undirected exploration is costly.
+
+#### Cache hit handling — the key design decision
+
+With UCT, cache hits require the virtual loss hack: backpropagate zero reward to dilute `mean_value` and steer exploration away from exhausted branches (§2.2). With TS, the situation is fundamentally different because selection is stochastic.
+
+**On novel evaluation**: standard Bayesian update — add the observed reward to the child's posterior (increase observation count, update sufficient statistics).
+
+**On cache hit**: do not update the posterior. No new information was gained, so no update is the correct Bayesian action. The critical insight is that TS is *stochastic* — the next iteration draws fresh samples from each posterior, so a different selection path naturally occurs without any need to artificially distort statistics. This is a fundamental advantage over UCT, where not updating means the deterministic score is unchanged and the algorithm deterministically repeats the same path forever.
```
TobyBoyne, line 652: > no new information was gained

Is this true? See my main comment about this being a stochastic problem. Information will be gained when performing a second `optimize_acqf`.

> UCT... deterministically repeats the same path

Is this true? I thought there is still randomness in the rollout policy, so different leaves may be sampled? But maybe I have misunderstood what this is trying to say.

mcts-report/REPORT.md
```diff
@@ -0,0 +1,2395 @@
...
+The best configuration (+rpol) uses a learned softmax rollout policy (§4.7) that:
+
+1. Maintains `(visits, total_reward)` per `(group_idx, action)` pair — STOP included as a regular action
+2. Scores each action: `mean_reward + novelty_weight / sqrt(visits + 1)`
+3. Applies softmax with temperature τ: `p_policy[a] = exp(score[a] / τ) / Z`
+4. Blends with uniform: `p[a] = (1 - ε) * p_policy[a] + ε / |legal_actions|`
+
+This requires three hyperparameters: `rollout_epsilon` (ε=0.3), `rollout_tau` (τ=1.0), `rollout_novelty_weight` (1.0).
+
+#### An important subtlety: adaptive p_stop is dead code in the best config
```
TobyBoyne, line 687: Nit: A bit weird to call this "dead code" - clearer to just say that adaptive `p_stop` is not compatible with a rollout policy, which is fine because that combination is never evaluated.

mcts-report/REPORT.md
```diff
@@ -0,0 +1,2395 @@
...
+7. **Adaptive pessimistic strength**: Scales the pessimistic pseudo-observation by each node's local exhaustion rate (1 - n_obs/n_visits). Fresh nodes get mild pessimism; exhausted nodes get full pessimism. Zero new hyperparameters.
+8. **Adaptive pseudo-count n₀ (negative result)**: Sets n₀ = 1 + log(branching_factor) so nodes with many siblings require more observations before departing from the prior. Tested but found harmful — over-corrects by making posteriors too conservative for the available budget.
+
+**Key result (UCT)**: The cumulative effect of improvements 1-4 transforms MCTS from underperforming random sampling to decisively outperforming it on every problem. The best UCT configuration (**MCTS +rpol**: no RAVE + adaptive p_stop + reward normalization + rollout policy) achieves 100% optimum-finding rate on needle_in_haystack (vs 10% for random), 80% on graduated_landscape (vs 7%), **77% on mixed problems** (vs 3%), and **50% on large_sparse** (vs 0%). Context-aware RAVE re-enables RAVE as a useful signal on mixed problems (80% with k=300 vs 77% for +rpol) while matching +rpol on other problems.
```
TobyBoyne, line 18: This implies that CRAVE is a good idea, but then in section 4 suggests that "Context RAVE is not recommended as a default". From this report, I agree with the latter, and CRAVE shouldn't be used (by default, at least).

mcts-report/REPORT.md
```diff
@@ -0,0 +1,2395 @@
...
+| `rollout_tau` | 1.0 | Rollout TS posterior |
+| `rollout_novelty_weight` | 1.0 | Rollout TS posterior |
+| `adaptive_p_stop` | True (dead code) | Can be removed |
+| `p_stop_rollout` | 0.35 | Can be removed |
+| `p_stop_warmup` | 20 | Can be removed |
+| `p_stop_temperature` | 0.25 | Can be removed |
+
+That is 9 hyperparameters reduced to 0 (or 1 if counting the prior variance, which can be set to a large value and forgotten). The only remaining MCTS hyperparameters would be structural: `pw_k0`, `pw_alpha`, `max_rollout_retries`, and the iteration budget.
+
+### 8.5 Two-Phase Burn-in with Cheap Evaluations
```
TobyBoyne, line 757: I don't think this will work particularly well. The reward at each leaf is the maximizer of the acquisition function. I would guess that - especially for higher dimensions - taking the best of 1,024 random samples would be a really poor approximation. Further, this section implies it would be a uniformly noisy estimate, but it would actually be a biased _under_-estimator - you will never randomly find a point that is actually better than the true acqf maximizer.

If doing some kind of burn-in, it would be much better to just use `optimize_acqf` with a smaller number of restarts, iterations, etc to get a quick approximation.

mcts-report/REPORT.md
```diff
@@ -0,0 +1,2395 @@
...
+
+#### Variance inflation effect — large_sparse
+
+![Variance inflation on large_sparse](convergence_ts_large_sparse_variance_inflation.png)
+
+Variance inflation helps substantially (TS(g,a) from 57 to 84 mean best), but the gap to UCT (130) remains large. The 960M search space requires more unique evaluations than TS with var_infl can produce (575 vs UCT's 750).
+
+### 11.5 Analysis
+
+#### 11.5.1 Variance Inflation Is the Critical Design Decision
```
TobyBoyne, line 1069: I'm really not sure I understand the motivation behind this, I think I need to have a further investigation. It seems to me like variance inflation also increases the probability that this sampled leaf has a *higher* value than it actually does. If we want to penalize visiting leaves that have already been visited, it seems to me like there should be an explicit loss (as with UCT), rather than a larger variance.

mcts-report/REPORT.md
```diff
@@ -0,0 +1,2395 @@
...
+| graduated_landscape | 53% | **70%** | +17pp |
+| simple_additive | 63% | **80%** | +17pp |
+
+Variance inflation roughly doubles the unique evaluations (e.g., needle: 63→159, multigroup: 127→378), confirming that the core problem is exploration — without inflation, TS gets trapped in locally-optimal exhausted subtrees.
+
+#### 11.5.2 TS Beats UCT on Interaction-Heavy Problems
+
+On **multigroup_interaction** (the hardest problem for UCT), TS + TS(g,a) + var_infl achieves **47% optimum rate vs UCT's 23%** — more than double. This is the only problem where TS clearly outperforms UCT.
+
+Why: multigroup_interaction has strong cross-group interaction bonuses (e.g., feature 1 + feature 9 = +12 reward). UCT with `c_uct=0.01` is near-greedy, committing to the first good subtree it finds. TS's posterior-driven exploration naturally samples from multiple high-potential subtrees, increasing the chance of discovering interaction combinations. The posterior captures *reward variance* — if a subtree leads to both high and low rewards depending on downstream choices, TS explores it more because the wide posterior occasionally samples high.
```
TobyBoyne, line 1090: Could this be a problem with `c_uct=0.01`? I don't see anywhere where this is tuned, would increasing it lead to more exploration?

> TS's posterior-driven exploration naturally samples from multiple high-potential subtrees

At least in BO, I have found that TS is actually overexploitative. Maybe the variance inflation is doing a lot more work than I realise?

mcts-report/REPORT.md
```diff
@@ -0,0 +1,2395 @@
...
+
+#### 11.5.3 UCT Dominates on Large Search Spaces
+
+On **large_sparse** (960M combinations), UCT (+rpol) achieves **50% vs TS's best 20%**. On **needle_in_haystack** (5K combinations), UCT achieves **100% vs TS's best 83%**.
+
+The explanation is exploration *efficiency*: UCT's near-greedy search with virtual loss concentrates evaluations on the most promising subtrees and then uses virtual loss to force exploration *within those subtrees* when they exhaust. TS explores more *broadly* — the stochastic sampling sends the search to diverse regions of the tree — but each individual subtree gets fewer evaluations. In large spaces where the number of feasible selections vastly exceeds the budget, UCT's focused exploitation finds the optimum more reliably.
+
+The unique evaluation counts confirm this: UCT evaluates 750 unique selections on large_sparse, while TS + var_infl evaluates 575. Those extra 175 evaluations, concentrated in promising regions, make the difference.
```
TobyBoyne, line 1098: I don't follow this logic. If UCT is more near-greedy/exploitative, then surely it would have fewer unique selections? And if TS is explorative, it would have more? Reading some of these comments, I worry that Claude has seen a benchmark result and is saying anything to justify it, even if it isn't really true/there isn't sufficient evidence to back up the claim.

mcts-report/REPORT.md
```diff
@@ -0,0 +1,2395 @@
...
+
+The unique evaluation chart reveals the core trade-off. UCT configs consistently evaluate more unique selections (455-750 per problem), while TS without variance inflation evaluates far fewer (42-211). Variance inflation partially closes the gap (87-575), but on large_sparse — where coverage matters most — TS still trails.
+
+The implication: TS with variance inflation spends ~25% of its budget on cache hits (re-visiting exhausted subtrees and inflating posteriors), while UCT with virtual loss spends a similar fraction but extracts more value because the deterministic virtual-loss mechanism is more efficient at redirecting search than the stochastic posterior widening.
+
+### 11.8 Summary Bar Chart
+
+![TS vs UCT: Final Best Reward](summary_bar_chart_ts.png)
+
+### 11.9 Adaptive Prior Variance
```
TobyBoyne, line 1182: To me, this whole section looks like it can be fixed by normalization of the rewards. Then we could use a fixed prior variance without problem.

mcts-report/REPORT.md
```diff
@@ -0,0 +1,2395 @@
...
+| graduated_landscape | **80%** | 73% | −7pp |
+| simple_additive | 83% | **87%** | **+4pp TS wins** |
+
+With adaptive prior variance, TS now **wins on 2 of 6 problems** (multigroup_interaction and simple_additive) and is within 7pp on a third (graduated_landscape). The gap on large_sparse has narrowed from 30pp to 17pp. UCT still dominates on needle (perfect 100%) and mixed (77% vs 60%).
+
+### 11.10 Pessimistic Pseudo-Observations
+
+#### 11.10.1 Motivation
+
+Variance inflation (§11.5.1) widens posteriors symmetrically on cache hits — the posterior could sample higher *or* lower — so ~50% of samples from an inflated exhausted node still select it. Pessimistic pseudo-observations provide *asymmetric* downward pressure: on each cache hit, we inject a pseudo-observation at `global_mean - global_std` into every node along the backpropagation path. This shifts the posterior mean downward, actively pushing the algorithm away from exhausted subtrees, analogous to UCT's virtual loss.
```
TobyBoyne, line 1261: Why `global_mean - global_std`? To me, it seems that is *too* punishing, and would push the algorithm away from promising subtrees? Shouldn't it be `leaf_mean - leaf_std`?

mcts-report/REPORT.md
```diff
@@ -0,0 +1,2395 @@
...
+| multigroup | 378 | 395 | 432 | **475** | **479** |
+| needle | 159 | 159 | 258 | **282** | **287** |
+| mixed | 342 | 336 | 353 | **360** | **359** |
+| large_sparse | 575 | 558 | 639 | **671** | **672** |
+| graduated | 102 | 112 | 153 | **175** | **181** |
+| simple_additive | 112 | 127 | 176 | **192** | **202** |
+
+#### 11.11.5 Analysis
+
+**Combined mode achieves the highest exploration efficiency of any config.** On every problem, `comb` evaluates more unique selections than either `var_infl` or `pess` alone. On large_sparse, `comb` reaches 671 unique evaluations vs 575 for `var_infl` and 639 for `pess`. The dual mechanism — decay followed by pessimistic injection — creates stronger pressure to leave exhausted subtrees than either mechanism alone.
```
TobyBoyne, line 1406: Again, I'm a bit confused about this idea of "exploration efficiency". This report is making it out to be a key important metric, but I think you can make deterministic changes to the algorithm to enforce it to 100% evaluate a novel leaf. Why don't we?

mcts-report/REPORT.md
```diff
@@ -0,0 +1,2395 @@
...
+    return max(empirical_var, 1e-8)
+```
+
+Early iterations (few rewards observed) use the fixed fallback, which provides wide priors and broad exploration. As rewards accumulate, the prior tightens to match the actual reward distribution. Newly expanded children then have priors that are appropriately calibrated — wide enough to explore on large-scale problems, tight enough to focus on small-scale ones.
+
+This is the TS analogue of reward normalization: instead of squashing rewards to [0, 1] to match `c_uct`, we scale the prior to match the rewards.
+
+#### 11.12.3 Two-Phase Burn-in with Cheap Evaluations
+
+**Problem**: TS's exploration efficiency gap (575 vs 750 unique evals on large_sparse) exists because each evaluation is expensive (`optimize_acqf` with multi-start L-BFGS), so wasted iterations on cache hits are costly. The cache itself exists because evaluations are expensive and deterministic.
```
TobyBoyne, line 1490: > TS's exploration efficiency gap (575 vs 750 unique evals on large_sparse) exists because each evaluation is expensive

I don't understand this statement.

mcts-report/REPORT.md
```diff
@@ -0,0 +1,2395 @@
...
+| Exploitation (N+1 to end) | Full `optimize_acqf` | On | 1x | Accurate exploitation |
+
+During burn-in, every evaluation is novel (no cache, no cache hits), so the exhausted-subtree problem disappears entirely. TS's posteriors accumulate diverse noisy observations across the combinatorial landscape. At transition, the posteriors already encode which regions of the space are promising, so the expensive budget is concentrated where it matters.
+
+TS is uniquely suited for this because the Bayesian update naturally handles heteroscedastic observations: noisy burn-in values produce wide posteriors (low confidence), and accurate post-burn-in values produce tight posteriors that dominate the mean. No statistic flushing or phase-tracking logic is needed. UCT's running mean cannot distinguish noisy from accurate observations, making a clean transition much harder (§8.5).
+
+The benchmark data suggests the burn-in length should scale with search space size: ~50 iterations for small spaces (graduated_landscape, 375 combinations), ~200 for large (large_sparse, 960M combinations).
+
+#### 11.12.4 Depth-Dependent Cache-Hit Handling
```
TobyBoyne, line 1507: This section introduces a new heuristic to address another heuristic. I'm a bit sceptical.

mcts-report/REPORT.md
```diff
@@ -0,0 +1,2395 @@
...
+
+#### 11.12.6 ~~Normal-Inverse-Gamma Posterior (Proper Conjugate Update)~~ — Implemented
+
+**Implemented and benchmarked in §11.13.** The NIG posterior is a transformative improvement. The best NIG config (NIG + TS(g,a) + vi + apv) achieves 80% on multigroup (vs Normal-TS's 47% and UCT's 23%), 100% on needle and mixed, and 47% on large_sparse (vs UCT's 50%). A single NIG config now matches or exceeds UCT on 5 of 6 problems.
+
+**Original problem statement**: The current TS implementation uses a Normal-Normal conjugate update that treats the reward variance σ² as a known plug-in estimate (`s² = max(sum_sq/n - x̄², 1e-8)`). This is reasonable when n is moderate, but it breaks down at low observation counts — exactly the regime that matters most for exploration:
+
+- With n=1: `s² = max(x²/1 - x², 1e-8) = 1e-8` — variance collapses to the floor. The posterior becomes absurdly tight around a single observation.
+- With n=2: sample variance is based on just 2 points — unreliable.
+- With n=0: we fall back to the prior, which is a Normal distribution.
```
TobyBoyne, line 1537: I might be missing something here, but I don't think the posterior variance is correct here (or in the code). If we are being Bayesian, then when we have one datapoint shouldn't the posterior variance be some function of the prior variance (posterior = prior * likelihood)? I will think about this a bit deeper.

TobyBoyne: Let $\theta=(\mu_{\theta}, \sigma_{\theta})$ be the parameters defining the distribution of the leaf.    
Let the reward $r$ at a leaf be sampled from $p(r \mid \theta) = N(r; \mu_{\theta}, \sigma_{\theta}^2)$.    

If we have a single observation, ie. our dataset $D = {r}$, and a fixed $\sigma_{\theta}$, then    
$p(\mu_{\theta} \mid D) \propto p(D \mid \mu_{\theta}) \cdot p(\mu_{\theta})$    
$=N(r; \mu_{\theta}, \sigma_{\theta}^2) \cdot N(\mu_{\theta}; \mu_0, \sigma_0^2)$    
$=N(\mu_{\theta}; r, \sigma_{\theta}^2) \cdot N(\mu_{\theta}; \mu_0, \sigma_0^2)$    
$=N \left( \mu_{\theta}; \frac{r \sigma_0^2 + \mu_0 \sigma_{\theta}^2}{\sigma_0^2 + \sigma_{\theta}^2}, \frac{\sigma_{\theta}^2 \sigma_0^2}{\sigma_{\theta}^2 + \sigma_0^2} \right)$

So I think the variance should be $\frac{\sigma_{\theta}^2 \sigma_0^2}{\sigma_{\theta}^2 + \sigma_0^2}$. But this is all for a fixed, known $\sigma_{\theta}^2$ - I suppose that's what the Normal Inverse Gamma is for. But I think the existing implementation of Thompson sampling, using this plug-in estimate, isn't correct (it definitely isn't Bayesian since the prior disappears!) Again, I might be missing something, so please correct me if I've misunderstood this.

PR #697, opened 2026-01-20
ContinuousMolecularInput

jduerholt:
## Motivation

In the spirit of the `ContinuousDescriptorInput`, this PR adds a `ContinuousMolecularInput`. The deprecated `MolcularFeature` is removed, as it was not longer supported already in the last releases. A `CategoricalMolecularFeature` should be used instead. In addition, a `MolecularWeightedSumFeature` is introduced for using the `ContinuosMolecularInput`s for surrogate building. 

@LukasHebing: I somehow accidentially added the improvements regarding the setup of the input transform, that we already discussed. I think, for this reason, it makes sense if you review this PR. 

Fine for you?

Yes.

## Test Plan

Unit tests.

### DIFF
--- a/bofire/data_models/molfeatures/molfeatures.py
+++ b/bofire/data_models/molfeatures/molfeatures.py
@@ -43,7 +43,7 @@ class MolFeatures(BaseModel):
     def _get_descriptor_values(self, values: pd.Series) -> pd.DataFrame:
         pass
 
-    def _update_uncorrelated_descriptors(self, molecules: List[str]):
+    def remove_correlated_descriptors(self, molecules: List[str]):
         # Get unique SMILES to avoid redundant calculations
         unique_smiles = pd.Series(molecules)
 
--- a/bofire/utils/torch_tools.py
+++ b/bofire/utils/torch_tools.py
@@ -1122,7 +1122,7 @@ def get_categorical_encoder(
     if isinstance(transform, AnyMolFeatures):
         assert isinstance(feature, CategoricalMolecularInput)
         # filter out the highly-correlated descriptors
-        transform._update_uncorrelated_descriptors(feature.categories)
+        transform.remove_correlated_descriptors(feature.categories)
         encodings = torch.from_numpy(
             feature.to_descriptor_encoding(
                 transform, pd.Series(feature.categories)
--- a/bofire/surrogates/engineered_features.py
+++ b/bofire/surrogates/engineered_features.py
@@ -97,7 +97,7 @@ def map_molecular_weighted_sum_feature(
 
     molecules = [inputs.get_by_key(key).molecule for key in feature.features]  # type: ignore
     # filter out the highly-correlated descriptors
-    feature.molfeatures._update_uncorrelated_descriptors(molecules)
+    feature.molfeatures.remove_correlated_descriptors(molecules)
     descriptors_df = feature.molfeatures.get_descriptor_values(pd.Series(molecules))
     descriptors = torch.tensor(descriptors_df.values, dtype=torch.double)
 
--- a/tests/bofire/data_models/test_molfeatures.py
+++ b/tests/bofire/data_models/test_molfeatures.py
@@ -81,7 +81,7 @@ def test_get_uncorrelated_descriptors(molfeatures):
     initial_descriptor_values = molfeatures.get_descriptor_values(pd.Series(molecules))
 
     # Run select_mordred_descriptors
-    new_names = molfeatures._update_uncorrelated_descriptors(molecules=molecules)
+    new_names = molfeatures.remove_correlated_descriptors(molecules=molecules)
 
     # Check that the number of descriptors has decreased
     final_descriptor_count = len(molfeatures.get_descriptor_names())


bofire/data_models/features/engineered_feature.py
```diff
@@ -108,3 +110,32 @@ def validate_features(self, inputs: "Inputs"):
...
+
+    Args:
+        features: The molecular features to be used to compute the weighted sum.
+        molfeatures: The molecular feature descriptor specification.
+        keep_features: Whether to keep the original features after
+            creating the engineered feature in surrogate creation.
+    """
+
+    type: Literal["MolecularWeightedSumFeature"] = "MolecularWeightedSumFeature"
+    molfeatures: MordredDescriptors
```
LukasHebing: Would it make sense, to use an abstract general descriptor input (possibly custom ones?) and not limit this to Mordred?

jduerholt: Yeah, I changed this.

bofire/data_models/molfeatures/molfeatures.py
LukasHebing, line 17: The abstract class for Molfeatures is way better, than what we had before :)

bofire/data_models/molfeatures/molfeatures.py
```diff
@@ -139,24 +179,87 @@ def validate_descriptors(cls, descriptors):
...
+        if combined.columns.duplicated().any():
+            duplicate_names = combined.columns[combined.columns.duplicated()].tolist()
+            raise ValueError(
+                "Duplicate descriptor names found in CompositeMolFeatures: "
+                f"{sorted(set(duplicate_names))}",
+            )
+        return combined
+
+
+def FingerprintsFragments(
```
LukasHebing, line 241: This is a good way of combining the different Molfeatures, with a dedicated mixing class.

bofire/utils/torch_tools.py
```diff
@@ -1121,6 +1121,8 @@ def get_categorical_encoder(
     """Get the categorical transformer for a given feature."""
     if isinstance(transform, AnyMolFeatures):
         assert isinstance(feature, CategoricalMolecularInput)
+        # filter out the highly-correlated descriptors
+        transform._update_uncorrelated_descriptors(feature.categories)
```
LukasHebing: What I don't like here, is that this method call is kind of "hidden", in the "apply-for-BO-parts" of the code. Wouldn't it be better to call the uncorrelated_features method in a model_validator, and then pass the data-class to the BO part?
Otherwise, the results of the feature selection are not visible.

I see, that this is difficult, because the data-model part must work also without rdkit. So, if that's the only way to do this, we can do it like this.

bofire/data_models/molfeatures/molfeatures.py
```diff
@@ -18,39 +18,109 @@ class MolFeatures(BaseModel):
...
+        all_descriptor_values = self._get_descriptor_values(values)
+        if self._descriptors is not None:
+            return all_descriptor_values[self._descriptors].copy()
+        return all_descriptor_values
+
+    @abstractmethod
+    def _get_descriptor_values(self, values: pd.Series) -> pd.DataFrame:
         pass
 
+    def _update_uncorrelated_descriptors(self, molecules: List[str]):
```
LukasHebing: - Naming: how about `remove_correlated_descriptors', because the descriptors itself are not updated, only the selection.
- Can we also make this a visible function without underscore? Then the user can use it after generating the data-model, and inspect the results. I think, modelers would be interested in the results.

docs/tutorials/advanced_examples/octane_number.qmd
```diff
@@ -0,0 +1,79 @@
...
+                    molfeatures=MordredDescriptors(descriptors=mordred_names,ignore_3D=False, correlation_cutoff=0.9),
+                    keep_features=False
+                )
+            ]
+        )
+)
+
+print("Number of molecular features before correlation filtering: ", len(surrogate_data.engineered_features[0].molfeatures.get_descriptor_names()))
+surrogate = surrogates.map(surrogate_data)
+cv_train, cv_test, _ = surrogate.cross_validate(df_experiments, folds=10 if not SMOKE_TEST else 3)
```
LukasHebing, line 72: I didn't know, that we had a cross_validate method :)

LukasHebing (approved):

docs/tutorials/advanced_examples/octane_number.qmd
jduerholt: It is super powerful, I am using it all the time.

bofire/utils/torch_tools.py
jduerholt: Yeah, I also have the feeling that the solution is not ideal, but all other solutions that I was thinking off had even more drawbacks, so let us keep it and maybe discuss at the next hackathon if we can come up with something better here.

bofire/data_models/molfeatures/molfeatures.py
jduerholt: Yep, will do!

jduerholt: Done, I wait for the tests and then merge it in.

PR #701, opened 2026-01-23
pre compute tanimote kernel distances

LukasHebing:
Tanimoto-similarities are computed upfront, using rdkit functions

- This avoids re-computation during model training and evaluation and exsessive memory usage
- Computation with rdkit is significantly faster (sparse bit vector vs. full float64 torch arrays may explain the difference)
- computation is done in kernel, cov-tensors are passed to the strategy via `re_init_kwargs`
- The `re_init_kwargs` are extended to the Kernel `map` function.

Performance tests:
On a private benchmark with >1k molecules, this computation only needed ~10% of the computation time. 

The old computation also lead to memory-errors, when many molecules were added as training data points leading to crashes.

### DIFF
--- a/bofire/utils/cheminformatics.py
+++ b/bofire/utils/cheminformatics.py
@@ -8,7 +8,6 @@ import pandas as pd
 try:
     from rdkit import RDLogger
     from rdkit.Chem import AllChem, Descriptors, MolFromSmiles
-    from rdkit.DataStructs import FingerprintSimilarity
 
     lg = RDLogger.logger()
     lg.setLevel(RDLogger.CRITICAL)
--- a/tests/bofire/surrogates/test_re_init_kwargs.py
+++ b/tests/bofire/surrogates/test_re_init_kwargs.py
@@ -6,6 +6,7 @@ import torch
 
 from bofire.benchmarks.api import Himmelblau
 from bofire.data_models.domain import api as domain_api
+from bofire.data_models.features import api as features_api
 from bofire.data_models.molfeatures.api import FingerprintsFragments
 from bofire.data_models.surrogates.api import SingleTaskGPSurrogate, TanimotoGPSurrogate
 from bofire.surrogates.api import map


bofire/data_models/kernels/molecular.py
```diff
@@ -10,3 +14,26 @@ class MolecularKernel(FeatureSpecificKernel):
 class TanimotoKernel(MolecularKernel):
     type: Literal["TanimotoKernel"] = "TanimotoKernel"
     ard: bool = True
+    _pre_compute_similarities: bool = False
+
+    # private attributes, for pre-computation of similarities: will be overridden by tanimoto_gp, or auto-computed
+    _fingerprint_settings_for_similarities: Optional[dict[str, Fingerprints]] = None
+    _molecular_inputs: Optional[list[CategoricalMolecularInput]] = None
```
jduerholt: Hmm, I a bit puzzled here, why is it a list of `CategoricalMolecularInput`s, the kernel should always acts on one CategoricalMolecularInput or what are you using this for?

bofire/data_models/surrogates/tanimoto_gp.py
```diff
@@ -33,6 +38,46 @@ class TanimotoGPSurrogate(TrainableBotorchSurrogate):
...
+        if isinstance(self.kernel, ScaleKernel):
+            base_kernel = self.kernel.base_kernel
+            if isinstance(base_kernel, TanimotoKernel):
+                molecular_inputs: list[CategoricalMolecularInput] = self.inputs.get(
+                    includes=CategoricalMolecularInput,
+                    exact=False,
+                ).features  # type: ignore
+                base_kernel._molecular_inputs = molecular_inputs  # type: ignore
+
+                # move fingerprint data model fro categorical encodings to kernel-specs
```
jduerholt, line 110:
```suggestion
                # move fingerprint data model for categorical encodings to kernel-specs
```

bofire/kernels/fingerprint_kernels/tanimoto_kernel.py
```diff
@@ -61,11 +67,107 @@ class TanimotoKernel(BitKernel):
...
+
+            cov = torch.zeros((*batch_shape, n1, n2), **tkwargs)
+
+            # Sum contributions for each feature index along the last dim
+            for idx, inp_ in enumerate(self.molecular_inputs):
+                D = self.sim_matrices[
+                    inp_.key
+                ]  # [Ni, Ni], precomputed distances for feature idx
+
+                # Gather integer indices for this feature from x1 and x2 (keep batch dims)
```
jduerholt, line 137: What do you mean with this feature? From my understanding x1 and x2 are sets matrices of molecules encoded by fingerprints, or?

So what is happening here? Are you trying to get the indices of the moleucles of x1 and x2 in the precomputed distance matrix?

bofire/data_models/kernels/molecular.py
LukasHebing: The kernel can also handle multiple `CategoricalMolecularInput`. Distances in more dimensions are just added. So, you can have a single kernel for multiple inputs (say e.g. ligand-smiles, solvent-smiles)

bofire/kernels/fingerprint_kernels/tanimoto_kernel.py
LukasHebing: yes, these are the indeces. 
In this setup we only encode x1, and x2 as integers (this is because, we remove the preprocessing step into fingerprints in the data model for the tanimoto kernel, and pass the fingerprint info to the kernel data-model:

```python
                # move fingerprint data model fro categorical encodings to kernel-specs
                base_kernel._fingerprint_settings_for_similarities = {}
                for inp_ in molecular_inputs:
                    if inp_.key in list(self.categorical_encodings):
                        assert isinstance(
                            self.categorical_encodings[inp_.key], Fingerprints
                        ), (
                            f"Categorical encoding for input {inp_.key} must be a Fingerprint. "
                            f"Found {type(self.categorical_encodings[inp_.key])}"
                        )
                        fingerprint: Fingerprints = self.categorical_encodings.pop(
                            inp_.key
                        )  # type: ignore
                        base_kernel._fingerprint_settings_for_similarities[inp_.key] = (
                            fingerprint  # type: ignore
                        )
```


In the initial kernel setup, fingerprints and all mutual similarities are computed and stored as a large matrix.

In the `forward` method, only the respective rows and cols of this matrix are selected.

LukasHebing:
@jduerholt Thanks for the review :)

One new problem came with merge. The automatic reduction of the correlated fingerprint vectors is happening in the computation which was used so far, but not in the pre-computed. Because the pre-computed tanimoto sims used the rdkit bit-vectors instead of the parsed numpy/pandas arrays (this is why they are way faster).

However, it seems that the tanimoto calculation (after this changes) is only responsible for <2% of the computation time, so this is not so relevant anymore. I try to change this and use the same tensor-based similarity computation, which also includes the removal of correlated fingerprint features.

@jduerholt: This may take some days. I will let you know when this is ready.

jduerholt:
Just as a comment, you can turn this behavior off, via setting `remove_correlated_features` to `False`. One could also enforce this in the future for tanimoto ...

bofire/data_models/kernels/molecular.py
jduerholt: nice, was this also possible before? Yes, or?

bofire/kernels/fingerprint_kernels/tanimoto_kernel.py
jduerholt: Ah ok, this means that in case that one wants a TanimotoGP with precompute, one has to remove the `input_transform_specs` and one just passes the integers?

LukasHebing: No, the specification is the same as before. We only have to set `pre_compute_similarities` to true. In the tanimoto-gp data-model, the categorical-specs (fingerprint calculation settings) are then passed to the kernel, and removed from the `categorical_encodings` in the model-validator automatically.

bofire/data_models/kernels/molecular.py
```diff
@@ -10,3 +14,26 @@ class MolecularKernel(FeatureSpecificKernel):
 class TanimotoKernel(MolecularKernel):
     type: Literal["TanimotoKernel"] = "TanimotoKernel"
     ard: bool = True
+    _pre_compute_similarities: bool = False
+
+    # private attributes, for pre-computation of similarities: will be overridden by tanimoto_gp, or auto-computed
```
jduerholt, line 19: Can you give this attributes the `PrivateAttr` type so that they are not used in serializatione etc as here: https://github.com/experimental-design/bofire/blob/a9973b18e98e780f05bf81cb1ab05bae1f4492f5/bofire/data_models/molfeatures/molfeatures.py#L23

bofire/data_models/kernels/molecular.py
```diff
@@ -10,3 +14,26 @@ class MolecularKernel(FeatureSpecificKernel):
...
+    _molecular_inputs: Optional[list[CategoricalMolecularInput]] = None
+
+    @model_validator(mode="after")
+    def compute_mutual_tanimoto_distances(self):
+        if not self._pre_compute_similarities:
+            return self
+
+        assert (
+            self._molecular_inputs is not None
+        ), "need molecular inputs ofr pre-computed distances"
```
jduerholt: What do you mean with `ofr` in the error string? Is this a typo?

bofire/data_models/surrogates/tanimoto_gp.py
```diff
@@ -33,6 +39,11 @@ class TanimotoGPSurrogate(TrainableBotorchSurrogate):
     noise_prior: AnyPrior = Field(default_factory=lambda: THREESIX_NOISE_PRIOR())
     scaler: ScalerEnum = ScalerEnum.IDENTITY
 
+    _skip_encodings_validation: bool = (
```
jduerholt, line 40: Is this a private thingy or can it be set by the user?

bofire/surrogates/botorch.py
```diff
@@ -100,17 +100,36 @@ def __init__(
...
     @property
     def re_init_kwargs(self) -> dict:
-        return {"input_transform": self._input_transform}
+        return {
+            "input_transform": self._input_transform,
+            "kernel_kwargs": self._kernel_re_init_kwargs,
+        }
+
+    @property
+    def _kernel_re_init_kwargs(self) -> dict[type, dict]:
```
jduerholt, line 128: I am not sure if this is generic enough, we also have kernel products and kernel sums, what happens if we are using a tanimoto kernel for one categorical molecular kernel and multiply this kernel with some RBF for the other features?

bofire/utils/cheminformatics.py
```diff
@@ -150,3 +153,16 @@ def smiles2fragments_fingerprints(
...
+
+def mutual_tanimoto_similarities(
+    smiles: list[str],
+    bond_radius: int = 5,
+    n_bits: int = 2048,
+) -> list[float]:
+    fingerprints = smiles2fingerprints(
+        smiles, bond_radius, n_bits, output_as_rdkit_bitvector=True
+    )
+
```
jduerholt: Can you provide a docstrin here?

jduerholt: Is this method used anywhere? No, or? And should it not also be applicable to fragments?

jduerholt: If we actually need it ...

tests/bofire/surrogates/test_tanimoto_gp.py
```diff
@@ -0,0 +1,151 @@
...
+            rtol=1e-3,
+            atol=1e-3,
+        )
+        assert np.allclose(
+            pred1.variance.detach().numpy(),
+            pred2.variance.detach().numpy(),
+            rtol=1e-3,
+            atol=1e-3,
+        )
+        assert np.allclose(mean1, mean2, rtol=1e-3, atol=1e-3)
```
jduerholt: the rtol and atol are quite loose, what is the reason for this?

jduerholt (commented):
Hi @LukasHebing,

thanks for it, looks good to me, only some minor remarks and questions.

In the end this is really an optimization which somehow violates some of the design principles behind BoFire (do not get me wrong, I like it), but for this reason, I think one has to live with more complicated code for this case ...

Best,

Johannes

bofire/data_models/kernels/molecular.py
LukasHebing: typo :)

bofire/utils/cheminformatics.py
LukasHebing: This is the fastest way of computing all the fingerprints, but it doesn't use the other pre-processing steps (filtering based on threshold, etc.). However, pre-computing them with the "old" torch function is fast enough.
We can delete it now.

tests/bofire/surrogates/test_tanimoto_gp.py
LukasHebing: That was also due to differences from `rdkit` and `torch` based calculations. Not necessary anymore.

bofire/data_models/surrogates/tanimoto_gp.py
LukasHebing: That should also be private. Serialization / deserialization was getting more complex.

PR #705, opened 2026-01-28
Support Multi-Output, Multi-Fidelity optimization

TobyBoyne:
## Motivation

The current support for multi-fidelity (MF) in BoFire is limited to the `MultiFidelityStrategy`, which is limited to (1) discrete fidelities, and (2) single output. I would like to be able to optimize in a multi-output MF setting, with a continuous fidelity parameter (eg. "simulation resolution"). This is implemented in botorch with the `qMultiFidelityHypervolumeKnowledgeGradient` AF.

I had a few questions I wanted to ask before I dived too deep into this:
1. Should this be a new `Strategy`? In my opinion, it should work directly with `MoboStrategy`, but this might require some changes.
2. Should `SingleTaskMultiFidelityGP` be a new surrogate, based on [this](https://botorch.org/docs/tutorials/Multi_objective_multi_fidelity_BO/)? Or should it just be a `SingleTaskGP` with the custom kernel?
3. We currently use `botorch.acquisition.get_acquisition_function` to build the AF objects. `qMFHVKG` is not available through that function. What do you think about creating a `get_acquisition_function_custom`, which we can register with new AFs so that we can be more active in adding AFs to the list?
4. How do you feel about splitting `TaskInput` into `ContinuousTaskInput` and `CategoricalTaskInput`? Will this break the API for existing users of MF models, and how big of an issue is that?

For (1) and (2) above, I am conscious about polluting BoFire with too many classes. But it might prove difficult to make the existing classes flexible enough.

There is an old PR, #533, which I will close, But maybe there is some good code in there that can be salvaged.

For reference, see https://botorch.org/docs/tutorials/Multi_objective_multi_fidelity_BO/

@jduerholt 

Yes

## Test Plan

I will write tests to cover all new code, once the approach has been decided.


TobyBoyne:
One issue with just adding another acqf is that, when building the qMFHVKG acqf you must first optimize the current posterior ([eg. in this tutorial](https://botorch.org/docs/tutorials/Multi_objective_multi_fidelity_BO/#define-helper-functions-for-mf-hvkg)). This then means you need to pass in some extra info. I still want to avoid creating many many new classes, but maybe it's required to make a `qMultiFidelityHypervolumeKnowledgeGradientStrategy`.

Jimbo994:
Hi guys,

Sorry for my late reply to the recent mention. I was playing around with a dataset for which a `MultiTaskSurrogate` on correlated outputs performed much better than using `SingleTaskGP`s. In addition, there was an output which could be modelled by a `LinearDeterministicSurrogate`. I then tried to set up a MoboStrategy using the `MultiTaskSurrogate`  on one of the tasks and the `LinearDeterministicSurrogate`, but this resulted in the following ValidationError in version 0.2.2.

[https://github.com/experimental-design/bofire/blob/v0.2.2/bofire/data_models/surrogates/botorch_surrogates.py#L144](url)

I actually checked just now in 0.3.2, and here the ValidationError was removed, and the setup seems to work! So it turns out there was no issue @jduerholt.

TLDR;
I was interested in a Multi-output, Multi-objective optimization comprised of a Multi-Task GP acting on one task and a LinearDeterministicSurrogate on another.

I do think Multi-Output, Multi-Fidelity that @TobyBoyne describes is interesting and useful.

TobyBoyne:
With #719, the surrogate now works properly, and the thing runs successfully :)

I've answered some of my own questions in implementing this strategy. My remaining questions for this PR before I take it out of draft are:
1. Do you agree with splitting `TaskInput` into `ContinuousTaskInput` and `CategoricalTaskInput`? If so, we may also want to think about assigning a `cost` to each of the fidelities, so that the acqf can weight the improvement of evaluating a fidelity against how expensive it is.
2. Would you like a quarto tutorial for this? I don't think it will be particularly interesting (it will look identical to most other BO tutorials), but still happy to make one to have a point of reference.

R-M-Lee:
Associating a cost with each fidelity makes a lot of sense. Assuming that the costs are in [0.0, 1.0], maybe the default could be `linspace(0.5, 1.0, N_fidelities)`. Or maybe logspace depending on what gives more plausible and realistic results. Until now we just have an ordering of the fidelities and I think that this is, in general, not enough.

And a notebook-style tutorial is also a very good idea in my opinion, even if you don't think it's very different from some existing tutorials. It can make it a lot easier for people to get started. Great work Toby

bofire/data_models/features/task.py
```diff
@@ -26,3 +33,23 @@ def validate_fidelities(self):
...
+    type: Any
+
+
+class AffineFidelityCostModel(CostModel):
+    type: Literal["AffineFidelityCostModel"] = "AffineFidelityCostModel"
+    weight: float
+    fixed_cost: float
+
+
+class ContinuousTaskInput(TaskInput, ContinuousInput):
```
jduerholt, line 45: A few questions here to the `ContinuousTaskInput` and the `AffineFidelityModel`:

Why do we need this ContinuousTaskInput, for me it is not bearing any additional information beyond the ContinuousInput if we would do the following two things:

- Mark which is the task feature in the surrogate, but ofc we can also have a class for doing this automatically. I do not have a strong opionin here.
- I find the AffineFidelityCostModel is a bit of double work for me, we have already the `LinearDeterministicSurrogate`, why not just using this one?

bofire/strategies/predictives/multi_fidelity_knowledge_gradient.py
```diff
@@ -0,0 +1,282 @@
...
+
+    def _get_objective(self) -> GenericMCMultiOutputObjective:
+        assert self.experiments is not None
+        objective = get_multiobjective_objective(
+            outputs=self.domain.outputs,
+            experiments=self.experiments,
+        )
+        return GenericMCMultiOutputObjective(objective=objective)
+
+    def get_adjusted_refpoint(self) -> List[float]:
```
jduerholt: Is this a copy of the code in mobo?

jduerholt (commented):
Hi @TobyBoyne,

thanks for this! Really nice! Does this only work for continuous task inputs or also categorical ones? 

I just skimmed it and had a few questions, especially with respect to the CostModel. I am wondering if we need new data models for this as we have it in principle already.

Best,

Johannes

bofire/data_models/features/task.py
TobyBoyne: > Mark which is the task feature in the surrogate

We can do this, but I think this is a property of the domain and a user should be able to express it without needing to define a surrogate/strategy. Also, you may want several task features (representing different axes that you can change fidelity - experiment duration, resolution, etc). Could the same argument not be made for the original `TaskInput`?

> the AffineFidelityCostModel is a bit of double work

I will make this switch! I was just trying to reflect how BoTorch represents it, but I like this approach too.

bofire/strategies/predictives/multi_fidelity_knowledge_gradient.py
TobyBoyne: Yes, it is. Technically we shouldn't inherit from MoboStrategy, since the types of `acquisition_function` don't overlap. `MoboStrategy.acquisition_function` has type `AnyMultiObjectiveAcquisitionFunction`, whereas my strategy only works with the `qMFHVKG` acquisition function.

I can inherit from Mobo and just `# type: ignore` the error raised about the acquisition function?

jduerholt: Maybe, make first the switch regarding the cost model and then we can think about this, maybe we can also share the code via a mixin class or so ...

bofire/data_models/features/task.py
TobyBoyne: Hmm, it is not trivial to add the `LinearDeterministicSurrogate` to the `TaskInput`, since the surrogate requires an `Inputs` object. So you end up with some infinite loop like:
```python
task_input = ContinuousTaskInput(
    key="task", 
    bounds=(0, 1), 
    fidelity_cost=LinearDeterminsticSurrogate(
        inputs=Inputs(features=[
            ContinuousTaskInput(key="task", ...)
        ]
    )
)
```

I can add the cost model to the surrogate, but again I think this leads to a weird separation, where the cost should be a part of the domain and not a part of the surrogate. What do you think @jduerholt?

jduerholt: I would not add it to the ContinuousTaskInput at all, just define it in the surrogate specs, and use this model then in the acqf, what do you think?

jduerholt: And check in the valdidator of the strategy if you have such a model, if not, error out?

TobyBoyne: It sounds like it isn't really feasible to have the cost model as a part of the `TaskInput`, even though I think it might make more sense (I think I had to make a similar concession with the conditional features). In that case, I'll do what you've described and define it in the surrogate specs.

jduerholt: Ah sorry, i do not want to appear so pedantic..., we could go with your first version. But one question: i had a quick look to the botorch implementation,  for me it seems that if you have two task Features with costs, you still go with one deterministic model, or?

jduerholt: If yes, the way over the surrogates would be cleaner or?

TobyBoyne: The botorch model lets you provide a dictionary of `{index: weight}`, which supports several different task features. In the last hr I've tried doing it with surrogates and I have basically just been using them as a container for this kind of a dictionary.

The only area for confusion is if we have many tasks, then each task will have its own `fixed_cost`, which is a bit meaningless. Really, we want the same interface as `LinearDeterministicModel`, where each task has some weight, but then there is a shared single `fixed_cost` across all tasks. So for that reason, I'll stick with the surrogates approach for now and see how it pans out.

TobyBoyne: Sorry I keep going back and forth on this. An issue I'm now running into with the surrogates is that you must create a new `Output` (since a Surrogate can't have `len(outputs)==0`), but then this output won't be in `Strategy.domain.outputs` which will also cause issues. So I then need to create a new output, which leads to some confusing behaviour behind the scenes.

I'm going to switch back to defining the costs on the input. Once I've done that, we can have another look to see if it makes sense. Sorry again for the indecision!

jduerholt: Okay, one question regarding this? why not putting the output also in the domain, somehow the cost is an output or? As it is a function of the inputs, and it used in optimization.

TobyBoyne: Hm that could work. We would then have to require that the cost has `objective=None`, and would we require a specific hardcoded name for that output? Or maybe it can be an argument to the strategy, something like `fidelity_cost_output_key`?

Would you like to have a call about this today, or shall I just try to implement it and see how it looks? I'm conscious I keep doing and undoing work so don't want to waste your time with reviewing code that I'll maybe change again!

jduerholt: This sounds like a good plan, forcing objective=None and map it in the strategy spec! I like it! For me, it is fine if you just go ahead. Put you can ping me anytime if you need a small hint.

TobyBoyne: Okay, will do! Thanks for all the help :)

TobyBoyne:
I've addressed the previous comment in two ways:

1. Can't fix discrete inputs -> removed `DiscreteTaskInput`s entirely. They should work, but since we can't fix discrete features, we needed to replace them with continuous features, which changed the order of inputs. This is a TODO, since if the handling of discrete features is ever refactored then this shouldn't be much change.
2. Cost output broke the botorch surrogate -> seperated cost model to a separate attribute. This just cleaned up a lot of the handling, removing some messy code that was used to handle the cost model simultaneously with the GP surrogate.

This PR is now ready for review (after I fix some small breaking tests)

bertiqwerty:
Reading Toby's last message, this looks as if it was waiting for review. @jduerholt, since you are entered as reviewer, do you want to review this? Or should I do it?

jduerholt:
Ah, totally overlooked this. Yeah, I will handle it. @TobyBoyne: can you get main merged in to resolve the conflicts? I will then review! Sorry!

TobyBoyne:
Merged @jduerholt :) I will check back here to make sure the tests are passing!

TobyBoyne:
Tests are failing but I think those are related to this (https://github.com/meta-pytorch/botorch/pull/3291) change to how fully bayesian models are exposed by BoTorch. The tests seem to be working on my machine at least!

jduerholt:
> Tests are failing but I think those are related to this ([meta-pytorch/botorch#3291](https://github.com/meta-pytorch/botorch/pull/3291)) change to how fully bayesian models are exposed by BoTorch. The tests seem to be working on my machine at least!

Perfect, I will also ask Max for a new botorch release ;)

bofire/benchmarks/multi.py
```diff
@@ -700,3 +701,63 @@ def _get_base_cost(self, base, mmol_base):
             "TEA": 0.01,
         }
         return float(base_prices[base] * mmol_base)
+
+
+class MOMFBraninCurrin(Benchmark):
```
jduerholt, line 706: I recently build a generic interaface for botorch test functions to not have so much code duplication, have a look here: https://github.com/experimental-design/bofire/blob/f9ffad5ef20765ffbe9c11c25cf9545d82d475aa/bofire/benchmarks/benchmark.py#L273

I would prefer if we extend it to MOMF optimization, because then we can use all kind of botorch MOMF test functions. Should also not be complicated. At some point we also should tidy up the whole thing. 

What do you think?

bofire/data_models/kernels/aggregation.py
jduerholt, line 1: Why are we getting the WedgeKernel stuff in here? This is already merged in another PR, or?

jduerholt: Ah, we did not had it in the aggregations? Then we maybe need to update the register functions ...

bofire/data_models/kernels/api.py
```diff
@@ -21,6 +21,7 @@
     RBFKernel,
     SphericalLinearKernel,
 )
+from bofire.data_models.kernels.fidelity import DownsamplingKernel, FidelityKernel
```
jduerholt, line 24: FidelityKernek is imported but not used?

bofire/data_models/kernels/fidelity.py
jduerholt, line 1: I think we also have to properly register the stuff here, with AnyPrior etc. Maybe we have to build an automatic machine for this. Need some discussion with Claude on this.

bofire/data_models/strategies/predictives/multi_fidelity_knowledge_gradient.py
```diff
@@ -0,0 +1,244 @@
...
+        if (
+            self.domain.inputs.get_keys(ContinuousTaskInput)
+            != fidelity_inputs.get_keys()
+        ):
+            raise ValueError("All fidelities should be included in the cost model.")
+
+        return self
+
+    @staticmethod
+    def _generate_fidelity_cost_model_spec(
```
jduerholt, line 84: We could also simplify it and let the user generate the fidelity cost model and do not provide a default, what do you think?

bofire/data_models/strategies/predictives/multi_fidelity_knowledge_gradient.py
```diff
@@ -0,0 +1,244 @@
...
+
+        for m in self.surrogate_specs.surrogates:
+            _validate_surrogate(m)
+
+        self.acquisition_optimizer.validate_surrogate_specs(self.surrogate_specs)
+
+        return self
+
+    @staticmethod
+    def _generate_surrogate_specs(
```
jduerholt: We have this method _generate_surrogate_specs now already in 3 different ocassions, with always some kind of code duplication. What do you think about having the general machinere just once in BotorchStrategy and only call the spesific stuff, namely the generated default models in the specific ocassions? 

Should be a quick change but leeds to less code duplication.

bofire/kernels/mapper.py
```diff
@@ -474,6 +475,39 @@ def map_SphericalLinearKernel(
...
+            if data_model.offset_prior is not None
+            else None
+        ),
+        offset_constraint=(
+            priors.map(data_model.offset_constraint)
+            if data_model.offset_constraint is not None
+            else None
+        ),
+        power_prior=(
+            priors.map(data_model.power_prior, d=len(active_dims))
```
jduerholt: Why do we need here the acive dims?

bofire/strategies/predictives/acqf_optimization.py
```diff
@@ -455,8 +455,10 @@ def _optimize_acqf_continuous(
             OptimizerEnum.OPTIMIZE_ACQF_MIXED: optimize_acqf_mixed,
             OptimizerEnum.OPTIMIZE_ACQF_MIXED_ALTERNATING: optimize_acqf_mixed_alternating,
         }
+        # TODO: does this break anything?
+        # exclude None since some acqfs (eg. HVKG) do not support generator args
         candidates, acqf_vals = optimizer_mapping[optimizer](
-            **optimizer_input.model_dump()
+            **optimizer_input.model_dump(exclude_none=True)
```
jduerholt, line 461: Why this change?

bofire/strategies/predictives/multi_fidelity.py
```diff
@@ -22,9 +22,10 @@
 
 
 class MultiFidelityStrategy(SoboStrategy):
+    # TODO: rename this to "MultiFidelityVarianceBasedStrategy"
```
jduerholt: Should we just do this? And why would you prefer this name?

bofire/strategies/predictives/multi_fidelity_knowledge_gradient.py
```diff
@@ -0,0 +1,324 @@
...
+)
+from bofire.data_models.surrogates.botorch_surrogates import BotorchSurrogates
+from bofire.data_models.surrogates.deterministic import LinearDeterministicSurrogate
+from bofire.strategies.predictives.botorch import BotorchStrategy
+from bofire.strategies.strategy import make_strategy
+from bofire.utils.multiobjective import get_ref_point_mask, infer_ref_point
+from bofire.utils.torch_tools import get_multiobjective_objective, tkwargs
+
+
+def _map_cost_model_to_botorch(
```
jduerholt, line 42: Hmm, this is nothing else as our linear deterministic model and the FilterFeatures transform cooked in. If we would have the cost model directly in the botorch surrogates specs we could get rid of this.

jduerholt: But I get, that this introduced problems elsewhere.

bofire/data_models/acquisition_functions/cost_aware_utility.py
```diff
@@ -0,0 +1,13 @@
+from typing import Any, Literal
+
+from bofire.data_models.base import BaseModel
+
+
+class CostAwareUtility(BaseModel):
```
jduerholt: Where is this used?

bofire/strategies/predictives/multi_fidelity_knowledge_gradient.py
jduerholt: I would prefer if we do not duplicate the code from the MoboStrategy, either we type ignore errors. We could also only use for the functional implementation the inheritance from MoboStrategy but not for the data_model, or we put all this ref point related stuff into a mixin class that we use both for Mobo or here.

bofire/strategies/predictives/multi_fidelity_knowledge_gradient.py
```diff
@@ -0,0 +1,324 @@
...
+            X_pending=X_pending,
+            mc_samples=self.acquisition_function.n_mc_samples,
+            num_fantasies=self.acquisition_function.num_fantasies,
+            num_pareto=self.acquisition_function.num_pareto,
+            cost_aware_utility=cost_aware_utility,
+            seed=self.seed,
+        )
+        return [acqf]
+
+    def _get_objective(self) -> GenericMCMultiOutputObjective:
```
jduerholt: Also this is a copy from Mobo.

bofire/strategies/predictives/multi_fidelity_knowledge_gradient.py
```diff
@@ -0,0 +1,324 @@
...
+def get_project(
+    target_fidelities: dict[int, float], d: int
+) -> Callable[[Tensor], Tensor]:
+    def project(X):
+        return project_to_target_fidelity(X=X, target_fidelities=target_fidelities, d=d)
+
+    return project
+
+
+def get_acquisition_function_qMFHVKG(
```
jduerholt, line 210: What do you think about putting this code directly here in botorch: 

https://github.com/meta-pytorch/botorch/blob/494652319c0c0233c6c103d64d1655e14660800c/botorch/acquisition/factory.py#L37

bofire/strategies/predictives/multi_fidelity_knowledge_gradient.py
```diff
@@ -0,0 +1,324 @@
...
+        self.ref_point_mask = get_ref_point_mask(self.domain)
+        self.fidelity_cost_model_spec = data_model.fidelity_cost_model_spec
+
+    def _get_acqfs(self, n: PositiveInt) -> List[AcquisitionFunction]:
+        assert self.is_fitted is True, "Model not trained."
+        assert self.experiments is not None, "No experiments available."
+
+        X_train, X_pending = self.get_acqf_input_tensors()
+
+        # We must subset the model here, to remove the deterministic cost model from the
```
jduerholt: Ok, now I also see the point of why not to have it in the real surrogates, this would be getting really complicated here at the end, on how botorch is composing the stuff.

bofire/strategies/predictives/multi_fidelity_knowledge_gradient.py
```diff
@@ -0,0 +1,324 @@
...
+                [
+                    ref_point[feat]
+                    for feat in self.domain.outputs.get_keys_by_objective(
+                        excludes=ConstrainedObjective,
+                    )
+                ],
+            )
+        ).tolist()
+
+    def get_current_value(self, target_fidelities: dict[int, float]):
```
jduerholt, line 138: Can you explain to me what this optimization is doing? What are we acutally optimizing?

bofire/strategies/predictives/multi_fidelity_knowledge_gradient.py
```diff
@@ -0,0 +1,324 @@
...
+        )
+        for key in cost_data_model.coefficients
+    }
+
+    return cost_function, target_fidelities
+
+
+def _get_domain_with_fixed_task_inputs(
+    domain: Domain,
+    target_fidelities: dict[int, float],
```
jduerholt, line 73: Why not pass in the target_fidelitis as dict[str, float], would be cleaner or?

jduerholt: But it is also okay like this.

bofire/strategies/predictives/multi_fidelity_knowledge_gradient.py
```diff
@@ -0,0 +1,324 @@
...
+        num_pareto=num_pareto,
+        sampler=sampler,
+        objective=objective,
+        inner_sampler=inner_sampler,
+        X_pending=X_pending,
+        X_evaluation_mask=None,
+        X_pending_evaluation_mask=None,
+        current_value=current_value,
+        cost_aware_utility=cost_aware_utility,
+        project=get_project(target_fidelities, X_observed.shape[-1]),
```
jduerholt: Why this additonal get_project? we could just hand it in as a callable via functools.partial, or? No need for the wrapper above.

jduerholt (changes requested):
Hi @TobyBoyne,

thank you very much, this looks really impressive and I think is almost done. 

I let a lot of small comments mostly focusing on small stuff. Main things, that I see:

- Code duplication from generate_surrogate_specs, we should further modularize it to not repeat always the boilerplate code.
- Code duplication from MOBO, we should either have a mixin for the mobo functionality or solve in a different way, duplicating this can lead to a lot of errors ...

Just go over it, in case that you do not have time etc. for the code duplication stuff, just tell me. Then I try to fix it ;) 

Best,

Johannes

bofire/benchmarks/multi.py
TobyBoyne: Yes I like that, I will have a look into this!

bofire/data_models/kernels/aggregation.py
TobyBoyne: It was causing an issue because I was combining the `WedgeKernel` with the `DownsamplingKernel`, and getting some issues: as you say, I think it was missing from some of the aggregations. I'm happy to remove it from this PR for now if you want to move the change to another PR?

bofire/data_models/kernels/api.py
TobyBoyne: There used to be something like `AbstractKernel` here that included the `FidelityKernel` type. You can also see that `AggregationKernel`, `FeatureSpecificKernel`, and `MolecularKernel` are all imported but not used. Happy to remove any/all of the above?

bofire/data_models/kernels/fidelity.py
TobyBoyne: Okay - I'll leave this one with you because I'm less familiar with the registration+Pydantic side of things. `test_(de)serialization.py` is passing so I'm not sure what is currently broken that needs to be fixed.

bofire/data_models/strategies/predictives/multi_fidelity_knowledge_gradient.py
TobyBoyne: I like having the default for a few reasons:
1. makes it clearer to the user what a fidelity cost model needs to look like. 
2. reduces the time to "first-working-prototype" for a user, which they can then refine with their own model
3. since the default is encapsulated in its own function, it doesn't really increase the complexity of the class in my opinion.

Do those reasons make sense?

bofire/data_models/strategies/predictives/multi_fidelity_knowledge_gradient.py
TobyBoyne: Yes definitely in favour of this, as long as the combined method is in fact cleaner. Should this be in the current PR or separate work?

bofire/kernels/mapper.py
TobyBoyne: Ah good point - for `DownsamplingKernel`, there is only ever one feature, so we don't need to worry about any kind of scaling. Nice catch!

TobyBoyne: I suppose there is nothing stopping someone from using `DimensionalityScaledLogNormalPrior` as a prior here, which would then cause an error in `priors.mapper`? Shall I just replace all of the `d=len(active_dims)` with `d=1` for the `DownsamplingKernel`?

bofire/strategies/predictives/acqf_optimization.py
TobyBoyne: Because of the `TODO` above. The error was that dumping the model would have a field something like `generator_args=None`. This would then get passed to the `qMultiFidelityHypervolumeKnowledgeGradient`, which doesn't take `generator_args`, and so would raise an error.

With my change, when the model dumps the optimizer config, it removes any fields which are set to `None`. I would guess that this wouldn't have any big impact (I would personally say that it's often unexpected behaviour if passing `None` to a Python function changes its function from the default), but I can't guarantee that.

bofire/strategies/predictives/multi_fidelity.py
TobyBoyne: Yes, happy to just do this but it would be a breaking change to the BoFire API.

The current name, `MultiFidelityStrategy`, to me suggests that it is the parent class of all MF strategies. You would expect any other MF strategy to inherit this one. But really, this class has a very specific strategy based on Kandasamy et al. 2016, which is not a general MF strategy.

bofire/strategies/predictives/multi_fidelity_knowledge_gradient.py
TobyBoyne: Yeah it would be nice to be able to rely on that behaviour, but unfortunately it caused a few issues of its own of handling adding the cost to the `Outputs` which had no neat solution (after many different attempts!) 

Also, one extra detail we need to map the datamodel to botorch's `AffineFidelityCostModel`, not to botorch's `AffineDeterministicModel`. The difference being the output shape of the `forward` method are different, I think. So even if we *did* use the existing BoFire mappers, we would still need to do some surgery to make sure the output shape is correct inside the botorch code.

bofire/data_models/acquisition_functions/cost_aware_utility.py
TobyBoyne: This is dead code - I'll remove it

bofire/strategies/predictives/multi_fidelity_knowledge_gradient.py
TobyBoyne: I am happy with any of the suggestions you have made. I agree that the duplicated code is not the way to go, just wasn't sure what the best way forward would be.

But we should be careful because this isn't just about the type checker being upset. It is currently incorrect to subclass MoboStrategy, since Mobo takes any acqf *except* for MFKG, and the new strategy takes *only* MFKG. The current implementation would break Liskov, so we would need to find a clean way around that.

I'm a bit obsessed with [`Generics`](https://typing.python.org/en/latest/reference/generics.html) so I think I did try to use them to type hint the allowed acqf, but I think it ended up feeling like overkill.

bofire/strategies/predictives/multi_fidelity_knowledge_gradient.py
TobyBoyne: Yes I think that's a good idea. Should I raise an issue, or just go straight in with a PR?

bofire/strategies/predictives/multi_fidelity_knowledge_gradient.py
TobyBoyne: Ah yes that comment is outdated and needs to be removed, but it is a perfect example of some of the subtle issues that came up when including it in the real surrgoates!

bofire/strategies/predictives/multi_fidelity_knowledge_gradient.py
TobyBoyne: (I will write this in the docstring to better explain this).

The thing we are trying to find in `get_current_value` is $\psi^{\*}$ in eq (5) in https://proceedings.mlr.press/v202/daulton23a/daulton23a.pdf.

My understanding is this:
In standard, single-output Knowledge Gradient (KG), the acquisition function is 

$$\alpha\_{t+1} (x) = E[ \mu^{\*}\_{t+1} - \mu^{\*}\_{t} ]$$

where $\mu^{\*}\_t$ is the maximum of the posterior mean under the model at time $t$. Note that this isn't the same as the $f^{\*}$ that is used in NoisyEI, which is only the maximum posterior mean evaluated at the observed points. Given a model trained on data up to time $t$, one must first compute $\mu^{\*}\_{t}$ via an optimization of the posterior mean of the model, then optimize $\alpha\_{t+1}$.

In the multi-output setting, this is exactly the same, except instead of optimizing the posterior mean under the model, we optimize the *hypervolume* of the posterior mean.

It's also worth noting that that $\mu^{\*}\_{t}$ is constant wrt $x$, so it can be [omitted entirely](https://github.com/meta-pytorch/botorch/blob/494652319c0c0233c6c103d64d1655e14660800c/botorch/acquisition/multi_objective/hypervolume_knowledge_gradient.py#L217-L219).

bofire/strategies/predictives/multi_fidelity_knowledge_gradient.py
TobyBoyne: Let me have a think. I wanted to match the BoTorch convention but actually maybe your suggestion is cleaner - I might make this change

bofire/strategies/predictives/multi_fidelity_knowledge_gradient.py
jduerholt: As you want, my latest changes to this function were all accepted, so I think the chances are quite high. But we can also keep it here first and make the change to botorch in parallel.

bofire/data_models/strategies/predictives/multi_fidelity_knowledge_gradient.py
jduerholt: For me it is fine to do it here, but you can also make a seperate PR and merge this one in. As you want.

bofire/data_models/kernels/fidelity.py
jduerholt: Fine for me, before final merge I will make a roundtrip regarding this.

bofire/data_models/kernels/aggregation.py
jduerholt: No, this is fine.

bofire/strategies/predictives/multi_fidelity_knowledge_gradient.py
jduerholt: Should I try to draft a proposal?

TobyBoyne: Yes please, if you're happy to! I think this should probably be a part of this PR, and not a separate PR - do you agree?

jduerholt: I try to draft it tmr, I think I make it part of a seperate PR, and then we can merge in from there.

bofire/strategies/predictives/multi_fidelity_knowledge_gradient.py
TobyBoyne: Decided to keep it as-is.

TobyBoyne:
@jduerholt, thanks for the review! I have made the requested changes. All that is left is a change to the inheritance structure such that the new strategy inherits from `MOBO` - are you still happy to draft that?

I haven't made the change to the benchmark to inherit from `SyntheticBoTorch`. I think that's best left to a separate PR, since the class doesn't support quite a few features for now.

jduerholt:
Hi @TobyBoyne — I just pushed a small follow-up commit (b23bb5df) to tighten a few registration / roundtrip loose ends I spotted while reviewing:

- `TaskInput` and `FidelityKernel` are now properly abstract (`type: Any` + docstring). They were carrying a `Literal` `type` but weren't in any `AnyFeature`/`AnyKernel` union, which made them look instantiable when they aren't.
- Added a valid `ContinuousTaskInput` spec so the strict feature-serialization roundtrip actually exercises the new class.
- Added `MultiFidelityVarianceBasedStrategy` and `MultiFidelityHVKGStrategy` to the `AnyPredictive` union — `MoboStrategy` was there but the MF strategies weren't.
- Added a `validate_ref_point` model validator to `MultiFidelityHVKGStrategy` mirroring `MoboStrategy.validate_ref_point`, so `dict[str, float]` / `None` ref points are normalized into an `ExplicitReferencePoint` before the functional class consumes them. The `assert not isinstance(data_model.ref_point, dict)` in the functional ctor would have tripped without this.

---

The bigger thing I wanted to discuss is the duplication between `MultiFidelityHVKGStrategy` and `MoboStrategy`. The current MFHVKG strategy reimplements `_get_objective`, `get_adjusted_refpoint`, the ref-point bookkeeping in `__init__`, the `is_feature_implemented` / `is_objective_implemented` matrices, and (until this commit) the ref-point normalization. A few directions I'd like your take on:

**Option A — inherit from `MoboStrategy` at both layers, override only what differs.**
- Data model: subclass `MoboStrategy`, narrow `acquisition_function` to `qMFHVKG`, add `fidelity_cost_model_spec`, replace `validate_surrogate_specs`, add the continuous-task-input domain validator. Inherits `validate_ref_point`, `is_feature_implemented`, `is_objective_implemented`, `ref_point` field for free.
- Functional: subclass functional `MoboStrategy`, override `_get_acqfs`. Inherits ref-point bookkeeping in `__init__`, `_get_objective`, `get_adjusted_refpoint`, `ref_point_mask`.
- Pros: smallest diff, biggest dedup win. Cons: claims an "is-a" relationship between MFHVKG and Mobo, which couples future Mobo changes into MFHVKG.

**Option B — extract a `_HypervolumeRefPointMixin`.**
- Mixin holds ref-point fields/init, `_get_objective`, `get_adjusted_refpoint`. Both strategies mix it in.
- Pros: avoids the LSP-style coupling. Cons: stateful mixins with model_validators can be fiddly with Pydantic MRO.

**Option C — composition: pull the ref-point logic into `ExplicitReferencePoint` itself (or a thin resolver).**
- e.g., `ExplicitReferencePoint.normalize(domain, candidate)` and `to_array(domain)`. Validators become one-liners and the helper is unit-testable in isolation.
- Pros: cleanest separation, mirrors how `acquisition_optimizer` already works as a swappable component. Cons: biggest refactor — touches `MoboStrategy` too.

My current preference: **Option A for the functional class** (the duplication there is genuinely line-for-line) combined with **Option C-lite for the data model** (a free helper `_normalize_ref_point(domain, ref_point)` reused by both `validate_ref_point` validators). That avoids forcing MFHVKG to inherit from `MoboStrategy` at the data-model layer (where the `acquisition_function` type really does diverge) while still killing the duplicated normalization.

Tangential thought: `is_feature_implemented` / `is_objective_implemented` could be class attributes on every strategy plus a single shared classmethod that does the lookup — independent of MFHVKG, and would DRY up several other strategies too. Probably out of scope for this PR.

What's your preference?

This was written by my agent ;)

TobyBoyne:
Thanks for committing those change @jduerholt :)

All the options make sense. I definitely agree with using **A for the functional class**. I would personally like to use **A for the data model** as well, if possible. It would be nice if the data model structure reflects the functional model structure as well. For example, `isinstance(MFHVKGStrategy(), MoboStrategy)` should be `True` whether those are data model or functional. I've given an example below of how this can be done without violating LSP.

The option you've given for data models does also make sense. If this seems like the best way forward to you, then I can get behind that. Let me know if you would like me to take this on, or if you'd like to do it.

And for your tangential comment, I think that can definitely be refactored in line with what you've suggested. I actually had some more ideas about that - I will speak to you separately. But yes, outside the scope of this PR.

---
My solution here would be to use Generics (which I think I have said for any problem I have with Python type hinting - I may be over-relying on them...). I've given a toy example below. The key change here is that you *can* narrow the type of `acqf` with Generics, since `MFHVKGStrategy` is now a subclass of `MoboStrategy[MFHVKGAcqf]`, which is already narrowed!

(To be clear, all of the redefinitions of `Acqf` are just for demonstration below, so that the script is self-contained. The only important change is the `AcqfT` type variable, and the Generic subclassing.)

```python
from pydantic import BaseModel
from typing import Generic, TypeVar

class Acqf(BaseModel): pass
class LogEIAcqf(Acqf): pass
class UCBAcqf(Acqf): pass
class MFHVKGAcqf(Acqf): pass

AnyAcqf = LogEIAcqf | UCBAcqf | MFHVKGAcqf

AcqfT = TypeVar("AcqfT", bound=AnyAcqf)

class MoboStrategy(BaseModel, Generic[AcqfT]):
    acqf: AcqfT

class MFHVKGStrategy(MoboStrategy[MFHVKGAcqf]):
    pass


strategy1 = MFHVKGStrategy(acqf=MFHVKGAcqf())  # this is fine
strategy2 = MoboStrategy(acqf=MFHVKGAcqf())    # this is also fine
strategy3 = MFHVKGStrategy(acqf=LogEIAcqf())   # this raises an error!
```

jduerholt:
Oh, this is really nice with the Generics. For me it is fine, if you would go behind it and unify MOBO and MFHVKGStrategy, I would love it. Let us start with the functional model and then do it with the generics maybe on the data model. But I think, we should merge this is in now, and do it in a seperate, one or?

jduerholt:
So, would this be fine for you?

TobyBoyne:
I have changed the inheritance so that we now inherit from Mobo. I ended up not doing any of the Generic things because my type checker is already complaining about "Variable not allowed in type expression" in `MoboStrategy.acquisition_function`. So the type checker is just inferring all the acquisition functions as `unknown`, which isn't very helpful anyway.

I have tried to create an instance of the MFHVKG strategy with qLogEI acquisition function, and Pydantic correctly raises an error. So:
1. Pydantic is fine with the current state, and works correctly
2. Pyright is already unhappy with the variable annotations introduced with the `@register`, so the generics won't make it any better.

Maybe we can come back and add the generics at some point, but for now I think this should be good to merge?

jduerholt:
Looks good to me, can you have a look to the failing test, there is something with the make method.

TobyBoyne:
This should fix it. There was a mismatch between `dict` and `typing.Dict`.

Issue #700, opened 2026-01-22
[Bug]: Pandas 3.0 and Formulaic


jduerholt:
### What happened?

The formulaic version that we are using is rather old, in addition there was a new pandas release yesterday. Using this pandas version with our current implementation for DoE creates problems.

@Osburg: could you have a look there?

### Please provide a minimal, reproducible example of the unexpected behavior.

Failing tests.

### Please paste any relevant traceback/logs produced by the example provided.

```shell

```

### BoFire Version

recent main

PR #707, opened 2026-02-02
Implement Pandas 3.0

jduerholt:
## Motivation

Allow for pandas 3.0. Closes https://github.com/experimental-design/bofire/issues/700.

Y.

## Test Plan

Unit tests.


jduerholt:
Hi @TobyBoyne,

switching to pandas 3.0 raised an error in the test suite regarding the multi fidelity strategy (among others, which I try to fix in this PR). I fixed the error, but I am wondering how this could ever work before. 

Can you have a look at this?

Best,

Johannes

bofire/strategies/predictives/multi_fidelity.py
```diff
@@ -93,10 +93,8 @@ def _select_fidelity_and_get_predict(self, X: pd.DataFrame) -> pd.DataFrame:  #
             fidelity_threshold = self.fidelity_thresholds[m] * fidelity_threshold_scale
 
             X_fid = X.assign(**{self.task_feature_key: fidelity_name})
-            transformed = self.domain.inputs.transform(
-                experiments=X_fid, specs=self.input_preprocessing_specs
-            )
-            pred = self.predict(transformed)
+
+            pred = self.predict(X_fid)
```
jduerholt, line 107: @TobyBoyne: I mean this. `self.predict` is expecting non-transformed input, as it is doing the `self.domain.inputs.transform` itself. So why was this working before? Maybe some lucky pandas type conversion? Can you have a look?

TobyBoyne: I'm on holiday this week so I can't have a proper look until next week. Is it maybe that the transformed space is the same as the untransformed space (eg if the bounds of the input is [0, 1]^d?

Sorry I can't be more helpful, happy to have a look next week when I'm back at my laptop.

jduerholt: Fine, I try to fix the rest and then we have a look next week. Enjoy your holidays!

TobyBoyne: TL;DR: this was a bug, worked fine in old versions but your fix is correct! Thanks for spotting :)

Just had a look, and I think it is because the behaviour of applying `self.domain.inputs.transform` twice has changed.

In `pandas<3.0`, applying the transform twice did nothing on the second time. In `pandas>=3.0`, applying it the second time raises an error.

Digging a bit deeper, it's because `Series.__getitem__` used to assume an indexer was an integer-indexer if the value wasn't found in the index. But they deprecated that behaviour in https://github.com/pandas-dev/pandas/issues/50617. In Pandas 3, this raises an issue because the key isn't found, and isn't silently changed to an integer-indexer.

Minimum reproduction below:
```python
import pandas as pd

categories = ["task_0", "task_1", "task_2"]
enc = pd.Series(range(len(categories)), index=list(categories))
values = pd.Series([2], index=[0])
print(enc[values].item())
# gives `2` on old pandas, and raises `KeyError` on new pandas.
```

pyproject.toml
```diff
@@ -26,11 +26,11 @@ readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.10.0"
 dependencies = [
     "numpy",
-    "pandas<=2.3.3",
+    "pandas",
```
TobyBoyne, line 29: Is it worth still pinning `pandas` to avoid a breaking change in the future? Maybe `"pandas<3.1"`?

jduerholt: I am not a fan of pinning, we will see when it breaks ...

PR #709, opened 2026-02-04
Added two output transforms: Log and ChainedOutputTransform

KislayaRavi:
## Motivation

This PR adds support for the Log output transform and Chained Output Transform (chaining Log and Standardized output transform) from BoTorch to BoFire's surrogate models. 
Log transformation is ideal for optimization problems where:
- Targets are strictly positive,
- Values span multiple orders of magnitude,
- The underlying distribution is log-normal.

Additionally, made linting corrections in other files too.

Yes

## Test Plan

1. Added new benchmark (Positive Himmelblau) by adding a small number(1e-8) to the output of Himmelblau function ensuring objective is always positive. This is added for Log transformation, because the input should always be positive.
2. Added new test cases in ```test_torch_models.py```, ```test_mlp.py```, ```test_map_saas.py```, ```test_gps.py``` and ```test_random_forest.py```.

### DIFF
--- a/bofire/surrogates/map_saas.py
+++ b/bofire/surrogates/map_saas.py
@@ -9,12 +9,7 @@ from botorch.models.map_saas import (
     EnsembleMapSaasSingleTaskGP,
 )
 from botorch.models.transforms.input import InputTransform
-from botorch.models.transforms.outcome import (
-    ChainedOutcomeTransform,
-    Log,
-    OutcomeTransform,
-    Standardize,
-)
+from botorch.models.transforms.outcome import OutcomeTransform, Standardize
 from gpytorch.mlls import ExactMarginalLogLikelihood
 
 from bofire.data_models.enum import OutputFilteringEnum
@@ -88,12 +83,6 @@ class EnsembleMapSaasSingleTaskGPSurrogate(TrainableBotorchSurrogate):
             outcome_transform = Standardize(
                 m=tY.shape[-1],
                 batch_shape=torch.Size([self.n_taus]),
-            )
-        if isinstance(outcome_transform, Log) or isinstance(
-            outcome_transform, ChainedOutcomeTransform
-        ):
-            raise RuntimeError(
-                "Log output transform is not supported with EnsembleMapSaasSingleTaskGP."
             )
         self.model = EnsembleMapSaasSingleTaskGP(
             train_X=tX,
--- a/tests/bofire/surrogates/test_map_saas.py
+++ b/tests/bofire/surrogates/test_map_saas.py
@@ -1,4 +1,3 @@
-import pytest
 from botorch.models.map_saas import (
     AdditiveMapSaasSingleTaskGP,
     EnsembleMapSaasSingleTaskGP,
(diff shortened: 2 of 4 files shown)


jduerholt:
Super nice, thanks, I will review this tmr, I was planning to do the same ;)

bofire/benchmarks/single.py
```diff
@@ -440,6 +440,31 @@ def get_optima(self) -> pd.DataFrame:
         )
 
 
+class PositiveHimmelblau(Himmelblau):
```
jduerholt, line 443: Hmm, the only difference is, that you are adding 1e-8?

jduerholt: Where do we need this?

jduerholt: Only for tests?

bofire/surrogates/map_saas.py
```diff
@@ -78,12 +83,18 @@ def _fit_botorch(
...
     ):
         # EnsembleMapSaasSingleTaskGP repeats the data to create a batch dimension
-        # The outcome_transform needs to have the correct batch_shape
+        # The Standardize outcome_transform needs to have the correct batch_shape
         if isinstance(outcome_transform, Standardize):
             outcome_transform = Standardize(
                 m=tY.shape[-1],
                 batch_shape=torch.Size([self.n_taus]),
             )
+        if isinstance(outcome_transform, Log) or isinstance(
```
jduerholt: Can you catch this directly in the data model via a validator?

jduerholt: And why is it not supported?

tests/bofire/surrogates/test_map_saas.py
```diff
@@ -50,3 +54,81 @@ def test_EnsembleMapSaasSingleTaskGPSurrogate():
...
+    dump = gp.dumps()
+    gp2 = surrogates.map(data_model=data_model)
+    gp2.loads(dump)
+    preds = gp.predict(experiments)
+    assert preds.shape == (10, 2)
+    preds2 = gp.predict(experiments)
+    assert_frame_equal(preds, preds2)
+
+
+def test_EnsembleMapSaasSingleTaskGPSurrogate_log_output_transform():
```
jduerholt: This should then also go to the specs with an add_invalid after we moved it to the data model and a validatr.

jduerholt (changes requested):
Looks good to me. Thanks. Only some minor changes

bofire/benchmarks/single.py
KislayaRavi: Yes, we need it only for test. This is added as a benchmark where the objective function is strictly positive.

bofire/surrogates/map_saas.py
KislayaRavi: I agree that I should add move the instance check to the data_models validation.
To answer your second question, Standardize is a linear transformation. So, the mean and variance can be calculated arithmetically even in multi-dimensional setups. 
With Log transformation, things are a not that straightforward because of non-linearity. So, ```untransform_posterior``` step of output_transform,  Botorch returns an object of type distribution from which one needs to draw samples and then get the statistical moments. 
When I run ```EnsembleMapSaasSingleTaskGPSurrogate``` with Log transform, I get the following error:
```AttributeError: 'TransformedPosterior' object has no attribute 'distribution'```
because ```EnsembleMapSaasSingleTaskGPSurrogate``` expects different type of object.

tests/bofire/surrogates/test_map_saas.py
KislayaRavi: Done.

PR #719, opened 2026-02-11
Normalize with engineered feats

TobyBoyne:
## Motivation

In #678, bounds were no longer provided to input transforms, and were instead learned from data. This was done to avoid explicitly providing the bounds of the engineered features.

However, this caused issues in #705, since I required an input to be strictly within the bounds [0, 1], but it was often scaled beyond that value if the training data didn't cover the full space. So instead, we revert back to the original behaviour for regular continuous features, and only learn the bounds for engineered features.

Yes

## Test Plan

I have adapted the existing test cases to verify the correct behaviour.

### DIFF
--- a/bofire/surrogates/utils.py
+++ b/bofire/surrogates/utils.py
@@ -1,4 +1,4 @@
-from typing import List, Union
+from typing import List, Mapping, Union
 
 import pandas as pd
 import torch
@@ -63,7 +63,7 @@ def get_scaler(
     categorical_encodings: InputTransformSpecs,
     scaler_type: ScalerEnum,
     X: pd.DataFrame | None = None,
-) -> InputStandardize | Normalize | dict[str, Normalize] | None:
+) -> Mapping[str, Normalize | InputStandardize] | None:
     """Returns the instanitated scaler object for a set of input features and
     categorical_encodings.
 
@@ -77,7 +77,7 @@ def get_scaler(
         X: Experimental values of input features
 
     Returns:
-        The instantiated botorch scaler object or None if no scaling is to be
+        The instantiated botorch scaler(s) object or None if no scaling is to be
             applied.
 
     """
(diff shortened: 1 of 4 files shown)


bofire/surrogates/utils.py
```diff
@@ -97,23 +97,48 @@ def get_scaler(
...
 
     if scaler_type == ScalerEnum.NORMALIZE:
-        return Normalize(
-            d=d,
-            # bounds=torch.tensor([lower, upper]).to(**tkwargs),
-            indices=ord_dims,
-            batch_shape=torch.Size(),
+        # We create a separate Normalize for non-engineered features,
+        # since bounds are known for these features.
+        lower, upper = inputs.get_bounds(
```
jduerholt, line 118: This would introduce a bug, you have to use the actual training data here, because it can be that the training data is wider as the bounds. Have a look here: https://github.com/experimental-design/bofire/pull/678/changes#diff-0e88d06b923cb05ccccfbb3e27430a73e500ba21811ed5520ade8602b9a52bb7

where I changed the original behavior, have a look here.

bofire/surrogates/utils.py
jduerholt, line 201: Do not return a chained input transform but a dictionary of scalers that we then use here.

bofire/surrogates/utils.py
```diff
@@ -61,7 +62,8 @@ def get_scaler(
     engineered_features: EngineeredFeatures,
     categorical_encodings: InputTransformSpecs,
     scaler_type: ScalerEnum,
-) -> Union[InputStandardize, Normalize, None]:
+    X: pd.DataFrame | None = None,
+) -> InputStandardize | Normalize | dict[str, Normalize] | None:
```
jduerholt: Can you always return a dictionary?

PR #724, opened 2026-02-13
Add BO termination conditions

wgst:
## Motivation

BO loops in BoFire currently run for a fixed number of iterations. In practice, especially for chemical experiments, where each iteration is very costly, we need a way to stop optimization early enough to reduce costs but still get a converged result.

This is an attempt to implement different termination conditions methods. The first method currently added in this PR is from [Makarova, A. et al. (2022)](https://proceedings.mlr.press/v188/makarova22a/makarova22a.pdf). The idea behind this method is to compute an upper bound on the simple regret of the current best recommendation using the GP posterior, and when this bound falls below a threshold ε_BO (derived from the observation noise variance), then we are confident that any further improvement will be smaller than the noise, and optimization can stop.

### Two other termination methods - issues with implementing

1. **EMSR Gap:** [Ishibashi, H. et al. (2023)](https://proceedings.mlr.press/v206/ishibashi23a.html). This method requires computing an Expected Minimum Simple Regret gap between consecutive iterations, which needs per-iteration model snapshots or knowledge of which experiments belong to which iteration. I have not found any tracking functionalities within BoFire, which makes it hard to implement this method at the moment.

2. **Probabilistic Regret Bound (PRB):** [Wilson, J. et al. (2024)](https://proceedings.neurips.cc/paper_files/paper/2024/file/b204de7078301292a8876a762eed3dcb-Paper-Conference.pdf). This is a more probabilistic method, in which a statistically valid stopping rule is constructed using Matheron-rule pathwise trajectory samples (Thompson-style posterior samples via pathwise conditioning). Their implementation is based on TensorFlow and BO packages around it, and some crucial functionalities are not available in BoTorch.

Both methods could be good candidates for future work. Adding the method by Ishibashi et al should be fairly easy once the tracking architecture is settled. Method by Wilson could be a bit more problematic, but let's see.

Yes

## Test Plan

Added tests:
- **data model** (`test_termination.py`)
- **evaluator** (`test_evaluator.py`)
- **mapper** (`test_mapper.py`)

Best,
Wojciech


bofire/data_models/termination/termination.py
```diff
@@ -0,0 +1,224 @@
...
+        self,
+        domain: Domain,
+        experiments: Optional[pd.DataFrame],
+        iteration: int,
+        **kwargs,
+    ) -> bool:
+        return iteration >= self.max_iterations
+
+
+class AlwaysContinue(TerminationCondition, EvaluatableTermination):
```
jduerholt: This one also already exisits in the StepwiseStrategy as a condition.

bertiqwerty (changes requested):
Please refactor into the `StepwiseStrategy`.

bofire/data_models/termination/termination.py
wgst: Good idea! I have now moved everything to StepwiseStrategy, so I hope it's in the right place now.

bertiqwerty:
Sorry, for the delay. I will try to look into this next week.

wgst:
Hi @jduerholt, and @bertiqwerty,

I implemented all four termination conditions that I intended to.

This includes:
- [Makarova et al. (2022)](https://proceedings.mlr.press/v188/makarova22a/makarova22a.pdf) method (UCBLCBRegretBoundCondition), which tracks an upper bound on the simple regret (the gap min UCB(evaluated) - min LCB(domain) from GP-UCB confidence bounds), and stops the BO loop whenever it falls below a threshold. The threshold can be a chosen fixed value (e.g., experimental error), the GP's estimated noise, or a cross-validation-based estimate.
- [Ishibashi et al. (2023)](https://proceedings.mlr.press/v206/ishibashi23a.html) method (ExpMinRegretGapCondition) that builds on Makarova et al. method by also evaluating the UCB-LCB regret bound, but it additionally introduces new terms, including the KL divergence between the predictive distribution of GP at the newest point before vs after conditioning on that new observation. The threshold can be adaptive (self-calibrating from the GP posterior variances and noise) or a median-based heuristic.
- [Xie. et al. (2023)](https://arxiv.org/html/2507.12453v5) method (LogEIPCCondition), a cost-aware stopping for when the cost of experiments plays a crucial role. It uses a cost-adjusted acquisition function, LogEIC(x) = LogEI(x) − α·log c(x), and stops when no candidate's expected improvement is worth its evaluation cost. Currently, the LogEIC acquisition function is implemented within the BoFire (temporarily). However, I already submitted a BoTorch PR, so that the acquisition functions stay in BoTorch, not BoFire: https://github.com/meta-pytorch/botorch/pull/3304.
- [Wilson, J. et al. (2024)](https://proceedings.neurips.cc/paper_files/paper/2024/file/b204de7078301292a8876a762eed3dcb-Paper-Conference.pdf), the Probabilistic Regret Bound (PRB) method (ProbabilisticRegretBoundCondition), which utilizes Matheron-rule pathwise trajectory samples (posterior samples via pathwise conditioning), minimises each sampled path, and stops once a sequential Clopper–Pearson test certifies that P(regret ≤ ε) ≥ 1 − δ (where ε is the simple-regret tolerance and δ is the total risk budget, split between model risk and Monte-Carlo estimation risk, and both ε and δ are set by the user).


As suggested by you previously, I've also moved all the conditions to StepwiseStrategy.

Best,
Wojciech

wgst:
Hi @jduerholt, and @bertiqwerty,

I have now reworked the code to match the new convergence criteria structure implemented in your recent PR: https://github.com/experimental-design/bofire/pull/786 (now in main).

Best,
Wojciech

bertiqwerty:
Thanks Wojciech. Looks good at first glance. I will have a closer look next week. Just two things that I realized:

1. For better separation, make abbreviations in CamelCase-identifiers lower case, at least if they are not at the end. For instance, `LogEipcCriterion` instead of `LogEIPCCriterion`. We also call it `SoboStrategy` instead of `SOBOStrategy`.
2. Create a docs page that shows how to use the convergence criteria. Feel free to pick one as example.

wgst:
That makes sense, Behrang. Thanks for the suggestions! I've added a new docs page on Convergence Criteria in the User Guides. Let me know if you think it should be somewhere else.

bofire/strategies/convergence_criteria/evaluator.py
```diff
@@ -0,0 +1,337 @@
...
+    experiments = strategy.experiments
+    if experiments is None:
+        return None
+    experiments = strategy.domain.outputs.preprocess_experiments_all_valid_outputs(
+        experiments
+    ).reset_index(drop=True)
+    if len(experiments) < min_experiments:
+        return None
+    if (
+        not getattr(strategy, "is_fitted", False)
```
bertiqwerty: Use `strategy.is_fitted` and `strategy.model`. Both should exist since convergence criteria can only be applied to predictive strategies anyway.
```
if not strategy.is_fitted or strategy.model is None:
    return None
```

bofire/strategies/convergence_criteria/evaluator.py
```diff
@@ -0,0 +1,337 @@
...
+        (which maximizes).  Returns ``None`` for any other objective
+        (``CloseToTargetObjective``, sigmoid, desirability, ...), for which the
+        regret-to-optimum bound logic does not apply, so the caller should skip
+        evaluation and return empty metrics.
+
+        The regret evaluators read the GP posterior in raw Y units and work in a
+        "lower is better" frame, so they negate this value
+        (``sign = -direction``) to obtain their minimisation-frame sign.
+        """
+        from bofire.data_models.objectives.api import (
```
bertiqwerty: Imports should be at the top of the file

bofire/strategies/convergence_criteria/exp_min_regret_gap.py
```diff
@@ -0,0 +1,622 @@
...
+
+    Returns the refit strategy, or ``None`` when the refit fails (callers
+    should skip evaluation). The main strategy is unaffected.
+    """
+    from bofire.strategies.mapper import map as map_strategy
+
+    try:
+        prefix_strategy = map_strategy(strategy._data_model)
+        prefix_strategy.tell(experiments.iloc[:m])
+    except Exception:
```
bertiqwerty: be more specific

bofire/strategies/convergence_criteria/exp_min_regret_gap.py
```diff
@@ -0,0 +1,622 @@
...
+        self._seq_values = state["seq_values"]
+
+
+def _refit_on_prefix(strategy, experiments: pd.DataFrame, m: int):
+    """Refit a copy of the strategy on the first ``m`` recorded experiments.
+
+    Returns the refit strategy, or ``None`` when the refit fails (callers
+    should skip evaluation). The main strategy is unaffected.
+    """
+    from bofire.strategies.mapper import map as map_strategy
```
bertiqwerty, line 469: imports at the top

bofire/strategies/convergence_criteria/evaluator.py
```diff
@@ -0,0 +1,337 @@
+"""Evaluator base classes that compute metrics for convergence criteria."""
+
+from abc import ABC, abstractmethod
+from typing import Any, Dict, Literal, Optional
```
bertiqwerty: Prefer `dict` and `list` over `Dict` and `List`. Also prefer `T | None` over `Optional`. Admittedly, we have many legacy `Optional`s in the code base. Same for `Union`s. Prefer `T1 | T2` over `Union[T1, T2]`.

bofire/strategies/convergence_criteria/exp_min_regret_gap.py
```diff
@@ -0,0 +1,622 @@
...
+            ``"sample"`` (default) or ``"optimize"``.
+
+    Reference:
+        Ishibashi et al. (2023): "A stopping criterion for Bayesian optimization
+            by the gap of expected minimum simple regrets" (AISTATS 2023).
+    """
+
+    def __init__(
+        self,
+        delta: float = 0.1,
```
bertiqwerty: don't duplicate the parameters of the data model. use the data model as parameter. Of course, this constructor can have more parameters than the data model. For instance `adaptive_median` already deviates.

bofire/data_models/strategies/convergence_criteria/exp_min_regret_gap.py
```diff
@@ -0,0 +1,78 @@
...
+        beta_scale: Scaling factor for the GP-UCB beta parameter.
+        n_samples_lcb: Random samples for the min-LCB estimate in kappa.
+        noise_var_override: If set, replaces the GP's learned noise variance
+            when computing the adaptive threshold.  Use a small value (e.g.
+            ``1e-6``) for exact (noise-free) objectives, where the GP can
+            otherwise over-estimate noise early and trigger a premature stop.
+    """
+
+    type: Literal["ExpMinRegretGapCriterion"] = "ExpMinRegretGapCriterion"
+    threshold_mode: Literal["adaptive", "median"] = "adaptive"
```
bertiqwerty: isn't `adaptive_median` missing?

bertiqwerty (approved):
Thanks, Wojciech. Looks good to me. Just a few minor things. Please, resolve them. Then you can merge. But I think you have to resolve some merge conflicts first.

bertiqwerty:
Ah, I didn't realize the workflows didn't run. Make sure the tests are ok before you merge.

bofire/strategies/convergence_criteria/exp_min_regret_gap.py
wgst: Here, the import at the top is circular, which is why I left it inside the function.

> File ".../bofire/strategies/convergence_criteria/exp_min_regret_gap.py", line 36, in <module>
>     from bofire.strategies.mapper import map as map_strategy
> ImportError: cannot import name 'map' from partially initialized module
> 'bofire.strategies.mapper' (most likely due to a circular import)
> (.../bofire/strategies/mapper.py)

bofire/strategies/convergence_criteria/evaluator.py
wgst: Done

bofire/strategies/convergence_criteria/evaluator.py
wgst: Done

bofire/strategies/convergence_criteria/exp_min_regret_gap.py
wgst: Done

bofire/strategies/convergence_criteria/evaluator.py
wgst: They are now updated

bofire/data_models/strategies/convergence_criteria/exp_min_regret_gap.py
wgst: Good catch, it slipped out somehow. It's now added

wgst:
Thanks, Behrang! I addressed all the comments now and updated the documentation formatting to match the new convention. Could you rerun the workflows?

wgst:
Everything looks green now, could you merge it (I don't seem to have permissions)?

PR #725, opened 2026-02-15
Improved `Features` types

TobyBoyne:
## Motivation

With the latest release of `ty==0.0.17`, the type checker now errors for attribute access on union where some elements lack the attribute (as mentioned in #719). This means that code like below now raises a typing error:

```python
inputs = Inputs(...)
for feature in inputs.get(CategoricalInput):
    print(feature.categories)
# >> ty check
#  Error: Attribute `categories` is not defined on `DiscreteInput`, `ContinuousInput` in union `DiscreteInput | CategoricalInput | ContinuousInput`
```

A short term fix is just to pin the version of ty. This PR presents two different possible approaches to fix this in the long term. Let me know what you think of them @jduerholt, and whether you would like me to go ahead with either/both of them :)

### 1. TypeGuards 

We can use [TypeGuard](https://docs.python.org/3/library/typing.html#typing.TypeGuard) to filter down the possible features contained within a `Features` object. For example, this PR currently includes one for checking if all the features in an `Inputs` object are continuous. This can be used as below (also see the change in `benchmarks.py` for an example).

```python
if Inputs.is_continuous(inputs):
    for feature in inputs.get():
        print(feature.lower_bound) 
# no ty check errors!
```

### 2. Overloads

We can also [overload](https://docs.python.org/3/library/typing.html#typing.overload) the `get` method on `Features` to specify the types of features in the containers. This approach should be a bit more seamless, since you won't need to call an extra function compared to approach (1). I've currently added an example implementing this for `Outputs`, with an example in `naming_conventions.py`.

```python
for feature in inputs.get(CategoricalInput):
    # thanks to the overload, we can now infer the type of `inputs.get(CategoricalInput)` is `Inputs[CategoricalInput]`
    print(feature.categories)
# again, no ty check errors!
```

--- 

Yes

Not yet

## Test Plan

The number ty errors that are currently raised on the CI should go down.

### DIFF
--- a/bofire/data_models/domain/features.py
+++ b/bofire/data_models/domain/features.py
@@ -66,7 +66,7 @@ EngineeredFeatureT = TypeVar(
     "EngineeredFeatureT", bound=AnyEngineeredFeature, default=AnyEngineeredFeature
 )
 OutputT = TypeVar("OutputT", bound=AnyOutput | Output, default=AnyOutput | Output)
-OutputGetT = TypeVar("OutputGetT", bound=AnyOutput, default=AnyOutput)
+OutputGetT = TypeVar("OutputGetT", bound=AnyOutput | Output, default=AnyOutput | Output)
 
 
 class _BaseFeatures(BaseModel, Generic[FeatureT]):
@@ -919,7 +919,7 @@ class Inputs(_BaseFeatures[InputT]):
     @overload
     def get(
         self,
-        includes: Type[InputGetT] | Sequence[Type[InputGetT]],
+        includes: Type[InputGetT] | Sequence[Type[InputGetT]] = Input,
         excludes: None = None,
         exact: bool = False,
     ) -> Inputs[InputGetT]: ...
--- a/bofire/data_models/domain/constraints.py
+++ b/bofire/data_models/domain/constraints.py
@@ -1,7 +1,7 @@
 import collections.abc
 from collections.abc import Iterator, Sequence
 from itertools import chain
-from typing import Generic, List, Literal, Type, Union, overload
+from typing import Generic, Literal, Type, Union, overload
 
 import pandas as pd
 from pydantic import Field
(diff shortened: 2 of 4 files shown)


bofire/data_models/domain/features.py
```diff
@@ -1215,3 +1224,27 @@ def preprocess_experiments_any_valid_output(
             ),
         )
         return clean_exp
+
+    @overload
```
jduerholt: Why do we need the overload two times?

jduerholt (commented):
This looks super interesting, I personally like the `overload` approach, despite not really understanding it :D 

@bertiqwerty what do you think? I personally think that implementing the overload would be quite nice.

bofire/data_models/domain/features.py
TobyBoyne: Unfortunately that's part of the specification for overloads ([link](https://typing.python.org/en/latest/spec/overload.html#invalid-overload-definitions)). I think the idea is that the actual implementation of `get` should have no type hints, and all of the type hints go into the different overloads. 

So the second overload is a fallback if the argument types don't match the first overload, which will happen if `excludes` is provided, since you can't express `Intersection[GetIncludesT, ~GetExcludesT]` in Python's type system (yet) so we need the fallback to be more generic!

jduerholt:
I like it and you are right we are only using it for the feature containers. It will definitely add some boiler plate code but will make typehints etc. much nicer. @bertiqwerty: should Toby go ahead with this? What do you think?

bofire/utils/torch_tools.py
```diff
@@ -71,15 +72,16 @@ def get_linear_constraints(
 
     """
     constraints = []
+    inputs = domain.inputs.get([ContinuousInput, DiscreteInput])
     for c in domain.constraints.get(constraint):
```
TobyBoyne, line 76: `ty` infers the type of `c` to be `Constraint`, whereas `pyright` correctly infers it to by `LinearEqualityConstraint | LinearInequalityConstraint`. Adding an extra `.get([LinearEqualityConstraint, LinearInequalityConstraint])` here reveals `@Todo` with `ty`.

PR #726, opened 2026-02-16
Refactor Scalers

jduerholt:
## Motivation

This PR refactors the scaler setup in the data models, and migrates it from being an enum to be a class, in this class one can specify which features should be scaled and which not. In the future we can further extend this to all kind of scalers (functions that operates) on features and engineered features, broadcasted to their final shape. One could also introduce here chained scalers. An example would be first a log10 transformation and then the normalization. Also warping could be implemented here.

Yes.

Yes.

## Test Plan

Unit tests.

cc: @chrihaas @Jimbo994

### DIFF
--- a/tests/bofire/data_models/specs/surrogates.py
+++ b/tests/bofire/data_models/specs/surrogates.py
@@ -167,7 +167,7 @@ specs.add_invalid(
         "scaler": Normalize(features=["d", "e"]),
     },
     error=ValueError,
-    message="The following features are missing in inputs: ['d', 'e']",
+    message="The following features are missing in inputs",
 )
 
 


bofire/data_models/surrogates/trainable_botorch.py
```diff
@@ -1,30 +1,22 @@
...
-            ScalerEnum: Scaler used to transform the output
-
-        """
-        if output_scaler == ScalerEnum.NORMALIZE:
-            raise ValueError("Normalize is not supported as an output transform.")
-
-        return output_scaler
+    @model_validator(mode="after")
+    def validate_scaler_features(self):
+        if self.scaler and self.scaler.features is not None:
```
TobyBoyne: When is `self.scaler.features` `None`? I thought it had to be of type `NonRestrictedFeatureKeys`, which is a list?

bofire/surrogates/engineered_features.py
```diff
@@ -110,12 +111,32 @@ def map_molecular_weighted_sum_feature(
...
+    inputs: Inputs,
+    transform_specs: InputTransformSpecs,
+    feature: CloneFeature,
+) -> AppendFeatures:
+    features2idx, _ = inputs._get_transform_info(transform_specs)
+    indices = [features2idx[key][0] for key in feature.features]
+
+    def clone_features(X: torch.Tensor, indices: torch.Tensor) -> torch.Tensor:
+        result = X[..., indices].unsqueeze(-2)
+        return result.expand(*result.shape[:-2], 1, -1)
```
TobyBoyne: Does this line do anything? From L123, `result.shape == (*result.shape[:-2], 1, result.shape[-1])`. So then L124 reshapes this to the same shape that it already is? I might be missing something here.

bofire/surrogates/utils.py
```diff
@@ -95,24 +95,49 @@ def get_scaler(
...
-        offset=offset, feature_keys=engineered_features.get_keys()
-    )
     ord_dims = cont_feat_dims + engineered_feat_dims
 
     if len(ord_dims) == 0:
-        return None
+        return {}
 
-    if scaler_type == ScalerEnum.NORMALIZE:
+    if isinstance(scaler_type, NormalizeDataModel):
```
TobyBoyne: (Very very minor) In the tests, you import this as `NormalizeScaler`, here you import it as `NormalizeDataModel`. I prefer the first one, so maybe change this for consistency?

bofire/surrogates/engineered_features.py
jduerholt: Good catch, some legacy story on this. I removed it  ;)

bofire/surrogates/utils.py
jduerholt: Fixed it.

bofire/data_models/surrogates/trainable_botorch.py
jduerholt: Good catch, I also added a test for this.

PR #729, opened 2026-02-17
Fix jacobian and hessian evaluation for list expressions

MrzvskK:
## Motivation

Hi Johannes, when working on nonlinear constraints I have noticed that two tests were failing, my guess it is due to `pandas.eval()` and `numexpr` compatibility issues with list expressions in Python 3.13.
The tests in question: 
- `test_nonlinear_constraints_jacobian_expression`
- `test_nonlinear_constraints_hessian_expression`

Cause analysis: 
The `jacobian()` and `hessian()` methods were attempting to evaluate string expressions containing list literals such as `"[2*x1, 2*x2, -1]"` - directly with `pandas.eval()`. 
When doing so you get error message: 
`ValueError: setting an array element with a sequence` due to inhomogeneous array shapes.

At first I created an issue, but then I made a few modifications to the nonlinear.py and now tests pass - I hope these fixes are good enough. 
Both for Jacobian and Hessian we first detect list/nested list expressions, then these components are evaluated separately and we handle all the scalars by pushing them to all rows. Not sure if using dataFrame was a good call. 

## Changes
- **`jacobian()`**: Added parsing for list expressions like `"[2*x1, 2*x2, -1]"`
- **`hessian()`**: Added parsing for nested list expressions like `"[[6*x1, 0, 0], [0, 2, 0], [0, 0, 0]]"`

All other tests in the constraints also passed, and we maintain compatibility with Callable expressions and backward one for non-list expressions

yes, but I am a new contributor

no

## Test Plan

I would run: pytest tests/bofire/data_models/constraints/test_nonlinear.py -v
Or if you want to test all tests run: pytest tests/bofire/data_models/constraints/ -v

### DIFF
--- a/bofire/benchmarks/multi.py
+++ b/bofire/benchmarks/multi.py
@@ -165,20 +165,20 @@ class BNH(Benchmark):
 
     def _f(self, candidates: pd.DataFrame) -> pd.DataFrame:
         experiments = candidates.eval(
-            "f1=4*x1**2 + 4*x2**2", inplace=False, backend="python"
+            "f1=4*x1**2 + 4*x2**2", inplace=False, engine="python"
         )
         experiments = experiments.eval(
-            "f2=(x1-5)**2 + (x2-5)**2", inplace=False, backend="python"
+            "f2=(x1-5)**2 + (x2-5)**2", inplace=False, engine="python"
         )
         experiments["valid_f1"] = 1
         experiments["valid_f2"] = 1
         if not self.constraints:
             return experiments[["f1", "f2", "valid_f1", "valid_f2"]].copy()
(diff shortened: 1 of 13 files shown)


TobyBoyne:
I agree that we should decide on a single backend. Since the pandas code isn't performance critical (all of the heavy lifting is done by `torch` or some optimization package like `pymoo|SCIP|...`), then we can just use the `"python"` backend since it seems to have the most lenient syntax?

Is there a way to globally set the `pandas` backend? Having to set `backend="python"` at every `eval` call seems a bit annoying but we should still do it.

jduerholt (approved):
Thank you very much!

Issue #655, opened 2025-11-07
[FEATURE REQUEST]: PFN Surrogates


jduerholt:
### Motivation

BoTorch is now supporting PFN models including input transforms. We should make them also available in BoFire. 

### Describe the solution you'd like to see implemented in BoFire.

This is the botorch implemention: https://github.com/meta-pytorch/botorch/blob/main/botorch_community/models/prior_fitted_network.py

Have a look at https://github.com/experimental-design/bofire/issues/650, there are some links on how surrogates are implemented.

### Pull Request

None

PR #731, opened 2026-02-18
Adding PFN Surrogate

KislayaRavi:
## Motivation
Adding PFN surrogates to BoFire. 
Closing Issue #655 
I used the ```botorch_community``` version of PFN as mentioned in the aforementioned issue.

(Write your motivation here.)

Yes

Yes

## Test Plan
-  Added valid and invalid specs check
-  Added ```test_pfn.py``` which test the PFN Surrogates.

### Some important things to note:
- The PFN surrogate in ```botorch_community``` does not apply ```outcome_transform``` automatically. So, I needed to add it explicitly in the ```_predict``` function of ```PFNSurrogates```. This also requires overriding ```_load``` and ```_dump``` functions for ```PFNsurrogate```.
- The performance depends considerably on the trained models. As of now, one can access the models in the [PFN4BO](https://github.com/automl/PFNs4BO/tree/main/pfns4bo/final_models) repo. Users can also access any models which they trained on their own via url requests.
- As of now ```TabPFN``` models cannot be used. It requires more work to add them (also some license issues). I have one eye on ```TabICL``` too. Its implementation is sometime further down the line (hopefully).
- Preferably run PFN on GPUs. Speed is considerably slow on CPUs.


LukasHebing:
> Thanks @KislayaRavi,
> 
> I think we should have a discussion (including @bertiqwerty and @LukasHebing ) on how to handle this. I am bit hesitant to merge it in due to the kind of dependency hell with the pfn4bo package (https://github.com/experimental-design/bofire/pull/731/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R55).
> 
> Maybe we should also add a bofire community directory (in analogy to botorch comunity) to handle this kind of stuff. From an architectural perspecitive, we would need to have an easy way of registering new data models and its functional equivalents, then people could easily put their custom models in without the need of changing the codebase.
> 
> What do you think?
> 
> Best,
> 
> Johannes

I like the idea of simplifying additional content (surrogates, objectives, etc.) with data-models without having to add them in 5 different bofire files. The architecture with data-model objects and "real" objects makes this difficult. 

A nice solution would be an abstract data-model for e.g. surrogates, which has also all abstract methods (including torch-callables, `_fit`,  etc.). The bahavior of the `map` functions could be pre-defined. But this requires some efforts.

If we have that in bofire, custom other content can be added way more easy.

bertiqwerty:
Hi all, thanks @KislayaRavi. 

I agree with @jduerholt that we should not merge this PR into BoFire due to the depedencies. Especially that only Python older 3.12 is allowed is a clear deal breaker.

Already some time ago I thought it could be helpful to have BoFire "extendible from the outside". Currently even if one would not want to use any mappers Pydantic would scream at you if you pass a datamodel of a surrogate that is not part of the corresponding union (e.g., `AnyBotorchSurrogate`) to a datamodel of a strategy , I guess. @jduerholt, what was the reason of creating these union types next to the inheritance hierarchy again? If this would be fixed, one maybe could create a BoFire community package maybe even in a separate repository under `experimental-design`. However, also a bofire-community-package should not be limited to Python versions `< 3.12`.

@LukasHebing, your proposal I did not understand. Maybe something to discuss in person or in a call?

jduerholt:
I agree with you all, from my perspective we need a `register` function for the core components of BoFire like 

```
register_surrogate(data_model, surrogate)
```

This method should then be exposed on the high level api, and can be used by people to add new surrogates, strategies, priors etc. without the need to modify the code. What do you think? Under the hood the method would register the data model and its functional equivalent in the example of a surrogate at this place:

https://github.com/experimental-design/bofire/blob/55fff4acfeb953a783196384c89c80a5fe59f069/bofire/surrogates/mapper.py#L100

What do you think?

I agree, a call for this could be helpful, who wants to join?

@KislayaRavi @LukasHebing @bertiqwerty @TobyBoyne

jduerholt:
Hi @KislayaRavi,

are you at the hackathon next week? Then we can discuss what we do here, also TabPFN3.0 is out. Maybe you can also share some of your experiences. We also have now the register functions, so it should be possible to just hook it in. Have you tried?

Best,

Johannes

KislayaRavi:
Yes, I will be coming to the Hackathon. Also, I have tried TabPFN and TabICL. Long story short, foundation models performed better compared to a well chosen GP. But they struggle in High dimensional spaces(like everything else 🙃 ). There is an algorithm that claims to solve this issue. Refer the paper
https://arxiv.org/abs/2505.20685
Let's discuss these things in person.

PR #722, opened 2026-02-12
Enable python 3.14

jduerholt:
## Motivation

This PR adds support for python 3.14 and adds it to the test pipelines.

In addition, we are introducing `CHANGELOG.md` to track changens in the future.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/experimental-design/bofire/blob/main/CONTRIBUTING.md#pull-requests)?

Yes.

## Test Plan

Unit tests.

PR #732, opened 2026-02-18
weightedmean implementation

Jimbo994:
## Motivation

Implemented WeightedMeanFeature to support descriptor-based weighted averaging of selected input features, alongside existing engineered feature transforms.

Yes

No, not yet

## Test Plan

Written tests for:

- data model validation of WeightedMeanFeature
- surrogate mapping correctness for WeightedMeanFeature (including explicit expected weighted-mean values)
- weighted-sum surrogate test strengthened with explicit expected-value assertion

### DIFF
--- a/bofire/surrogates/engineered_features.py
+++ b/bofire/surrogates/engineered_features.py
@@ -10,9 +10,11 @@ from bofire.data_models.features.api import (
     CloneFeature,
     EngineeredFeature,
     MeanFeature,
+    MolecularWeightedMeanFeature,
     MolecularWeightedSumFeature,
     ProductFeature,
     SumFeature,
+    WeightedMeanFeature,
     WeightedSumFeature,
 )
 from bofire.data_models.types import InputTransformSpecs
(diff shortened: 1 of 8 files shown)


jduerholt (changes requested):
Hi @Jimbo994,

nice addition! But can we not just add a booleanargument to the already exisiting `WeightedSumFeature` and its molecular equivalent that either enforces normalization by the sum of weights or not? Than we do not need to add a new feature type to BoFire ...

And please also merge the current main in. There have been some recent updates here, I think. 

Best,

Johannes

bofire/surrogates/engineered_features.py
```diff
@@ -18,15 +18,22 @@
...
-    ).unsqueeze(-2)
+    weights = X[..., indices]
+    weighted_sum = torch.matmul(weights, descriptors)
+    if normalize:
+        weight_sum = torch.clamp(
+            torch.sum(weights, dim=-1, keepdim=True),
+            min=torch.finfo(weights.dtype).eps,
+        )
+        weighted_sum = weighted_sum / weight_sum
+    result = weighted_sum.unsqueeze(-2)
```
jduerholt:
```suggestion
    return weighted_sum.unsqueeze(-2)
```

You can just return here, the expand should not be necessary. I also removed it on other ocassions.

TobyBoyne:
@jduerholt I messed around with updating entmoot for #722, not entirely sure what I've done wrong. Let me investigate.

Jimbo994:
What do you think of this @jduerholt, I added the data models `WeightedMeanFeature `and `MolecularWeightedMeanFeature` as we discussed, and is this close to what you mentioned about using the partials?

PR #736, opened 2026-02-23
Add register() functions for strategies and surrogates

jduerholt:
## Motivation

Provide public API to register custom strategy and surrogate mappings instead of requiring users to mutate internal dictionaries directly. For BotorchSurrogate subclasses, registration automatically rebuilds the AnyBotorchSurrogate union and BotorchSurrogates Pydantic model so custom surrogates work in botorch-based strategies out of the box. It addresses the discussion in this PR: https://github.com/experimental-design/bofire/pull/731

@TobyBoyne @bertiqwerty @LukasHebing @KislayaRavi what do you think?

This is a draft which should serve for discsussion, code build with Claude Code.

Yes.

Not yet.

## Test Plan

Unit tests.

### DIFF
--- a/bofire/data_models/features/_register.py
+++ b/bofire/data_models/features/_register.py
@@ -1,7 +1,5 @@
 """Registration utilities for custom engineered feature types."""
 
-import typing
-from collections.abc import Sequence
 from typing import Union
 
 
@@ -24,15 +22,8 @@ def register_engineered_feature(data_model_cls: type) -> None:
         tuple(features_api._ENGINEERED_FEATURE_TYPES)
     ]
 
-    # Lazy import to avoid circular dependencies
+    from bofire.data_models._register_utils import append_to_union_field
     from bofire.data_models.domain.features import EngineeredFeatures
 
-    # Patch the Sequence[Union[...]] annotation on EngineeredFeatures.features
-    old = EngineeredFeatures.model_fields["features"].annotation
-    inner_args = typing.get_args(typing.get_args(old)[0])
-    if data_model_cls not in inner_args:
-        new_inner = Union[tuple(list(inner_args) + [data_model_cls])]
-        new_ann = Sequence[new_inner]
-        EngineeredFeatures.__annotations__["features"] = new_ann
-        EngineeredFeatures.model_fields["features"].annotation = new_ann
+    append_to_union_field(EngineeredFeatures, "features", data_model_cls)
     EngineeredFeatures.model_rebuild(force=True)


bofire/data_models/kernels/api.py
KislayaRavi, line 1: Do we also need `_rebuild_dependent_strategy` similar to `rebuild_dependent_models` if we want full dynamic Pydantic acceptance of new strategy data-model types?

bofire/data_models/kernels/api.py
```diff
@@ -75,3 +77,82 @@
...
+        inner = typing.get_args(old)[0]
+        inner_args = typing.get_args(inner)
+        if new_kernel_cls not in inner_args:
+            new_inner = Union[tuple(list(inner_args) + [new_kernel_cls])]
+            new_ann = Sequence[new_inner]
+            model_cls.__annotations__[field_name] = new_ann
+            model_cls.model_fields[field_name].annotation = new_ann
+
+    # ScaleKernel.base_kernel is Union[...] (not Sequence)
+    old = ScaleKernel.model_fields["base_kernel"].annotation
```
jduerholt: Why not doing this in the loop from above, when we loop also over the MultiplicativeKernel etc.?

bofire/data_models/kernels/api.py
```diff
@@ -75,3 +77,82 @@
...
+
+AnyKernel = Union[tuple(_KERNEL_TYPES)]
+
+
+def _rebuild_dependent_models(new_kernel_cls: Type[Kernel]) -> None:
+    """Rebuild all Pydantic models whose fields reference AnyKernel."""
+    from bofire.data_models.surrogates.botorch_surrogates import BotorchSurrogates
+    from bofire.data_models.surrogates.multi_task_gp import MultiTaskGPSurrogate
+    from bofire.data_models.surrogates.single_task_gp import SingleTaskGPSurrogate
+    from bofire.data_models.surrogates.tanimoto_gp import TanimotoGPSurrogate
```
jduerholt: For the mixedSingletaskGp, there is also an AnyContinuousKernel for example? How to handle this, and how to rebuild this? 

do we need specific parameters to specify if we place it in AnyContinuousKernel, or AnyCategoricalKernel? Or should we release it and make it to AnyKernel? And not have the specific AnyContinuous and AnyCategoricalKernels?

bofire/data_models/priors/api.py
```diff
@@ -25,18 +25,199 @@
...
+    LogTransformedInterval,
+    Positive,
+    GreaterThan,
+    LessThan,
 ]
 
+AnyPriorConstraint = Union[tuple(_PRIOR_CONSTRAINT_TYPES)]
+
+
+def _patch_field(model_cls: type, field_name: str, new_union: type) -> None:
```
jduerholt: Could we reuse this function also add other places?

bofire/data_models/priors/api.py
```diff
@@ -25,18 +25,199 @@
...
+    )
+    from bofire.data_models.surrogates.shape import PiecewiseLinearGPSurrogate
+    from bofire.data_models.surrogates.single_task_gp import (
+        SingleTaskGPHyperconfig,
+        SingleTaskGPSurrogate,
+    )
+    from bofire.data_models.surrogates.tanimoto_gp import TanimotoGPSurrogate
+
+    # Patch AnyPrior fields
+    for model_cls, field_name in [
```
jduerholt: How to make sure that we update this method in the future automatically? Is there a way, or do we have to do it always manually?

bofire/kernels/mapper.py
```diff
@@ -21,6 +21,52 @@
...
+
+    Returns:
+        The mapping function (unchanged) when used as a decorator, None otherwise.
+    """
+
+    def _register(fn: Callable) -> Callable:
+        KERNEL_MAP[data_model_cls] = fn
+
+        # Also register with the data model union so Pydantic accepts the type
+        from bofire.data_models.kernels.api import register_kernel
```
jduerholt: Do we need the import here, can we not just say `data_models.register_kernel()`

bofire/priors/mapper.py
```diff
@@ -12,14 +12,69 @@
...
+    def _register(fn: Callable) -> Callable:
+        PRIOR_MAP[data_model_cls] = fn
+
+        # Also register with the data model unions so Pydantic accepts the type
+        from bofire.data_models.priors.constraint import PriorConstraint
+        from bofire.data_models.priors.interval import Interval
+        from bofire.data_models.priors.prior import Prior
+
+        if issubclass(data_model_cls, Prior):
+            from bofire.data_models.priors.api import register_prior
```
jduerholt: Do we actually need this import? Or can we just say data_models.register_prior?

bofire/strategies/mapper.py
```diff
@@ -1,9 +1,53 @@
...
 import bofire.data_models.strategies.api as data_models
 from bofire.strategies.mapper_actual import STRATEGY_MAP as ACTUAL_MAP
 from bofire.strategies.mapper_meta import STRATEGY_MAP as META_MAP
 from bofire.strategies.strategy import Strategy
 
 
+def register(
+    data_model_cls: Type[data_models.Strategy],
+    strategy_cls: Optional[Type[Strategy]] = None,
+    meta: bool = False,
```
jduerholt: We do not want to allow for registering new meta strategies, there is only one meta strategy, namely the stepwise strategy, and we do not want the user to add new ones.

jduerholt: But we would need to rebuild the data model of the stepwise strategy, or? We are currently not doing it.

jduerholt: There should be something in data_models_strategy for this purpose ...

bofire/data_models/kernels/api.py
jduerholt: Yes, correct. Should be fixed now. My Agent was overlooking this.

bofire/data_models/kernels/api.py
jduerholt: Fixed!

bertiqwerty:
> Ok, I did more stuff here, this is now almost final @bertiqwerty can you have a look? what do you think? I will also add a tutorial etc. But would be nice to get your assesment! And of course also the assesment from the others!

Sure. I will look into it.

bofire/data_models/features/api.py
```diff
@@ -1,4 +1,6 @@
-from typing import Union
+import typing
+from collections.abc import Sequence
+from typing import List, Type, Union
```
bertiqwerty: use `list` instead of `List`. Don't import `List`.

bofire/data_models/features/api.py
```diff
@@ -82,3 +84,35 @@
     InterpolateFeature,
     CloneFeature,
 ]
+
+AnyEngineeredFeature = Union[tuple(_ENGINEERED_FEATURE_TYPES)]
+
+
+def register_engineered_feature(data_model_cls: Type[EngineeredFeature]) -> None:
```
bertiqwerty: Maybe don't clutter the `api.py` with implementations? Create a separate file and import the function here?

bofire/data_models/features/api.py
```diff
@@ -82,3 +84,35 @@
...
+    Args:
+        data_model_cls: A concrete subclass of ``EngineeredFeature``.
+    """
+    global AnyEngineeredFeature
+    if data_model_cls in _ENGINEERED_FEATURE_TYPES:
+        return
+    _ENGINEERED_FEATURE_TYPES.append(data_model_cls)
+    AnyEngineeredFeature = Union[tuple(_ENGINEERED_FEATURE_TYPES)]
+
+    # Lazy import to avoid circular dependencies
```
bertiqwerty: see above. if you had separate file for the implementation, would the import still be circular?

bofire/data_models/features/api.py
```diff
@@ -82,3 +84,35 @@
...
+    from bofire.data_models.domain.features import EngineeredFeatures
+
+    # Patch the Sequence[Union[...]] annotation on EngineeredFeatures.features
+    old = EngineeredFeatures.model_fields["features"].annotation
+    inner_args = typing.get_args(typing.get_args(old)[0])
+    if data_model_cls not in inner_args:
```
bertiqwerty: this peace of code looks fragile. but currently i don't have a better idea. hmm... at least if this breaks at some point only the registration functinality is affected, not the existing modules.

bofire/data_models/priors/api.py
```diff
@@ -1,5 +1,5 @@
 from functools import partial
-from typing import Union
+from typing import List, Type, Union
```
bertiqwerty: see above

bofire/data_models/priors/api.py
```diff
@@ -25,18 +25,187 @@
...
+    LogTransformedInterval,
+    Positive,
+    GreaterThan,
+    LessThan,
 ]
 
+AnyPriorConstraint = Union[tuple(_PRIOR_CONSTRAINT_TYPES)]
+
+
+def _rebuild_dependent_models() -> None:
```
bertiqwerty: I prefer a separate file to keep the `api.py` as clean as possible.

bofire/data_models/strategies/actual_strategy_type.py
```diff
@@ -1,4 +1,4 @@
-from typing import Union
+from typing import List, Type, Union
```
bertiqwerty: see above

bofire/data_models/strategies/actual_strategy_type.py
```diff
@@ -42,3 +43,30 @@
     ShortestPathStrategy,
     FractionalFactorialStrategy,
 ]
+
+ActualStrategy = Union[tuple(_ACTUAL_STRATEGY_TYPES)]
+
+
+def register_strategy(data_model_cls: Type[Strategy]) -> None:
```
bertiqwerty: again don't clutter the `api.py`

bofire/kernels/mapper.py
```diff
@@ -21,6 +21,50 @@
...
+def register(
+    data_model_cls: Type[data_models.Kernel],
+    map_fn: Optional[Callable] = None,
+):
+    """Register a custom kernel mapping from data model to factory function.
+
+    Can be used as a decorator or as a direct function call::
+
+        # Decorator form
+        @register(MyKernelDataModel)
```
bertiqwerty, line 33: Doesn't adding the decorator to existing functions trigger unnecessary rebuild-model calls on import time? Our import time is already pretty long. Further, I find decorators less transparent. If someone wants to register, they could simply use the direct call form. For the existing stuff, I liked the old and stupid dict. And it is somehow incosistent to the data models where we kept the lists of types and extended them. I would prefer to keep the dicts and extend those.

bofire/data_models/kernels/api.py
bertiqwerty: > Yes, correct. Should be fixed now. My Agent was overlooking this. ([last week](https://github.com/experimental-design/bofire/pull/736#discussion_r2847562717))

Haha. Nice try blaming the agent. :D

bofire/kernels/mapper.py
jduerholt: @bertiqwerty: But the decorator in general would be fine? So, for registering stuff outside of the library? Or do you want to remove the decorator support in general?

bertiqwerty: You mean as option for people to use to add with new code outside of BoFire? There is no way to remove it, is there? The `register` function can always be used as decorator or with a simple function call, right?

jduerholt: True, so you would like to remove the decorator use from the inner workings of BoFire, correct?

bertiqwerty: Yes. And I want the dict back.

jduerholt: Ok, I will add the dict back in the internal mappings.

PR #738, opened 2026-02-25
allow only one descriptor

Jimbo994:
## Motivation

Was using the `WeightedMeanFeature` on `ContinuousDescriptorInput`s with only one descriptor, and was running into Pydantic Validation errors because we set `Descriptors` to have `min_length=2` is there a specific reason for this @jduerholt or can we simply alter this as I suggested?

Yes

No

## Test Plan

Run current tests. Ran locally without issues.

### DIFF
--- a/bofire/data_models/types.py
+++ b/bofire/data_models/types.py
@@ -100,7 +100,7 @@ CategoryVals = Annotated[
 
 Descriptors = Annotated[
     List[str],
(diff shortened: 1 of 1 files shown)


jduerholt (approved):
Makes sense, orginally we enforced it to at least two, but now with the engineered features it makes also sense to have just one.

PR #739, opened 2026-03-03
Feature/pre compute tanimoto2

LukasHebing:
-- Simpler implementation of #701 --

Tanimoto-similarities are computed upfront, now in new surrogate `TanimotoGPSurrogate`

This avoids re-computation during model training and evaluation and excessive memory usage

computation is done in kernel, cov-tensors are passed to the strategy via `re_init_kwargs`

Performance tests:
On a private benchmark with >1k molecules, this computation only needed ~10% of the computation time.

The old computation also lead to memory-errors, when many molecules were added as training data points leading to crashes.

## Test Plan

New tests is added, for comparing old and new calculation

### DIFF
--- a/bofire/surrogates/tanimoto_gp_surrogate.py
+++ b/bofire/surrogates/tanimoto_gp_surrogate.py
@@ -57,7 +57,7 @@ class TanimotoGPSurrogate(SingleTaskGPSurrogate):
                 )  # temporary kernel to compute the tanimoto similarity matrix
                 fingerprints = input_transform.encoders[0].encoding  # type: ignore[index]
                 cov_module = covar_module_pre_calc
-                while not hasattr(cov_module, "pre_computed_tanimoto"):
+                while not hasattr(cov_module, "tanimoto_similarity_matrix"):
                     cov_module = cov_module.base_kernel
                 cov = cov_module.forward(fingerprints, fingerprints)
                 self.tanimoto_similarity_matrix = cov.detach()


bofire/data_models/molfeatures/molfeatures.py
```diff
@@ -44,6 +44,10 @@ def _get_descriptor_values(self, values: pd.Series) -> pd.DataFrame:
         pass
 
     def remove_correlated_descriptors(self, molecules: List[str]):
+        if not self.filter_descriptors:
```
jduerholt, line 47: Why add the flag here? Then it is not so easy anymore to use the function as self-contained function?

bofire/kernels/fingerprint_kernels/tanimoto_kernel.py
```diff
@@ -61,11 +63,44 @@ class TanimotoKernel(BitKernel):
     is_stationary = False
     has_lengthscale = False
 
-    def __init__(self, **kwargs):
+    def __init__(
+        self,
+        pre_computed_tanimoto: bool = True,
+        tanimoto_similarity_matrix: Optional[torch.Tensor] = None,
```
jduerholt, line 68: Could one not build it just via the tanimoto_similarity_matrix argument, if it is not None, we use the precomputed, else we compute it?

bofire/surrogates/mapper.py
```diff
@@ -71,29 +72,12 @@ def map_MixedSingleTaskGPSurrogate(
...
-    )
-
-
 DATA_MODEL_MAP: Dict[
     Type[data_models.MixedSingleTaskGPSurrogate],
     Callable[[data_models.MixedSingleTaskGPSurrogate], data_models.AnySurrogate],
 ] = {
     data_models.MixedSingleTaskGPSurrogate: map_MixedSingleTaskGPSurrogate,
-    data_models.TanimotoGPSurrogate: map_TanimotoGPSurrogate,
+    # data_models.TanimotoGPSurrogate: map_TanimotoGPSurrogate,
```
jduerholt: Remove this line.

bofire/surrogates/tanimoto_gp_surrogate.py
```diff
@@ -0,0 +1,93 @@
...
+
+class TanimotoGPSurrogate(SingleTaskGPSurrogate):
+    def __init__(
+        self,
+        data_model: DataModel,
+        tanimoto_similarity_matrix: Optional[torch.Tensor] = None,
+        **kwargs,
+    ):
+        self.tanimoto_calculation_mode = data_model.tanimoto_calculation_mode
+        self.tanimoto_similarity_matrix = tanimoto_similarity_matrix
```
jduerholt, line 24: also here, do we not both? Why not just use the similarity_matrix as flag?

bofire/surrogates/tanimoto_gp_surrogate.py
```diff
@@ -0,0 +1,93 @@
...
+        input_transform: Optional[InputTransform] = None,
+        outcome_transform: Optional[OutcomeTransform] = None,
+        **kwargs,
+    ):
+        if input_transform is not None:
+            n_dim = input_transform(tX).shape[-1]
+        else:
+            n_dim = tX.shape[-1]
+
+        if self.tanimoto_calculation_mode == "pre_computed":
```
jduerholt, line 50: Ah, forget it, here you need both.

uv.lock
jduerholt, line 1: Can you remove this file please?

jduerholt (changes requested):
Hi @LukasHebing,

thanks for the PR, and please apologize that it took me so long for the review. I only have some minor comments. 

Main comment is that I was a bit puzzled if we really need so often both the boolean and the similarity matrix as inputs, maybe we can also only use one at some points, but I am not sure.

In general, really minor comments, should be in now really fast.

Best,

Johannes

bofire/data_models/molfeatures/molfeatures.py
LukasHebing: Hi Johannes,
in the benchmarks, I saw that the code below the statement takes ~5 min to compute, even if the `filter_descriptors` flag is set to `False`. A list of descriptors is computed, and then ignored. 

This is, why I added an early `return` before the expensive computation.

This is not related to this PR, though.

bofire/kernels/fingerprint_kernels/tanimoto_kernel.py
LukasHebing: For the kernel, we could do this!

uv.lock
LukasHebing: Nice, that you find all the accidentally committed files :)

bofire/data_models/molfeatures/molfeatures.py
jduerholt: Ok, this is fine for me. Maybe we also have to speed this part a bit up. I think there are more efficient ways to compute this.

PR #740, opened 2026-03-04
Nonlinear constraints to BoFire (integration with acqf + tests + tutorials)

MrzvskK:
## Motivation

This is an early attempt, I am sure there will be a few changes needed along the way. But to start with something, so Nonlinear constraints (`NonlinearInequalityConstraint` and `NonlinearEqualityConstraint`) were defined in BoFire's domain model but were not integrated into the optimization pipeline. The constraints were extracted from the domain but never passed to BoTorch's `optimize_acqf`, causing the optimizer to ignore them and propose infeasible candidates.
 
This PR:
1. Fixes the constraint integration pipeline in `acqf_optimization.py` to properly pass constraints to BoTorch
2. Corrects a critical bug where the `is_equality` flag was inverted (all constraints were treated as equality constraints)
3. Enhances `torch_tools.py` to properly format constraints for BoTorch compatibility
4. Adds feasibility verification in `botorch.py` to ensure proposed candidates satisfy all constraints
5. Provides tests and a tutorial demonstrating usage

Yes

I will do that later

### New Test Suite
Created `tests/bofire/strategies/test_nonlinear_constraints.py` with 16 comprehensive test cases:
 
**Constraint Types Tested:**
- Single and multiple inequality constraints (2-5 simultaneous)
- Equality constraints
- Mixed equality/inequality constraints
 
**Optimization Scenarios:**
- Different acquisition functions (qLogEI, qLogNEI, qUCB)
- Tight constraints (small feasible regions)
- Boundary conditions (initial data on constraint manifold)
- Single-variable constraints
- Empty feasible regions (proper error handling)
 
**Results:**
- All 16 nonlinear constraint tests passing
- 352/354 total strategy tests passing (99.4% pass rate)
- All integration tests passing
 
### Tutorial
Created `docs/tutorials/basic_examples/nonlinear_constraints_basic_usage.py` demonstrating:
- Competing reactions optimization with nonlinear constraints
- Constraint definition using string expressions
- Initial feasible sampling
- Multi-objective optimization with constraints
- Constraint verification utilities
 
### Verification Steps
To verify the changes work:
```bash
# Run nonlinear constraint tests
pytest tests/bofire/strategies/test_nonlinear_constraints.py -v
 
# Run full strategy test suite
pytest tests/bofire/strategies/ -v
 
# Run tutorial example - they also exist in .qmd - after revision we can delete .py files 
python docs/tutorials/advanced_examples/nonlinear_advanced.py
python docs/tutorials/advanced_examples/nonlinear_constraints_maximizing_yield.py

### CODE REVIEW

**Comment by LukasHebing** [comment]
Hi @MrzvskK,
nonlinear constraints are not handled by the standard BoFire optimizer, but the GA handles those:
https://experimental-design.github.io/bofire/docs/tutorials/advanced_examples/genetic_algorithm.html

Does this help?

**Comment by LukasHebing** [comment]
> @LukasHebing, this PR adds functionality to the BotorchOptimizer to handle nonlinear constraints.

Ok, great :)
I just wanted to make sure that this is not overlooked

**Comment by MrzvskK** [comment]
@jduerholt Hello, this is a quick reminder, I wanted to check if you approve the changes or not.

**Comment by jduerholt** [comment]
Ah sorry, did not see that the tests are now passing, I will review over the weekend or beginning of next week ;)

**Comment by MrzvskK** [comment]
@jduerholt  kind reminder, did you have time to take a look at the PR?

**Comment by jduerholt** [comment]
> @jduerholt kind reminder, did you have time to take a look at the PR?

Sorry, too much travelling, I will do it over the weekend or the latest on Monday.

File: bofire/data_models/constraints/nonlinear.py [hunk 1]
```diff
@@ -146,18 +171,102 @@ def set_hessian_expression(cls, hessian_expression, info) -> Union[str, Callable
 
         return hessian_expression
 
-    def __call__(self, experiments: pd.DataFrame) -> pd.Series:
+    def __call__(
```
jduerholt, line 174: From my perspective, this is highly botorch dependable code, for this reason, I propose to move it to `utils/torch_tools.py`, where also the other constraints are prepared to be botorch ready. Then we also do not need the option here to pass torch ensors into the function etc, and it is much cleaner to read. You can have a look there how the `ProductConstraint` is build up.

bofire/data_models/constraints/nonlinear.py
```diff
@@ -298,6 +407,31 @@ class NonlinearEqualityConstraint(NonlinearConstraint, EqualityConstraint):
 
     type: Literal["NonlinearEqualityConstraint"] = "NonlinearEqualityConstraint"
 
+    def is_fulfilled(self, experiments: pd.DataFrame, tol: float = 1e-6) -> pd.Series:
```
jduerholt, line 410: Honestly, I would not support them directly out of the box in bofire, as they are not naturally supported in botorch. If somebody wants to do something like this, he or she should do this breakdown into individual nonlinear constraints by hand.

bofire/data_models/strategies/predictives/acqf_optimization.py
```diff
@@ -138,14 +141,40 @@ def is_constraint_implemented(self, my_type: Type[constraints.Constraint]) -> bo
             constraints.NonlinearInequalityConstraint,
             constraints.NonlinearEqualityConstraint,
         ]:
-            return False
+            return True  # was False
```
jduerholt, line 144: I would only support the NonlinearInequalityConstraint and would the rest be handled by the user.

bofire/data_models/strategies/predictives/acqf_optimization.py
```diff
@@ -138,14 +141,40 @@ def is_constraint_implemented(self, my_type: Type[constraints.Constraint]) -> bo
...
+                        "Nonlinear constraints require n_restarts=1 "
+                        "to avoid parallel batch optimization. Overriding current value.",
+                    )
+                    object.__setattr__(self, "n_restarts", 1)
+
         def validate_local_search_config(domain: Domain):
             if self.local_search_config is not None:
                 if has_local_search_region(domain) is False:
-                    warnings.warn(
+                    logger.info(
```
jduerholt, line 177: Why this change from warning to logger, the logging mechanism in BoFire is somehow not well implemented, so I would like to keep it as warning for now. We need an overhaul there ...

bofire/data_models/strategies/predictives/acqf_optimization.py
```diff
@@ -138,14 +141,40 @@ def is_constraint_implemented(self, my_type: Type[constraints.Constraint]) -> bo
...
+                for c in domain.constraints
+            ):
+                if self.batch_limit != 1:
+                    logger.info(
+                        "Nonlinear constraints require batch_limit=1. "
+                        "Overriding current value.",
+                    )
+                    # Use object.__setattr__ to bypass Pydantic's frozen model behavior
+                    object.__setattr__(self, "batch_limit", 1)
+                if self.n_restarts != 1:
```
jduerholt, line 167: I do not think, that we need `n_restarts` ==1, it should also work with more restarts, but it will run them sequentially. So this can go, I think.

bofire/data_models/strategies/predictives/acqf_optimization.py
```diff
@@ -138,14 +141,40 @@ def is_constraint_implemented(self, my_type: Type[constraints.Constraint]) -> bo
             constraints.NonlinearInequalityConstraint,
             constraints.NonlinearEqualityConstraint,
         ]:
-            return False
+            return True  # was False
         return True
 
     def validate_domain(self, domain: Domain):
+        def validate_nonlinear_equality_constraints(domain: Domain):
```
jduerholt, line 148: No need for this, handle it as in `_get_optimizer_options` in `bofire/strategies/predictives/acqf_optimization.py`

bofire/strategies/doe/utils.py
jduerholt, line 1: Hmm, why these edits on the doe part?

bofire/strategies/predictives/acqf_optimization.py
```diff
@@ -434,6 +434,37 @@ def _optimize(
 
         return candidates
 
+    def _project_onto_nonlinear_constraints(
```
jduerholt, line 437: Why are we doing this? And in which case is it needed?

bofire/strategies/predictives/acqf_optimization.py
```diff
@@ -455,9 +486,54 @@ def _optimize_acqf_continuous(
             OptimizerEnum.OPTIMIZE_ACQF_MIXED: optimize_acqf_mixed,
             OptimizerEnum.OPTIMIZE_ACQF_MIXED_ALTERNATING: optimize_acqf_mixed_alternating,
         }
-        candidates, acqf_vals = optimizer_mapping[optimizer](
-            **optimizer_input.model_dump()
-        )
+        # Prefer principled retries over post-hoc projection/repair:
```
jduerholt, line 489: The botorch optimizer always starts from the feasible region, and it also automatically does retries if it fails, or? Currently it does only one retry, you can find the logic here: https://github.com/meta-pytorch/botorch/blob/c1aa7dd858b5959f39f23cfc591c568be5a08e3a/botorch/optim/optimize.py#L480 We can easily extend this to more restarts. What do you think? I also can file a botorch PR for this. I would go for this solution because it would keep the bofire implementation cleaner. 

For linear constraints, there is also an automatic repair function in botorch: https://github.com/meta-pytorch/botorch/blob/c1aa7dd858b5959f39f23cfc591c568be5a08e3a/botorch/optim/parameter_constraints.py#L722 I would extend this to nonlinear constraints. As the slsqp starts from a feasible point, the constraint violations should be minimal and the solution should be close to the feasible region, so I think the slsqp should be able to handle this. What do you think?

The ideal solution would be to linearize the nonlinear constraints in the botorch repair by a first order taylor exansion around the non-feasible candidate. Solve the problem only with the linear constraints, and iterate until the lonlinear constraints are fulfilled.

bofire/strategies/predictives/acqf_optimization.py
```diff
@@ -496,44 +572,439 @@ def _determine_optimizer(self, domain: Domain, n_acqfs) -> OptimizerEnum:
             return OptimizerEnum.OPTIMIZE_ACQF_MIXED
         return OptimizerEnum.OPTIMIZE_ACQF_MIXED_ALTERNATING
 
+    def _get_nonlinear_constraint_setup(
```
jduerholt, line 575: Hmm, I do not understand why this is necessary, currently we use our random strategy as initical candidates generator, and the random strategy can already handle nonlinear constraints via rejection sampling. Why not keep it as is?

bofire/strategies/predictives/acqf_optimization.py
```diff
@@ -496,44 +572,439 @@ def _determine_optimizer(self, domain: Domain, n_acqfs) -> OptimizerEnum:
...
-        input_preprocessing_specs = self._input_preprocessing_specs(domain)
         features2idx = self._features2idx(domain)
         inequality_constraints = get_linear_constraints(
             domain, constraint=LinearInequalityConstraint
         )
         equality_constraints = get_linear_constraints(
             domain, constraint=LinearEqualityConstraint
         )
-        if len(nonlinear_constraints := get_nonlinear_constraints(domain)) == 0:
+
```
jduerholt, line 845: I also do not understand this, this should all be handled in the original implementation using the random strategy as generator for the feasible candidates.

bofire/strategies/predictives/botorch.py
```diff
@@ -208,18 +209,48 @@ def _get_acqfs(self, n: int) -> List[AcquisitionFunction]:
     def has_sufficient_experiments(
         self,
     ) -> bool:
+        """Check if sufficient feasible experiments are available.
```
jduerholt, line 212: Hmm, I am not convinced that this is a good idea, from my perspective there is no need that past/historical experiments that were just used to train the surrogate models should fulfill the constraints. Often we have usecases, where new constraints are added over the course of the project, and this should keep working. 

What was your intention for adding?

bofire/strategies/random.py
```diff
@@ -70,45 +98,66 @@ def has_sufficient_experiments(self) -> bool:
         """
         return True
 
-    def _ask(self, candidate_count: PositiveInt) -> pd.DataFrame:
-        """Generate candidate samples using the random strategy.
-
-        If the domain is compatible with polytope sampling, it uses the polytope sampling to generate
-        candidate samples. Otherwise, it performs rejection sampling by repeatedly generating candidate
-        samples until the desired number of valid samples is obtained.
+    def _ask(self, candidate_count: int) -> pd.DataFrame:
```
jduerholt, line 101: Hmm, I do not get why this change was necessary, can you elaborate a bit?

bofire/utils/torch_tools.py
```diff
@@ -257,29 +259,95 @@ def product_constraint(indices: Tensor, exponents: Tensor, rhs: float, sign: int
...
         callables += get_product_constraints(domain)
 
+    # Handle NonlinearInequalityConstraint
+    if NonlinearInequalityConstraint in includes:
+        for constraint in domain.constraints.get(NonlinearInequalityConstraint):
+            # Create a wrapper that converts to the expected format
+            # The constraint should evaluate to <= 0
+            def make_constraint_callable(c):
+                def constraint_fn(x: Tensor) -> Tensor:
+                    # c.__call__ expects x with shape (batch_size, n_features)
```
jduerholt, line 317: I would move all the logic on handling torch tensors of different here.

jduerholt (commented):
Hi @MrzvskK,

thank you very much for the PR. I went over the actual implementation (ignoring the tests) for now and provided inline feedback. Regarding the botorch based suggestions, I also created an issue there: https://github.com/meta-pytorch/botorch/issues/3280

Best,

Johannes

bofire/data_models/strategies/predictives/acqf_optimization.py
MrzvskK: sorry, I forgot to revert it.

MrzvskK: will do that now.

bofire/strategies/predictives/acqf_optimization.py
MrzvskK: decided waiting for response for BoTorch team.

bofire/strategies/predictives/acqf_optimization.py
MrzvskK: The motivation for _get_nonlinear_constraint_setup is that once we pass nonlinear_inequality_constraints to BoTorch, it validates feasibility in tensor space with its own convention/tolerances, and ICs produced via BoFire rejection sampling can still be rejected there. If we decide not to pass nonlinear constraints into BoTorch and rely purely on BoFire’s rejection sampling, then I agree we can drop most of this setup and keep the implementation simpler. I will double check the RandomStrategy, since it might be doing the same thing.

MrzvskK: I will explore get_initial_conditions_generator

bofire/strategies/predictives/botorch.py
MrzvskK: I can revert this change.

bofire/strategies/predictives/acqf_optimization.py
MrzvskK: would you say it should be part of the [3280](https://github.com/meta-pytorch/botorch/issues/3280) in BoTorch?

bofire/data_models/constraints/nonlinear.py
MrzvskK: update to PR has this logic moved toe torch_tools.py

bofire/data_models/constraints/nonlinear.py
MrzvskK: as per discussion with Chryssa we keep the nonlinearequality logic

bofire/data_models/strategies/predictives/acqf_optimization.py
MrzvskK: same as above, in the joint discussion we decided to keep them

bofire/data_models/strategies/predictives/acqf_optimization.py
MrzvskK: included in the update

bofire/data_models/strategies/predictives/acqf_optimization.py
MrzvskK: Agreed — n_restarts > 1 is fine; restarts run sequentially. Removed the override that forced n_restarts=1 for nonlinear constraints.

bofire/data_models/strategies/predictives/acqf_optimization.py
MrzvskK: Removed validate_nonlinear_equality_constraints from the data model. batch_limit=1 for nonlinear constraints is now handled in _get_optimizer_options in bofire/strategies/predictives/acqf_optimization.py, consistent with how NChooseK/Product constraints are handled.

bofire/strategies/doe/utils.py
MrzvskK: good catch, it was an unrelated change, reverted back

bofire/strategies/predictives/acqf_optimization.py
MrzvskK: we will keep this logic for now and change after botorch approves changes.

bofire/strategies/predictives/acqf_optimization.py
MrzvskK: For this PR review we’re keeping the feasibility check as a temporary solution until upstream BoTorch configuration follows. I’ll remove this logic from bofire once BoTorch implements it

bofire/strategies/predictives/acqf_optimization.py
MrzvskK: Yes the RandomStrategy would be prefered for sampling and originally tried get_initial_conditions_generator(RandomStrategy), but it currently does not work with the feasibility check internal logic. 

Currently if we keep the feasibility sampling logic the custom IC generator would oversample/reject before optimization start. 

Once the #3280 functionality in BoTorch will be added we will move this part upstream and update it back to using RandomStrategy.  Until then we need to keep it to avoid regressions on tight nonlinear constraints.

bofire/strategies/predictives/acqf_optimization.py
MrzvskK: same as above it's a temporary bridge until upstream repair/retries

bofire/strategies/predictives/botorch.py
MrzvskK: reverted in the PR update

bofire/strategies/random.py
MrzvskK: i see now the _ask re-branching wasn't needed. I've reverted to main's flow (polytope when linear-type only, else rejection via _sample_with_nchooseks + is_fulfilled, which already handles nonlinear and mixed domains).

Only kept two small fixes on top of main: samples[valid.values] (pandas indexing) and a relaxed tol in is_fulfilled for nonlinear numerics.

bofire/utils/torch_tools.py
MrzvskK: yes, already addressed

MrzvskK:
@jduerholt can you re-run Quarto check? it keeps getting cancelled

jduerholt:
> @jduerholt can you re-run Quarto check? it keeps getting cancelled

I think the nonlinear constraint notebook takes too long, and then it is aborting. Maybe you can trim it down or add a smoketest as in others? Also the botorch folks approved your PR, I would then recommend to integrate in this PR directly the botorch changes to not add stuff this is removeed later. `testing_against_latest_botorch` is then the test pipeline to watch.

MrzvskK:
@jduerholt not really, It was passing before, but then i pushed one commit before Quarto finished and it started to go sideways (seems like it's one run behind). If you can just rerun Quarto test it should work. I think even if i push a new commit it will cancel again

bofire/strategies/predictives/acqf_optimization.py
jduerholt: Hmm, why would the curstom IC generator obversample/reject before otpimization start?

jduerholt: If there is a  problem with the RandomStrategy, then we should fix it there, or?

jduerholt:
> @jduerholt not really, It was passing before, but then i pushed one commit before Quarto finished and it started to go sideways (seems like it's one run behind). If you can just rerun Quarto test it should work. I think even if i push a new commit it will cancel again

I just restarted. Nice that the botorch stuff is reviewed. Then this will move forward ;)

MrzvskK:
> > @jduerholt not really, It was passing before, but then i pushed one commit before Quarto finished and it started to go sideways (seems like it's one run behind). If you can just rerun Quarto test it should work. I think even if i push a new commit it will cancel again
> 
> I just restarted. Nice that the botorch stuff is reviewed. Then this will move forward ;)

Yes for sure! There were some failed test and one issue (i fixed the lint error from my side), but it think most of the tests were failing also upstream, let's see, hopefully it will move forward quicker

bofire/strategies/predictives/acqf_optimization.py
MrzvskK: Let me investigate this a bit!

MrzvskK: ok, so looks like RandomStrategy produces feasible samples in dataframe space (bofire style), but BoTorch validates batch_initial_conditions in tensor space using its own feasibility tolerance (see atol inside gen_batch_initial_conditions). These are different checks:

1.  BoFire's rejection sampling uses tol=1e-3 for NChooseK/nonlinear constraints.
2. BoTorch's internal initial condition validation uses a stricter tolerance on the transformed tensor, and can silently drop or error on initial conditions that bofire would see as feasible.

So in case of NonlinearEqualityConstraint, the feasible set is a manifold (measure zero in R^n), so rejection sampling almost never produces points that pass BoTorch atol, hence we need feasible_ic_generator before passing ICs to BoTorch.

For NonlinearInequalityConstraint we already sidestep this by not passing them into BoTorch at all so there is no IC generator override.

MrzvskK: but on the second thought i can simplify this to be the same way we handle nonlinearinequality - since we represent Nonlinear Equality as two NonlinearInequalities

MrzvskK:
@jduerholt seems like there is a bug in `tests/bofire/strategies/test_space_filling.py::test_ask[domain5-16]` which fails intermittently (~1 in 3 runs) with: ConstraintNotFulfilledError: Constraints not fulfilled — multiple features are 
non-zero when NChooseKConstraint(max_count=1) should allow at most one.

Example failing rows: if1=0.812, if2=0.229, if3=0.712 (3 non-zero, max_count=1)

To check: pytest tests/bofire/strategies/test_space_filling.py::test_ask[domain5-16] -v
Running 3-5 times; fails ~1/3 of runs

Probable cause: DoEStrategy uses 'ipopt' to optimize a space-filling criterion subject to the NChooseK constraint. When the 'ipopt' initial point is random and doesn't satisfy NChooseK, 'ipopt' occasionally converges to a locally infeasible solution. DoEStrategy._validation_tol=1e-4  then correctly catches this.

This then causes intermittent CI failures when cyipopt is available. A potential fix here would be to mark the test case as xfail(strict=False). Or i can investigate more solutions. 

Should i open an issue for this or just make a new PR?

bofire/strategies/predictives/acqf_optimization.py
MrzvskK: ok, i made a simpler version, but then also found some inconsistencies in handling NChooseK, which is addressed here too, but i will need to make more changes once botorch feature goes through.

jduerholt:
> Should i open an issue for this or just make a new PR?

Thanks for catching this, just  go for a new PR here. cc: @dlinzner-bcs

MrzvskK:
@jduerholt 
Already made two new PRs there is also an issue with new release of pydantic from yesterday, tagged you and I will tag @dlinzner-bcs as well

MrzvskK:
@jduerholt seems like all the checks pass now (assuming the pydantic update fix) can we merge now? When botorch goes through I will update the worktree

MrzvskK:
Hello @jduerholt, I hope your summer went well, do we have any news on this PR? Should I include the Botorch feature now?

PR #747, opened 2026-03-26
Add BONSAI-style greedy pruning for NChooseK constraints

jduerholt:
## Motivation

BONSAI for handling of NChooseK constraint (https://arxiv.org/abs/2602.07144).

@dlinzner-bcs: one could do the same in the DOE module.

Yes.

Not yet.

## Test Plan

Unit tests.


jduerholt:
@Jimbo994, if you are interested ;)

bofire/strategies/predictives/botorch.py
```diff
@@ -201,6 +209,345 @@ def _ask(self, candidate_count: int) -> pd.DataFrame:
...
+
+        # Transform candidates to tensor
+        transformed = self.domain.inputs.transform(
+            candidates, self.input_preprocessing_specs
+        )
+        X = torch.from_numpy(transformed.values).to(**tkwargs)  # (q, d)
+
+        # Precompute linear constraints and bounds for local optimize_acqf
+        if use_qp:
+            from bofire.strategies import utils
```
jduerholt: Why is this import here as lazy import?

bofire/strategies/predictives/botorch.py
```diff
@@ -201,6 +209,345 @@ def _ask(self, candidate_count: int) -> pd.DataFrame:
...
+                variants[k] = projected
+                variant_af_vals[k] = -float("inf")
+                continue
+
+            # Local optimize_acqf with fixed_features={k: 0}
+            # Use QP-projected point as initial condition
+            fixed_features = {idx: 0.0}
+            batch_initial = projected.unsqueeze(0).unsqueeze(0)  # (1, 1, d)
+
+            try:
```
jduerholt: We could massively speed this up, using bach optimization, or? We can bundle every pruning candidates from the QP projections into a batch of initial candiation and optimize over them in parallel.

bofire/benchmarks/benchmark.py
jduerholt: We need to adapt the docstring here regarding the nchoosek as we are now applying it globally.

bofire/data_models/domain/features.py
```diff
@@ -837,6 +838,9 @@ def get_bounds(
             then the local bounds based on a local search region are provided as reference to the
                 reference experiment. Currently only supported for continuous inputs.
                 For more details, it is referred to https://www.merl.com/publications/docs/TR2023-057.pdf. Defaults to None.
+            relax_allow_zero (bool, optional): If True, semi-continuous continuous inputs
+                (`allow_zero=True` with positive lower bound) report a relaxed lower bound of 0.
+                Other input types ignore this flag. Defaults to False.
```
jduerholt, line 853: Is this a good idea to propagate it everywhere? Could we not have this via kwargs or so to not have to introduce this relax_allow_zero flag everywhere?

bofire/data_models/features/continuous.py
```diff
@@ -221,14 +221,27 @@ def get_bounds(
         transform_type: Optional[TTransform] = None,
         values: Optional[pd.Series] = None,
         reference_value: Optional[float] = None,
+        relax_allow_zero: bool = False,
     ) -> Tuple[List[float], List[float]]:
         assert transform_type is None
         if reference_value is not None and values is not None:
             raise ValueError("Only one can be used, `local_value` or `values`.")
 
+        # Effective lower bound: 0 for semi-continuous features when the
```
jduerholt, line 253: This should be tested, there are also tests that test local bounds, we should do the test for the effective lower.

bofire/strategies/predictives/_nchoosek_pruning.py
```diff
@@ -0,0 +1,1600 @@
...
+
+class PruningInfeasibleError(RuntimeError):
+    """Raised when greedy pruning cannot satisfy every NChooseK and
+    semi-continuity constraint — typically because the per-constraint
+    ``min_count`` guard empties the action set before all
+    ``max_count`` constraints are met.
+    """
+
+
+class ActionKind(Enum):
```
jduerholt, line 67: Add to the docstring, what the actions actually mean, Active and Activate sound very similar.

bofire/strategies/predictives/_nchoosek_pruning.py
```diff
@@ -0,0 +1,1600 @@
...
+    drop-in replacement for the legacy ``(af_red, 0/1, j)`` ordering.
+    """
+
+    ZERO = 0
+    ACTIVE = 1
+    ACTIVATE = 2
+
+
+@dataclass
+class Action:
```
jduerholt, line 98: Document the attributes.

bofire/strategies/predictives/_nchoosek_pruning.py
```diff
@@ -0,0 +1,1600 @@
...
+    variant: Tensor
+    valid: bool
+
+
+@dataclass(frozen=True)
+class PruningContext:
+    """All static-per-call inputs to the pruning loop.
+
+    Bundling these reduces the parameter-count of the inner helpers
+    (variant builders, action collectors) from ~10 to ~3 and gives
```
jduerholt: It is not necessary to give this comments on the history of the fatures.

bofire/strategies/predictives/_nchoosek_pruning.py
```diff
@@ -0,0 +1,1600 @@
...
+    a single source of truth for the per-call configuration.
+
+    ``pinned_columns`` lists tensor columns that must be frozen at
+    the candidate's per-row value throughout pruning — for example,
```
jduerholt: Also document the attributes as we do it in the rest of the codebase.

bofire/strategies/predictives/_nchoosek_pruning.py
```diff
@@ -0,0 +1,1600 @@
...
+    domain: Domain,
+    features2idx: Dict[str, Tuple[int, ...]],
+) -> Dict[int, Tuple[float, float]]:
+    """Build the ``semicontinuous_specs`` mapping for ``prune_nchoosek``.
+
+    Returns a dict ``{tensor_column_index: (lb, ub)}`` for every continuous
+    input with ``allow_zero=True`` and a positive lower bound.
+    """
+    specs: Dict[int, Tuple[float, float]] = {}
+    for feat in domain.inputs.get(ContinuousInput):
```
jduerholt, line 283: We do this check at several ocassions, why not have a method is_semicontinuous in ContinuousInput?

bofire/strategies/predictives/_nchoosek_pruning.py
```diff
@@ -0,0 +1,1600 @@
...
+    """
+    if is_nchoosek_pruning_applicable(domain):
+        return True
+    if not has_semicontinuous_features(domain):
+        return False
+
+    blocking = _features_in_blocking_constraints(domain)
+    for feat in domain.inputs.get(ContinuousInput):
+        assert isinstance(feat, ContinuousInput)
+        if feat.allow_zero and feat.bounds[0] > 0 and feat.key in blocking:
```
jduerholt: also here we can use the is_semicontinuous thing

bofire/strategies/predictives/_nchoosek_pruning.py
```diff
@@ -0,0 +1,1600 @@
...
+            return False
+    return True
+
+
+# ---------------------------------------------------------------------------
+# State helpers
+# ---------------------------------------------------------------------------
+
+
+def _classify_features_for_row(
```
jduerholt, line 323: I am not sure if this method is correct as we are also classifiyng non continuous features, features outside of constraints etc, or? Or does this not matter due to the pinned ones? And why are we not pinning also everything that is not semi continuous, not nchoose k or not something that is involved in a linear constraint whre also at least one feature involved nchoosek/semicontinuous is involved.

bofire/strategies/predictives/_nchoosek_pruning.py
```diff
@@ -0,0 +1,1600 @@
...
+    counts: Dict[int, int] = {}
+    for c_idx, c in enumerate(nchoosek_constraints):
+        indices: List[int] = []
+        for feat_key in c.features:
+            indices.extend(features2idx[feat_key])
+        counts[c_idx] = int((x[indices].abs() > tol).sum().item())
+    return counts
+
+
+def _violated_constraints(
```
jduerholt: Should we not name this _max_count_violated_constraints?

bofire/strategies/predictives/_nchoosek_pruning.py
```diff
@@ -0,0 +1,1600 @@
...
+) -> Dict[int, int]:
+    """Per-constraint active counts, keyed by constraint position."""
+    counts: Dict[int, int] = {}
+    for c_idx, c in enumerate(nchoosek_constraints):
+        indices: List[int] = []
+        for feat_key in c.features:
+            indices.extend(features2idx[feat_key])
+        counts[c_idx] = int((x[indices].abs() > tol).sum().item())
+    return counts
+
```
jduerholt, line 412: In general the helpers below have often something mechanism in common, maybe one can furhter streamline it.

bofire/strategies/predictives/_nchoosek_pruning.py
```diff
@@ -0,0 +1,1600 @@
...
+            acqf,
+            X_pending_extra=X_pending_extra,
+        )
+        if refined is not None:
+            return refined, True
+
+    return projected, True
+
+
+def _build_zero_variant(
```
jduerholt: I think this wrappers are not really helpful, I think we can delete them, or? Just blowing up the module.

bofire/strategies/predictives/_nchoosek_pruning.py
```diff
@@ -0,0 +1,1600 @@
...
+        per_step_local_reopt=per_step_local_reopt,
+        pinned_zero_indices=pinned_zero_indices,
+        fixed_features=fixed_features,
+        active_set=active_set,
+        semicontinuous_specs=semicontinuous_specs,
+        tol=tol,
+    )
+
+
+def _collect_actions(
```
jduerholt, line 886: Some code replication here, but I do not know if it makes sense to try to trim it down,

bofire/strategies/predictives/_nchoosek_pruning.py
```diff
@@ -0,0 +1,1600 @@
...
+        PruningInfeasibleError: if the greedy loop cannot satisfy all
+            constraints (typically because the ``min_count`` guard
+            empties the action set before the ``max_count`` constraints
+            are met).
+    """
+    # Defensive guard: NChooseK features must each map to a single tensor
+    # column. The NChooseKConstraint validator already restricts to
+    # ContinuousInput, but the optimizer's input_preprocessing_specs could
+    # in principle multi-encode an ill-typed feature; fail loudly if so.
+    for c in nchoosek_constraints:
```
jduerholt: This check can be removed, it is guranteed by bofire.

bofire/strategies/predictives/_nchoosek_pruning.py
```diff
@@ -0,0 +1,1600 @@
...
+                    f"Pruning requires every NChooseK feature to map to a "
+                    f"single tensor column. Feature {feat_key!r} maps to "
+                    f"{cols} via features2idx — this typically means it "
+                    f"was one-hot or otherwise multi-column encoded, which "
+                    f"is out of scope for the BONSAI pruning module."
+                )
+
+    X = X.clone()
+    q = X.shape[0]
+    if q == 0:
```
jduerholt: Remove this, does not make sense.

bofire/strategies/predictives/acqf_optimization.py
```diff
@@ -379,8 +391,19 @@ def _optimize(
         domain: Domain,
         experiments: Optional[pd.DataFrame] = None,
     ) -> pd.DataFrame:
+        # NChooseK + LSR-BO and semi-continuous + LSR-BO are both
```
jduerholt: Remove this comment

bofire/strategies/predictives/acqf_optimization.py
jduerholt: can we not also do this via get_torch_bounds_from_domain, or extend it?

bofire/strategies/predictives/acqf_optimization.py
```diff
@@ -397,6 +420,17 @@ def _optimize(
             acqfs=acqfs,
             bounds=bounds,
         )
+        # print(candidates)
+
+        if pruning_applicable:
```
jduerholt, line 422: This reads a bit misleading, or doubled, can we not just do _prune_if_applicable, or we clarify the names and calle the method only _prune.

jduerholt: I think if would go for renameing to _prune

bofire/strategies/predictives/acqf_optimization.py
```diff
@@ -487,11 +594,30 @@ def _determine_optimizer(self, domain: Domain, n_acqfs) -> OptimizerEnum:
         n_categorical_combinations = (
             domain.inputs.get_number_of_categorical_combinations()
         )
+        # When pruning is applicable, semi-continuous features
```
jduerholt, line 617: let us move this into the function in the inputs with a flag, much cleaner to read and make it easy to reuse.

bofire/strategies/random.py
```diff
@@ -115,6 +117,106 @@ def _ask(self, candidate_count: PositiveInt) -> pd.DataFrame:
             n_iters += 1
         return pd.concat(valid_samples, ignore_index=True).iloc[:candidate_count]
 
+    @staticmethod
```
jduerholt: Let us remove this from the current PR, it is bloating it and just rely on the try except around the polytope sampler, which is exactly doing the same but implicitly using botorch. What do you think?

bofire/utils/torch_tools.py
```diff
@@ -55,6 +55,31 @@
 }
 
 
+def get_torch_bounds_from_domain(
+    domain: Domain,
+    input_preprocessing_specs: InputTransformSpecs,
+    relax_allow_zero: bool = False,
```
jduerholt, line 61: here we could also build in the local bounds.

bofire/utils/torch_tools.py
```diff
@@ -258,12 +283,15 @@ def product_constraint(indices: Tensor, exponents: Tensor, rhs: float, sign: int
 def get_nonlinear_constraints(
```
jduerholt, line 289: Why this strange API here? with include and exclude_nchoosek that overrides the include?

tests/bofire/data_models/domain/test_domain.py
```diff
@@ -549,3 +549,9 @@ def test_domain_to_description_with_constraints():
         "\n## Constraints (candidates MUST satisfy all of these)\n"
         "- 1.0*x1 + 1.0*x2 <= 1.5 — Budget constraint"
     )
```
jduerholt, line 551: delete this comment

tests/bofire/strategies/test_pruning_gate.py
```diff
@@ -0,0 +1,367 @@
+"""Tests for the domain-level pruning applicability gates.
```
jduerholt: we are testing here sutff from the pruining module, should be in the same as the rest of test regarding this module

jduerholt (commented):
Nothing big, small improvements.

jduerholt:
Hi @bertiqwerty, @LukasHebing @TobyBoyne,

this is the new way of handling NChooseKs that I was cooking for the last month. First benchmarks show very superior performance in comparison to the original implementation via nonlinear constraints. Furthermore, it is much faster.

It would be nice if you all can have a look at this and review it. It would be nice if we can land it soon as it is from my perspective a big enhancements, as it now also fully supports semicontinuous features in optimization etc.

There is also much room for further improvement as outlined in the markdown file, I especially like the beam search but we can do it later ;) 

Are you also interested in the benchmark results?

Best,

Johannes

bofire/data_models/domain/features.py
TobyBoyne: Maybe we could consider (in another PR) having some kind of context manager that can handle this kind of thing? Similar to [gpytorch settings](https://docs.gpytorch.ai/en/latest/settings.html). So I would imagine it could be used as

```python
with bofire.optimization_settings.relax_allow_zero(True):
    strategy.ask(1)
```

This would avoid passing it to all these functions. The functions can then access the value of `bofire.optimization_settings.relax_allow_zero` to determine whether to relax.

bofire/strategies/predictives/_nchoosek_pruning.py
```diff
@@ -0,0 +1,1326 @@
...
+
+
+def is_pruning_applicable(domain: Domain) -> bool:
+    """Unified gate: pruning runs if either NChooseK pruning is
+    applicable, or the domain has standalone semi-continuous features —
+    and no semi-continuous feature appears in a blocking nonlinear /
+    interpoint constraint.
+    """
+    if is_nchoosek_pruning_applicable(domain):
+        return True
```
TobyBoyne: I think this might have issues if you have semicontinuous features that overlap with blocking constraints *as well as* NChooseKConstraints that don't overlap with blocking constraints? Since this early return will skip the checks that happen a few lines below in this same function.

bofire/strategies/predictives/_nchoosek_pruning.py
```diff
@@ -0,0 +1,1326 @@
...
+            ``{col: x_i[col]}`` for every ``col`` in
+            :attr:`PruningContext.pinned_columns`. Computed once at row
+            entry and forwarded to :func:`_build_variant`,
+            :func:`_local_optacqf`, and :func:`_final_local_reopt`.
+    """
+
+    x: Tensor
+    zero_set: Set[int] = field(default_factory=set)
+    frac_set: Set[int] = field(default_factory=set)
+    active_set: Set[int] = field(default_factory=set)
```
TobyBoyne, line 184: (Nit) I don't love this structure, since to me it implies that a feature index can be in multiple of these sets at once. My understanding is that each column is exactly one of `{zero, fractional, active}`, so this data structure should surely be a dictionary `column_state: dict[int, Literal[Zero | Fractional | Active]]`?

bofire/strategies/predictives/_nchoosek_pruning.py
```diff
@@ -0,0 +1,1326 @@
...
+                continue
+        active.add(j)
+    return zero, fractional, active
+
+
+def _is_nchoosek_fulfilled(
+    x: Tensor,
+    nchoosek_constraints: Sequence[NChooseKConstraint],
+    features2idx: Dict[str, Tuple[int, ...]],
+    tol: float = 1e-6,
```
TobyBoyne: Remove default here, no other functions have a default tolerance

bofire/strategies/predictives/_nchoosek_pruning.py
TobyBoyne: I agree with this - not necessary but would be easier to follow I think.

bofire/strategies/predictives/_nchoosek_pruning.py
```diff
@@ -0,0 +1,1326 @@
...
+    return violated
+
+
+def _action_violates_count_bound(
+    j_idx: int,
+    active_counts: Dict[int, int],
+    nchoosek_constraints: Sequence[NChooseKConstraint],
+    features2idx: Dict[str, Tuple[int, ...]],
+    *,
+    delta: int,
```
TobyBoyne: It would be clearer to me if this function accepted an `ActionKind` enum, and worked out delta from that. At the very least, `delta` is of the type `Literal[-1, 1]`.

bofire/strategies/predictives/_nchoosek_pruning.py
```diff
@@ -0,0 +1,1326 @@
...
+    Exposed via the two ``partial`` specialisations below
+    (:data:`_zero_action_blocked_by_min_count` and
+    :data:`_activate_action_blocked_by_max_count`).
+    """
+    for c_idx, c in enumerate(nchoosek_constraints):
+        constraint_indices: Set[int] = set()
+        for feat_key in c.features:
+            constraint_indices.update(features2idx[feat_key])
+        if j_idx not in constraint_indices:
+            continue
```
TobyBoyne: (Nit) I would rewrite this whole block as below - to me, it's clearer what the code below is doing.
```python
if j_idx not in map(features2idx.get, c.features):
   continue
```

bofire/strategies/predictives/_nchoosek_pruning.py
```diff
@@ -0,0 +1,1326 @@
...
+
+    Honours ``none_also_valid``: if a constraint allows
+    ``count == 0`` and the current count is 0, it is *not* reported
+    as violated.
+    """
+    violated: Set[int] = set()
+    for c_idx, c in enumerate(nchoosek_constraints):
+        a_c = active_counts[c_idx]
+        if a_c < c.min_count and not (c.none_also_valid and a_c == 0):
+            violated.add(c_idx)
```
TobyBoyne: This pattern appears often enough, could it be put into `NChooseKConstraint`? Just a helper function like:

```python
def count_is_valid(self, count: int) -> bool:
    return count <= self.max_count and (count >= self.min_count or (count == 0 and self.none_also_valid)
```

Then this whole function is just below, which to me is more readable.
```python
return {
    c_idx for c_idx, c in enumerate(nchoosek_constraints)
    if c.count_is_valid(active_counts[c_idx])
}
```

bofire/strategies/predictives/_nchoosek_pruning.py
```diff
@@ -0,0 +1,1326 @@
...
+# are documented as a separate appendix in ``docs/pruning.md``. They
+# all build on the greedy machinery below (action set, guards, variant
+# builders, ``_collect_actions``); the greedy stays the default code
+# path.
+# ---------------------------------------------------------------------------
+
+
+_OPTIMIZE_ACQF_DEFAULTS: Dict[str, Any] = {
+    "q": 1,
+    "num_restarts": 1,
```
TobyBoyne, line 625: Why is the default for `"num_restarts"` 1?

bofire/strategies/predictives/_nchoosek_pruning.py
```diff
@@ -0,0 +1,1326 @@
...
+    try:
+        yield
+    finally:
+        # ``set_X_pending`` succeeded above so this won't raise; keep
+        # the try/except as a defensive net for third-party AFs that
+        # might behave asymmetrically.
+        try:
+            acqf.set_X_pending(saved)
+        except UnsupportedError:
+            pass
```
TobyBoyne: This is very tough to parse - nested try/except, `return` in except followed by a `finally` clause, `yields` and `returns`. Please consider if there is a better way to write this function. I also can't work out the purpose of this function at all. Comparing this module to `botorch.optim.optimized_mixed`, they don't have any kind of special treatment like this. Is it definitely necessary?

TobyBoyne (changes requested):
Looks good, and great that we're seeing fast and strong performance!

Most of my comments are about coding style, which you may want to ignore - if the code works, it works! But given there are plans to continue working on this and developing it, I think there is value in keeping the code clean. There is a lot of code duplication, and a few things are written in quite unintuitive ways.

The one thing I really don't understand is all of the `X_pending` stuff. I can't see why we need to do anything non-standard here. I would have thought BONSAI is independent to pending points in the acqf optimization. If you have any further explanation, or there is something in the BONSAI paper that explains it, I would appreciate it :)

bofire/strategies/predictives/_nchoosek_pruning.py
jduerholt: Good catch, I fixed it.

bofire/strategies/predictives/_nchoosek_pruning.py
jduerholt: Interesting point, but I would like to keep the current structure as it, as we are making heavy use of this data structure in downstream code and this would create some heaviy changes.

bofire/strategies/predictives/_nchoosek_pruning.py
jduerholt: Done.

bofire/strategies/predictives/_nchoosek_pruning.py
jduerholt: Changed it to the the recommended `Literatl[-1,1]`.

bofire/strategies/predictives/_nchoosek_pruning.py
jduerholt: Good point, done.

bofire/strategies/predictives/_nchoosek_pruning.py
jduerholt: Because we only do a local reoptimization starting from the pruned candidate.

bofire/strategies/predictives/_nchoosek_pruning.py
jduerholt: it is much simpler now, it is not necessary for us, as we do not support analytic acqfs in Bofire.

TobyBoyne:
Thanks for responding to the reviews @jduerholt, no more changes requested! And yes regarding the the acqf optimization, even if BoFire did support analytic acqfs they don't support `q>1` anyway. I still think that function is unecessary, since the `botorch` acquisition function optimizers don't mind modifying the acqf objects in-place, but happy to leave it.

I would be very interested to see the benchmark results! What benchmarks are you using? I don't think I have seen many NChooseK constraint problems in the BO literature.

bofire/strategies/predictives/_nchoosek_pruning.py
jduerholt: This has been also updated due to the new helpers to reduce code duplication.

PR #748, opened 2026-04-04
register LKJPrior if it's set.

Jimbo994:
## Motivation

closes #745 

now the LKJPrior is registered if it is specified as a task prior. 

Yes

Yes

## Test Plan

Added tests.

### DIFF
--- a/bofire/surrogates/multi_task_gp.py
+++ b/bofire/surrogates/multi_task_gp.py
@@ -1,4 +1,3 @@
-import warnings
 from typing import Dict, Optional
 
 import botorch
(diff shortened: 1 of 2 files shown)


Jimbo994:
@jpfolch can you have a doublecheck if this correctly sets the LKJPrior? If yes @jduerholt can you review this?

jduerholt (approved):
Looks good to me in general, but do you know if the original issue is fixed? The one with the LKJ Prior mentioned in the comment?

Jimbo994:
Yes, so I now test that the LKJPrior is set, and test if the model is successfully fitted using `model.fit`, which uses `fit_gpytorch_mll`, and this works. So with that I assume it is fixed.

jpfolch:
Here's the [original issue](https://github.com/meta-pytorch/botorch/issues/1860) which remains open, with some code snippets that claim to fix it [here](https://github.com/meta-pytorch/botorch/discussions/2656#discussioncomment-11785034), but not sure if it was merged. Maybe it was fixed in some other PR but I can't find where exactly.

Jimbo994:
Ah, thanks for the context @jpfolch, I just checked, and it is indeed not solved/merged! If the model fits correctly on the first try, no error is thrown. If it doesn't, then `sample_all_priors` is called, and this still throws an error. When I ran the test I added 20 times, it sometimes fails the first fit and then indeed throws this error.

Let's maybe leave this PR open and keep an eye out for fixes in botorch.

PR #750, opened 2026-04-09
remove redundant PieceWiseLinearGP add ExactW Kernels

Jimbo994:
## Motivation

With the advent of the engineered features, we can now, in principle, get rid of the `PieceWiseLinearGP` because it can be fully recreated using the `InterpolateFeature` engineered feature. So, for future maintenance, it is better to remove the now-redundant PieceWiseLinearGP and accompanying code from, for instance, `torch_tools.py` and other locations.

In the meantime, I also worked on exact implementations of the `WassersteinKernel`. The `WassersteinKernel` was computed on an interpolated grid. The `ExactWassersteinKernel` considers the union of unique x-values and then uses the trapezoidal rule, together with identifying the sign of break points, to perform the computation exactly. Both W1 and W2 distances are supported. The W1 implementation matches the approximate `WassersteinKernel` at an adequately high number of interpolation points.

The union of unique x-values can become quite large when having high-dimensional problems, a large number of raw samples/restarts in the acq. func. optimization, etc. Which on moderate compute (e.g., small k8 compute), can lead to memory issues. Therefore, there is also some chunking in the code so that we loop over the union. For CPU this was also slighly faster. Thoughts on this implementation @jduerholt?

Lastly, fixed a bug in the interp1d, which could lead to gradfNone errors due to division by zero. In principle, this should never happen, as inequality constraints are set, but initial candidates provided by Botorch sometimes violate these.

Yes

Yes 

## Test Plan

Removed old tests.
Created new tests for new features.
Ran tests.


bofire/data_models/surrogates/trainable_botorch.py
```diff
@@ -12,9 +12,10 @@ class TrainableBotorchSurrogate(BotorchSurrogate, TrainableSurrogate):
     @model_validator(mode="after")
     def validate_scaler_features(self):
         if self.scaler and len(self.scaler.features) > 0:
-            missing_features = list(
-                set(self.scaler.features) - set(self.inputs.get_keys())
-            )
+            known_keys = set(self.inputs.get_keys())
+            if hasattr(self, "engineered_features"):
```
jduerholt: It always has the attribute, or? Only in most cases it is empty, so I think you do not need the eck here. The engineered features object exists, you can call get_keys in will get back en empty list if not there.

jduerholt: And can you not kust take the sum of the two lists coming from get_keys?

bofire/kernels/shape.py
```diff
@@ -20,8 +187,122 @@ def forward(
         diag: bool = False,
         last_dim_is_batch: bool = False,
     ) -> Tensor:
-        dists = self.calc_wasserstein_distances(x1, x2)
-        dists = dists / self.lengthscale
+        x1_scaled = x1 / self.lengthscale
```
jduerholt, line 190: Why did you change the implementation here?

bofire/data_models/kernels/shape.py
```diff
@@ -27,3 +27,22 @@ class WassersteinKernel(Kernel):
     type: Literal["WassersteinKernel"] = "WassersteinKernel"
     squared: bool = False
     lengthscale_prior: Optional[AnyPrior] = None
+    lengthscale_constraint: Optional[AnyPriorConstraint] = None
+
+
+class ExactWassersteinKernel(FeatureSpecificKernel):
```
jduerholt, line 33: Could you add here documentation on what the attributes mean?

tests/bofire/data_models/specs/surrogates.py
```diff
@@ -78,6 +76,42 @@
     },
 )
 
+specs.add_valid(
```
jduerholt, line 79: Why this test?

jduerholt (changes requested):
Looks really, really good to me. I love it. Only some minor comments.

bofire/kernels/shape.py
Jimbo994: was a bit inspired by how gpytorch implements kernels: https://github.com/cornellius-gp/gpytorch/blob/main/gpytorch/kernels/rbf_kernel.py

I was thinking it is maybe a bit more stable.

jduerholt: Ok, then we just keep it ;)

bofire/data_models/kernels/shape.py
Jimbo994: done.

bofire/data_models/surrogates/trainable_botorch.py
Jimbo994: yes you are right on both, I corrected both.

Jimbo994:
How about this @jduerholt

jduerholt (approved):
Looks good to me. Thanks!

PR #751, opened 2026-04-13
Add Pathwise Thompson Sampling Acquisition Function

jduerholt:
## Motivation

This PR adds the pathwise Thompson Sampling ACQF to BoFire which is based on Matheron Samples.

Yes.

Yes.

## Test Plan

Unit tests, currently failing since it is not yet in BoFire main, so we should wait with merging it until it is in BoFire main. But we can review it already before.


jduerholt:
@bertiqwerty: can you review this? Failing tests are due to the botorch version ...

bertiqwerty:
Why are all tests failing? Not only against latest botorch?

jduerholt:
@bertiqwerty: the test were failing as we are using here a method which is not yet in a botorch release. I would like to merge it into BoFire as soon as the botorch release is there and then also file a BoFire release. Can you have a look again to this PR?

jduerholt:
@bertiqwerty: Ok, the relevant tests are now passing.

bofire/strategies/predictives/sobo.py
```diff
@@ -91,7 +96,9 @@ def _get_acqfs(self, n) -> List[AcquisitionFunction]:
             X_observed=X_train,
             X_pending=X_pending,
             constraints=constraint_callables,
-            mc_samples=self.acquisition_function.n_mc_samples,
+            mc_samples=self.acquisition_function.n_mc_samples
+            if hasattr(self.acquisition_function, "n_mc_samples")
```
bertiqwerty: This looks inconsistent. Below, you we do `isinstance`, here we do `hasattr`. In general I don't really like both approaches. `hasattr` needs the attribute's name as string making it vulnerable to refactoring. With `isinstance` you would need to keep track of all the acquisition functions that have or don't have the attribute. I would propose duck typing.
```python
def _get_parameter_values(acqf):
    try:
        n_mc_samples = acqf.n_mc_samples
    except AttributeError:
        n_mc_samples = 512
    try:
        beta = acqf.beta
    except AttributeError:
        beta = 0.2
    try:
        tau= acqf.tau
    except AttributeError:
        tau = 1e-3
    try:
        prune = acqf.prune_baseline
    except AttributeError:
        prune = True

    return n_mc_samples, beta, tau, prune

def _get_acqfs():
    ...
    n_mc_samples, beta, tau, prune = _get_parameter_values(self.acqusition_function)
    ...


```

bertiqwerty (approved):
Thanks Johannes! Please check my comment before merging.

bofire/strategies/predictives/sobo.py
jduerholt: Hi @bertiqwerty,

good catch, this is indeed not nice. I think, I even found amore elegant ways as the one that you are proposing. Just have a look to my latest commit.

Best,

Johannes

bertiqwerty: Now it's again all strings making it vulnerable to refactoring.

jduerholt: hmm, for me it is as vulnerable as the try except solution, also there if you rename one of the attributes, you will get an attribute error and the default will be set. Where is the difference?

jduerholt: In principle this unifies the current implementation, no `hasattr` and no `isinstance` and it is much cleaner code using already existing infrastructure.

bertiqwerty: If you rename `self.acquisition_function.n_mc_samples`, it will be renamed. Not the local variable but the `acqf.n_mc_samples`. That was the whole point. Ok, you have to add a type hint to the function so that it is picked up. Or don't do a separate function. About the cleaner code, I disagree. In my opinion less code is not always cleaner. But I admit that clean code is often subjective. Your proposal is basically similar to using `hasattr` everywhere with the additional effort of serializing via model-dump.

bertiqwerty: Ah. I think I got what you mean. I meant refactor the code in the IDE. But you meant the code won't crash when the variable has been renamed. Anyway, not hill I will die on.

jduerholt: Also no hill for me :D Thanks for the review. I would propose to leave it now as is and revisit it under the next refactoring ;)

jduerholt:
We need to run later again, new botorch release is not yet at pypi.

PR #756, opened 2026-04-24
Refactor Any Unions

### LINKED ISSUE #755: [FEATURE REQUEST]: Use Field(discriminator="type") on data-model Unions
### Motivation

Across `bofire/data_models/**/api.py`, type unions are declared as plain `Union[...]` without a Pydantic discriminator. Examples: `AnyFeature`, `AnyInput`, `AnyOutput` in `bofire/data_models/features/api.py`; analogous unions for constraints, kernels, priors, strategies, surrogates.

Without a discriminator:

- **Validation is O(N)** in the number of branches — Pydantic tries each one on every deserialization.
- **Error messages are noisy** — a single malformed spec emits one error per branch (18+ for `AnyFeature`), burying the actually-relevant message.
- **Branch order becomes load-bearing** in smart-mode disambiguation. The TODO at `bofire/data_models/features/api.py:39` (`CategoricalOutput has to be the first item here, no idea why ...`) is a direct symptom of this: some specs are ambiguous enough that Pydantic's branch ordering decides which class is picked.
- **Emitted JSON Schema is weaker** — `anyOf` without a `discriminator` hint. Tooling that consumes the schema (OpenAPI generators, pydantic-ai tool synthesis, OpenAI strict structured-output mode) has to infer discrimination heuristically or fall back to less precise constraints.

Every branch in these unions already declares `type: Literal["<ClassName>"] = "<ClassName>"`, so the codebase is one annotation away from a properly tagged union.

Wrap every relevant union in `Annotated[Union[...], Field(discriminator="type")]`.

Example (`bofire/data_models/features/api.py`):

```python
from typing import Annotated, Union
from pydantic import Field

AnyInput = Annotated[
    Union[
        DiscreteInput,
        CategoricalInput,
        ContinuousInput,
        ContinuousDescriptorInput,
        CategoricalDescriptorInput,
        CategoricalMolecularInput,
        TaskInput,
        ContinuousMolecularInput,
    ],
    Field(discriminator="type"),
]
```

Apply the same change to:

- `AnyFeature`, `AnyInput`, `AnyOutput`, `AnyEngineeredFeature` (features)
- `AnyConstraint` and variants (constraints)
- `AnyKernel` (kernels)
- `AnyPrior`, `AnyPriorConstraint` (priors)
- `AnyStrategy` (strategies)
- `AnyBotorchSurrogate` / `AnySurrogate` (surrogates)
- `AnyObjective`, `AnyCategoricalObjective` (objectives)

Update the `register_*` helpers in `bofire/data_models/**/_register.py` so that registration rebuilds the `Annotated[Union[...], Field(discriminator="type")]` wrapper rather than just extending a bare `Union`.

Expected benefits:

1. Removes the `# TODO: here is a bug...` in `features/api.py:39` (ordering becomes irrelevant).
2. O(1) dispatch on `type` instead of O(N) branch trials.
3. Error messages narrow to the intended branch, which is what both humans and LLM-driven `ModelRetry` flows want.
4. Emitted JSON Schema gains explicit `oneOf` + `discriminator` metadata, which is consumed correctly by OpenAPI tooling and by LLM tool-calling APIs (relevant for the LLM strategy landed in #749 and future agentic tooling).
5. `AbstractFeature` on `features/api.py:32` stays as-is — it mixes abstract and concrete classes and is not a serialization target.

Gotchas to audit during migration:

- Every branch's `type` field must be a `Literal` with a unique value. This is already the convention but worth a one-time check.
- No branch should declare `type` as `Optional[Literal[...]]` or with a non-string default — that breaks discriminator dispatch.
- Tests pinning the shape of emitted JSON Schema will need updates (`anyOf` → `oneOf` + `discriminator`).

- **Status quo.** Accept the ordering sensitivity and noisy errors. Cost: the TODO stays, and every union consumer (including agent tooling) has to work around it.
- **Custom root validators to short-circuit union validation manually.** Reinvents what `Field(discriminator=...)` already provides; extra code to maintain.
- **Switch to `TypeAdapter`-based dispatch at deserialization boundaries only.** Fixes validation speed but doesn't improve emitted JSON Schema, and leaves the inline union behavior unchanged for objects constructed in Python.

- The TODO at `bofire/data_models/features/api.py:39` is the concrete symptom this issue resolves.
- Context from PR #749 (LLM Strategy): the LLM tool-calling path consumes emitted JSON Schema, so clean discriminator metadata on unions directly improves agentic usage.

### Pull Request

No (design-tracking issue; happy to scope the work separately).

jduerholt:
## Motivation

This PR addresses and closes https://github.com/experimental-design/bofire/issues/755. 

In short, it makes the error messages much cleaner when deserializing from json, which will improve agentic appications as the agent has an easier error message to reason over. 

Yes.

No.

## Test Plan

Unit tests.

### DIFF
--- a/bofire/data_models/molfeatures/molfeatures.py
+++ b/bofire/data_models/molfeatures/molfeatures.py
@@ -1,6 +1,6 @@
 import warnings
 from abc import abstractmethod
-from typing import Annotated, List, Literal, Optional, Union
+from typing import Annotated, Any, List, Literal, Optional, Union
 
 import pandas as pd
 from pydantic import Field, PrivateAttr, field_validator
@@ -17,7 +17,7 @@ from bofire.utils.cheminformatics import (  # smiles2bag_of_characters,
 class MolFeatures(BaseModel):
     """Base class for all molecular features"""
 
-    type: str
+    type: Any
     filter_descriptors: bool = True
     correlation_cutoff: float = 0.95
     _descriptors: Optional[Annotated[List[str], Field(min_length=1)]] = PrivateAttr(
--- a/bofire/data_models/_register_utils.py
+++ b/bofire/data_models/_register_utils.py
@@ -6,13 +6,7 @@ from typing import Annotated, Optional, Union
 
 from pydantic import Field
 
-
-def _unwrap_annotated(tp):
-    """Return (inner, metadata). If *tp* is not Annotated, metadata is ()."""
-    if typing.get_origin(tp) is Annotated:
-        args = typing.get_args(tp)
-        return args[0], args[1:]
-    return tp, ()
+from bofire.data_models.unions import unwrap_annotated
 
 
 def _discriminator_name(metadata):
@@ -76,7 +70,7 @@ def append_to_union_field(model_cls: type, field_name: str, new_type: type) -> N
     if origin in (list, Sequence):
         # Sequence[Union[...]] or Sequence[Annotated[Union[...], Field(...)]]
         inner = typing.get_args(old)[0]
-        inner_unwrapped, inner_meta = _unwrap_annotated(inner)
+        inner_unwrapped, inner_meta = unwrap_annotated(inner)
         discriminator = _discriminator_name(inner_meta)
         inner_args = typing.get_args(inner_unwrapped)
         if new_type not in inner_args:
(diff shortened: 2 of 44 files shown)


bofire/data_models/molfeatures/molfeatures.py
Jimbo994, line 1: why does this change from str to Any?

jduerholt:
@Jimbo994: this is ready to go.

Jimbo994:
I left one comment about something I do not fully understand. Most of it makes sense and looks good.

bofire/data_models/molfeatures/molfeatures.py
jduerholt: Because, when we have the acutal non-abstract class we overwrite type with `Literal`, and overwriting `str` with `Literal` throws type checker errors. Here we fix it.

PR #757, opened 2026-04-28
Scale RandomStrategy NChooseK sampling and add allow_zero support

jduerholt:
## Motivation

Move ideas/implementations regarding random sampling from https://github.com/experimental-design/bofire/pull/693 into a seperate PR. This should massively speed up everything where random sampling and NChooseK constraints are involved. 

Yes.

Not yet.

## Test Plan

Unit tests.


bofire/data_models/domain/domain.py
```diff
@@ -236,6 +238,76 @@ def get_nchoosek_combinations(self, exhaustive: bool = False):
 
         return used_features_list_final, unused_features_list
 
+    def sample_valid_nchoosek_features(
+        self,
+        rng: random.Random,
```
jduerholt: Do the sampling not via giving an rng object but via a seed as done here: https://github.com/experimental-design/bofire/blob/f9ffad5ef20765ffbe9c11c25cf9545d82d475aa/bofire/data_models/domain/features.py#L382

bofire/strategies/random.py
```diff
@@ -124,32 +126,38 @@ def _sample_with_nchooseks(
...
+            assert isinstance(constraint, NChooseKConstraint)
+            nchoosek_feature_keys.update(constraint.features)
+        allow_zero_feature_keys = {
+            feat.key
+            for feat in self.domain.inputs.get(ContinuousInput)
+            if isinstance(feat, ContinuousInput) and feat.allow_zero
+        }
+        zeroable_keys = nchoosek_feature_keys | allow_zero_feature_keys
+
+        if zeroable_keys:
```
jduerholt, line 158: Every strategy object has a random number generator, from this we can draw an int and pass as seed to the sampling of the nchoosek constraints.

bofire/strategies/random.py
```diff
@@ -124,32 +126,38 @@ def _sample_with_nchooseks(
...
+            rng = random.Random(self._get_seed())
+            n_combos = min(self.max_combinations, candidate_count)
+            drawn = self.domain.sample_valid_nchoosek_features(rng, n=n_combos)
+            combinations: Dict[tuple, int] = {}
+            for combo in drawn:
```
jduerholt: Add some comments here what we are doing.

bofire/strategies/random.py
```diff
@@ -124,32 +126,38 @@ def _sample_with_nchooseks(
...
                 # create new domain without the nchoosekconstraints
                 domain = deepcopy(self.domain)
                 domain.constraints = domain.constraints.get(excludes=NChooseKConstraint)
-                # fix the unused features
-                for key in u:
+                # fix the unselected zeroable features
+                for key in zeroable_keys - set(combo):
                     feat = domain.inputs.get_by_key(key=key)
                     assert isinstance(feat, ContinuousInput)
+                    if feat.allow_zero:
```
jduerholt: I think this is not necesary anmore, or, this was due to a validator that was changed recently.

tests/bofire/data_models/domain/test_domain_nchoosek_combinatorics.py
```diff
@@ -247,3 +247,102 @@ def test_nchoosek_combinations_nonexhaustive():
...
+    """Domain without NChooseK and without allow_zero features yields ()."""
+    import random
+
+    inputs = [ContinuousInput(key="x", bounds=(0, 1))]
+    domain = Domain(inputs=Inputs(features=inputs))
+    samples = domain.sample_valid_nchoosek_features(random.Random(3), n=4)
+    assert samples == [(), (), (), ()]
+
+
+def test_sample_valid_nchoosek_features_default_n_is_one():
```
jduerholt: Test can be removed.

tests/bofire/data_models/domain/test_domain_nchoosek_combinatorics.py
jduerholt, line 1: move this into test_domain.py

tests/bofire/strategies/test_random.py
```diff
@@ -239,6 +239,73 @@ def test_nchoosek():
...
+    assert len(samples) == 50
+    # At most 2 features should be non-zero per sample (from NChooseK)
+    nonzero_counts = (samples[["if1", "if2", "if3"]] != 0.0).sum(axis=1)
+    assert (nonzero_counts >= 1).all()
+    assert (nonzero_counts <= 2).all()
+    # if4 (allow_zero, not in NChooseK) should have some zeros
+    assert (samples["if4"] == 0.0).any()
+
+
+def test_nchoosek_scalability():
```
jduerholt: Remove this test.

jduerholt (commented):
First review for claude.

jduerholt:
@TobyBoyne: can you review this PR?

TobyBoyne:
> @TobyBoyne: can you review this PR?

Sure thing, I'll review it now :)

bofire/data_models/constraints/nchoosek.py
```diff
@@ -41,6 +41,10 @@ def validate_inputs(self, inputs: Inputs):
             assert isinstance(
                 feature_, ContinuousInput
             ), f"Feature {f} is not a ContinuousInput."
+            if feature_.bounds[0] < 0:
+                raise ValueError(
+                    f"Feature {f} must have a lower bound of >=0, but has {feature_.bounds[0]}",
+                )
```
TobyBoyne: Should this condition instead be `if feature_.bounds[0] == 0 or (feature_.bounds[0] > 0 and feature_.allow_zero)`? If `allow_zero==False`, then the lower bound must be 0 I think?

Also, do you need to add a new `Spec` to test this?

bofire/data_models/domain/domain.py
```diff
@@ -236,6 +238,78 @@ def get_nchoosek_combinations(self, exhaustive: bool = False):
 
         return used_features_list_final, unused_features_list
 
+    def sample_valid_nchoosek_features(
```
TobyBoyne: This method feels a bit involved to be a part of `Domain`. Generally when sampling with constraints, a user must define a `RandomStrategy`. Why is that not the case for `NChooseKConstraint`s?

bofire/data_models/domain/domain.py
```diff
@@ -236,6 +238,78 @@ def get_nchoosek_combinations(self, exhaustive: bool = False):
...
+        groups: list[Tuple[list[str], list[int], list[int]]] = []
+        nchoosek_keys: set[str] = set()
+        nchoosek_cons = list(self.constraints.get(NChooseKConstraint))
+        for con in nchoosek_cons:
+            assert isinstance(con, NChooseKConstraint)
+            ks = list(range(con.min_count, con.max_count + 1))
+            if con.none_also_valid and 0 not in ks:
+                ks.insert(0, 0)
+            weights = [math.comb(len(con.features), k) for k in ks]
+            groups.append((list(con.features), ks, weights))
```
TobyBoyne: (Nit) `con.features` is already a `list`, no?

bofire/data_models/domain/domain.py
```diff
@@ -236,6 +238,78 @@ def get_nchoosek_combinations(self, exhaustive: bool = False):
...
+                    or con.min_count <= len(active & set(con.features)) <= con.max_count
+                    for con in nchoosek_cons
+                ):
+                    results.append(tuple(sorted(active)))
+                    break
+            else:
+                raise ValueError(
+                    f"Failed to sample a valid NChooseK combination after "
+                    f"{max_iters} attempts.",
+                )
```
TobyBoyne: Is it worth defaulting to a fallback value, instead of raising an error here? I worry that, for large `n`, all it takes is for *one* inner loop to fail generating a point for the whole function to fail, which feels undesirable. Although computing a default is easier said than done (I think this is a satisfiability problem, which is NP-hard!)

bofire/data_models/domain/domain.py
```diff
@@ -236,6 +238,78 @@ def get_nchoosek_combinations(self, exhaustive: bool = False):
...
+            ks = list(range(con.min_count, con.max_count + 1))
+            if con.none_also_valid and 0 not in ks:
+                ks.insert(0, 0)
+            weights = [math.comb(len(con.features), k) for k in ks]
+            groups.append((list(con.features), ks, weights))
+            nchoosek_keys.update(con.features)
+        for feat in self.inputs.get(ContinuousInput):
+            assert isinstance(feat, ContinuousInput)
+            if feat.allow_zero and feat.key not in nchoosek_keys:
+                groups.append(([feat.key], [0, 1], [1, 1]))
```
TobyBoyne: I think this is fine. But I don't know if it's desirable to have equal weights for zero and non-zero values. Much more of the search space is taken up by non-zero values of a given input, so a better initialization scheme might be to put more weight to exploring these values?

I would consider making the default weights `[1, 3]`. But maybe users would prefer avoiding such magic constants. Your call - just something to think about!

bofire/strategies/random.py
```diff
@@ -124,42 +125,55 @@ def _sample_with_nchooseks(
...
+        if zeroable_keys:
+            # Draw a uniform sample of valid active-feature subsets (one per
+            # NChooseK constraint plus one per allow_zero singleton). We draw
+            # at most `max_combinations` distinct subsets; their multiplicities
+            # in `drawn` determine how many polytope samples each subset gets.
+            n_combos = min(self.max_combinations, candidate_count)
+            drawn = self.domain.sample_valid_nchoosek_features(
+                seed=self._get_seed(),
+                n=n_combos,
+            )
```
TobyBoyne: Pass in `self.max_iters` here

tests/bofire/strategies/test_random.py
```diff
@@ -239,6 +239,47 @@ def test_nchoosek():
     assert len(samples) == 50
 
 
+def test_allow_zero_without_nchoosek():
+    """Test random sampling with allow_zero features but no NChooseK constraint."""
+    if1 = ContinuousInput(bounds=(0.1, 1), key="if1", allow_zero=True)
+    if2 = ContinuousInput(bounds=(0.1, 1), key="if2", allow_zero=True)
+    if3 = ContinuousInput(bounds=(0.1, 1), key="if3")
+    domain = Domain.from_lists(inputs=[if1, if2, if3])
```
TobyBoyne, line 247: (Nit) why do you use `Domain.from_lists` here, but not in `test_domain.py`?

TobyBoyne (approved):
All looks good! Only very minor comments, I think this looks good as-is!

bofire/data_models/domain/domain.py
jduerholt: Yeah, you are right, it was for legacy reasons due to this old enumeration function for nchoosek, I will move it to random strategy and also remove the old legacy enumeration function which is only used and is a big mess.

bofire/data_models/domain/domain.py
jduerholt: Yep, I think so ;) I will fix it

bofire/data_models/domain/domain.py
jduerholt: I would keep it for now. Also I do not have really seen a lot of use cases of overlapping nchoosek constraints. Furhtermore, we have per sample 1000 trials ...

bofire/data_models/domain/domain.py
jduerholt: Yeah, nice idea, but I would keep it for now. Or we expose it as a parameter in the random strategy ...

tests/bofire/strategies/test_random.py
jduerholt: Claude did this :D

bofire/data_models/constraints/nchoosek.py
jduerholt: Yeah, we should implement it as you suggest here. And should also test it with `add_invalid` in the data model tests.

PR #759, opened 2026-05-11
Data Model Migrator

jduerholt:
## Motivation

We introduce quite often breaking changes on data models, which is problematic for updating BoFire versions in production systems (we have such an issue right now). Out of urgent need, I created a migrator setup for data models (purly coded by claude on the basis of the examples from our production system). 

My question is: do we want to maintain something like this in BoFire or not? If yes, we could register here also future breaking changes, but we need to properly review this ;) ... If we decide to not have it here, we keep it out and I just use it for our current need as a script. 

What is your opinion? @bertiqwerty @LukasHebing @TobyBoyne 

Yes.

Not yet.

## Test Plan

Unit tests.

### DIFF
--- a/bofire/data_models/migration/MIGRATION.md
+++ b/bofire/data_models/migration/MIGRATION.md
@@ -1,0 +1,153 @@
+# BoFire data-model migration
+
+A batch tool that converts legacy BoFire payloads (strategy / surrogate / domain
+JSON dicts) to the current schema.
+
+## Scope
+
+- **Offline / batch only.** No runtime adapter, no validation hooks. Callers
+  read from their store, pass dicts through `migrate(...)`, and write back.
+- **Pure dict -> dict.** Never calls `model_validate` internally. The output is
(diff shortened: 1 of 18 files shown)


TobyBoyne:
No strong thoughts from me. This seems a bit too heavy given BoFire is still in v0 - maybe this would be good for a v0 -> v1 change but seems like it could be more work than it's worth for every release? But this is much more relevant on the production side so I will leave it to the industrial people to decide :)

bertiqwerty:
Thanks Johannes. I think we should not keep this as part of core BoFire. We can create a separate thing under the experimental-design-group. But in general I think we should start to work more carefully with breaking changes and use deprectations in BoFire. I haven't put too much thought into this. Maybe something to discuss at the hackathon.

KislayaRavi:
@jduerholt I also need this too. I have write a `futurize` function which updates the data model. But, it is a patchy solution. 
I agree with @bertiqwerty suggestion of keeping it outside the core BoFire.

PR #764, opened 2026-05-20
Added a new function for validating strictly increasing objective bounds instead of monotone increasting

### LINKED ISSUE #690: [FEATURE REQUEST]: drop single-point bounds on domain objectives, because they lead to unintelligible errors downstream
### Motivation

Downstream, sometimes have had a user set bounds on an objective. If the bounds do not have identical left and righthand sides, then they are generally ignored by ipopt. 

ParEGO/botorch errors out in searching when the bounds reduce the space to have nothing in it.
This will happen due to human error if the bounds for an objective get set to constant, e.g.

`bounds=(0.0, 0.0)`

Error: 

`RuntimeError, probability tensor contains either `inf`, `nan` or element < 0'`

Would like to check for this and in the case of identical left and right,  just drop the bounds (and let the default (0,1) then get assigned) instead.  

[Obviously, will not fix all problems about search spaces being empty, but will catch some problems.]

Minimally, check the domain for inconsistencies like this and then raise a verbose error. 

_No response_

### Pull Request

None

rosonaeldred:
## Motivation
This PR tightens objective bounds validation and aligns tests with the intended behavior.

Previously, objective bounds accepted non-strict bounds. For IdentityObjective-based objectives, that can create degenerate normalization ranges and lead to unstable behavior. This change enforces strict bounds for objectives so lower must be strictly less than upper.

To avoid unintended side effects, monotonic validation in shared type utilities remains available for other non-objective uses.

While running the related tests, I also found a deprecation warning for FactorialStrategy during serialization specs and updated the serialization test to explicitly expect that warning.

## Have you read the Contributing Guidelines on pull requests?
Yes.

## Have you updated CHANGELOG.md?
Yes. Added entries for:

- strict objective bounds validation behavior
- serialization test handling for FactorialStrategy deprecation warning
- testing that domains reject bad objectives

### Test Plan
Added strict validator test coverage in test_types.py:

rejects decreasing
rejects equal adjacent values
accepts strictly increasing sequences

Updated objective invalid specs in objectives.py:

MaximizeObjective with equal bounds is invalid
MinimizeObjective with equal bounds is invalid

expected error matching aligned with [identity.py] validator
Updated serialization test in test_serialization.py:

explicitly expects DeprecationWarning for FactorialStrategy

Added regression coverage for Domain.model_validate_json(...) with [bad_objective_bounds.json]

#### Ran:

[test_types.py]
objective-invalid slice in [test_invalid.py]
strategy serialization path in [test_serialization.py]
full data-model suite under data_models

Outcome:

strict objective bounds behavior validated
non-objective monotone behavior preserved
full data-model suite passed


bofire/data_models/types.py
```diff
@@ -73,6 +73,25 @@ def validate_monotonically_increasing(sequence: Sequence[float]) -> Sequence[flo
...
+        sequence: Sequence of values.
+
+    Raises:
+        ValueError: If lower bound is greater or equal to the upper bound.
+
+    Returns:
+        Validated sequence
+
+    """
+    if len(sequence) > 1:
```
bertiqwerty, line 89: I think you don't need the check but please double check.

bertiqwerty (approved):
Thanks @rosonaeldred, looks good to me. I left just one comment. Feel free to merge as soon as all the tests pass.

PR #763, opened 2026-05-20
fix(surrogates): noise_prior and noise_constraint on SingleTaskGPSurr…

R-M-Lee:
…ogate now correctly influence GP fitting

attempt to understand and fix https://github.com/experimental-design/bofire/issues/762

PR #766, opened 2026-05-20
Hotfix/noiseprior

jduerholt:
## Motivation

This continues the issue with the noise prior problem from https://github.com/experimental-design/bofire/pull/763.

Yes.

No.

## Test Plan

Unit tests.


R-M-Lee:
Looks good to me. There is a failing test though, with

```
FAILED tests/bofire/surrogates/test_gps.py::test_noise_prior_affects_single_task_gp - AssertionError: Large-noise prior should yield higher fitted noise: large=0.002720 vs default=0.005658
assert 0.0027199727785463222 > 0.0056578318500057085
```

we should understand what is going on here. Maybe the test is not checking the right thing or the particular example is not good

jduerholt:
Ok, @R-M-Lee, this should work now. The test was a bit flaky. Can you approve, then I will merge when the tests have passed?

R-M-Lee:
Very good. A minor thing I want to mention:

The noise_constraint field changing from Optional (defaulting to None) to a required GreaterThan(lower_bound=1e-4) default, and Interval.initial_value becoming optional, are both minor API/serialization breaks that should probably be called out in the changelog.

R-M-Lee (approved):

jduerholt:
> Very good. A minor thing I want to mention:
> 
> The noise_constraint field changing from Optional (defaulting to None) to a required GreaterThan(lower_bound=1e-4) default, and Interval.initial_value becoming optional, are both minor API/serialization breaks that should probably be called out in the changelog.

Good catch, I changed it so that also the old behavior (providing `None`) will be still working!

Issue #657, opened 2025-11-07
[FEATURE REQUEST]: Aggregate duplicates before cross validation


jduerholt:
### Motivation

Currently cross validation is applied on all experiments passed to the `cross_validate` method, if there are duplicates in the dataset, data leakage can lead to positively distorted model metrics. We should have a keyword argument in `cross_validate` if we want to aggreate before running the cross validation. Aggregation can be done by `domain.aggregate_by_duplicates`. I would see this aggregation as default behavior.

### Describe the solution you'd like to see implemented in BoFire.

See above.

### Pull Request

None

FrankWanger:
Working on this, wanted to confirm:

1. If `experiments` passed to `cross_validate` has an optional `group_split_column` column, it would be ignored by `aggregate_by_duplicates` method and disappears in the returned `experiments` because this method processes input/outputs.

Since `aggregate_by_duplicates` does not return original index of removed rows after aggregation, it would be tricky to handle this extra `group_split_column` in `cross_validate`. And it would not make sense to average the group label. I am thinking of the potential solutions to:

A. add `aux_columns` argument in the `aggregate_by_duplicates` to handle this special case. By default the first occurrence of this value would be retained when duplication found. So the expected behaviour becomes:

```python
  experiments = pd.DataFrame({
        "x": [1.0, 1.0, 3.0, 4.0],
        "y": [1.0, 2.0, 3.0, 4.0],
        "group": [0, 1, 2, 3],
    })
```
->
```python
  experiments_aggregated = pd.DataFrame({
        "x": [1.0, 3.0, 4.0],
        "y": [1.5, 3.0, 4.0],
        "group": [0, 2, 3],
    })
```

B. return the raw index in the `aggregate_by_duplicates` and handle `group_split_column` inside `cross_validate`. Although this seems to be less favourable for backward compatibility.

2. default precision for the aggregation?

Happy to discuss.

jduerholt:
In case somebody provides a group_split_column, we should ignore the parameter for the aggregation and not apply any aggregation. This is the easiest one, the group split stuff is also an edge case. As precision, I would take 3.

jduerholt:
Closed by https://github.com/experimental-design/bofire/pull/769

PR #769, opened 2026-05-22
Aggregate duplicates before cross validation

FrankWanger:
## Motivation

The Feature Request has been specified and discussed in #657 

Currently cross validation is applied on all experiments passed to the cross_validate method, if there are duplicates in the dataset, data leakage can lead to positively distorted model metrics. 

Yes

Yes

### Added a test case `test_model_cross_validate_aggregate` to verify if samples are correctly aggregated:

The input contains 5 samples with two duplicated inputs.
```python
   experiments = pd.DataFrame({
        "x_1": [1.0, 1.0, 2.0, 2.0, 3.0],
        "x_2": [1.0, 1.0, 2.0, 2.0, 3.0],
        "y": [1.0, 2.0, 3.0, 4.0, 5.0],
    })
```

This test verifies if the number of samples becomes 3 after aggregation. And also tested compatibility to see if number of samples is 5 when `aggregate` is set to `False`.

### Modified a test case `test_model_cross_validate_groupfold` to add validation of aggregate is correctly switched off when group fold is supplied:

```python
    # Test if aggregate fallback to False when group split is supplied
    with pytest.warns(
        UserWarning,
        match="Aggregation is not compatible with group split, fallback to no aggregation.",
    ):
        train_cv, test_cv, hook_results = model.cross_validate(
            experiments,
            folds=4,
            aggregate=True,
            random_state=random_state,
            group_split_column="group",
        )
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ and this project adheres to [Pragmatic Versioning](https://github.com/experiment
 - Configurable `noise_constraint` support for GP-based surrogates (`SingleTaskGP`, `MixedSingleTaskGP`, `TanimotoGP`, `MultiTaskGP`, and `RobustSingleTaskGP`) and corresponding linear/polynomial wrappers.
 - Optional `initial_value` field on the `GreaterThan`, `LessThan`, and `Positive` prior constraint data models (already present on `Interval`), letting users opt-in to a warm-start of the constrained gpytorch parameter at construction time.
 - Generalized NChooseK constraint support in DoE: `min_count > 0` is now supported, non-zero lower bounds (`lb > 0`) are allowed for NChooseK features, overlapping NChooseK constraints (shared features) are handled via incremental pairwise merge with consistency filtering, and `nchoosek_constraints_as_bounds` generates deactivation patterns for all activity levels `k ∈ [min_count, max_count]`.
-- Aggregation of duplicated experiments in the `cross_validate` method of trainable surrogates to avoid data leakage, controlled via the `aggregate` boolean flag, default `True`.
+- Aggregation of duplicated experiments in the `cross_validate` method of trainable surrogates to avoid data leakage, controlled via the `aggregate` boolean flag, default `False`.
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


Jimbo994:
Hi Frank, thanks for the contribution. I started the tests.

Perhaps it is better to set `aggregate: bool = False`, instead of the current True. Not sure how many people rely on the default settings, and this could potentially be an impactful change if they don't know about it. The user setting it to True if they require it still benefit from the new feature.

FrankWanger:
> Hi Frank, thanks for the contribution. I started the tests.
> 
> Perhaps it is better to set `aggregate: bool = False`, instead of the current True. Not sure how many people rely on the default settings, and this could potentially be an impactful change if they don't know about it. The user setting it to True if they require it still benefit from the new feature.

Thanks for the advice - I've switched to False by default. :)

Jimbo994:
Hi Frank, nearly looks good. You still have some Linting errors, likely due to whitespace or something.

Try and run:
pip install pre-commit          # if you don't have it
pre-commit run --all-files      # this auto-fixes the files
git commit -am "Fix lint: apply pre-commit formatting."
git push

I can't do that for you because I cannot push to your branch.

FrankWanger:
> Hi Frank, nearly looks good. You still have some Linting errors, likely due to whitespace or something.
> 
> Try and run: pip install pre-commit # if you don't have it pre-commit run --all-files # this auto-fixes the files git commit -am "Fix lint: apply pre-commit formatting." git push
> 
> I can't do that for you because I cannot push to your branch.

Amazing! Just got it done. Thanks for the tips :)

Hopefully this fix everything.

jduerholt:
Hi @Jimbo994  and @FrankWanger,

I will have a look today. @FrankWanger : Thank you very much already!

Best,

Johannes

Jimbo994:
Thanks @jduerholt, before I hand it over to you, a few more comments @FrankWanger.

The tests now actually don't test if the aggregate_prec works because you use exact matches, perhaps a good idea to also add an example where there are some differences. I think it is better to add this to the tests in `test_domain.py `related to `domain.aggregate_by_duplicates`

@jduerholt I think we also don't handle CategoricalOutputs correctly in `domain.aggregate_by_duplicates`! We should address this in a separate PR.

jduerholt (approved):
Looks good to me. Thanks!

PR #770, opened 2026-05-26
add handling of CategoricalOutput in aggregrate_by_duplicates

Jimbo994:
## Motivation

PR #657 introduces the option to aggregate duplicates in a dataset before doing cross validation, here duplicates are based on input similarity. For this the pre-existing function `domain.aggregate_by_duplicates` is used, but this function was not handling `CategoricalOutput`s and these were dropped from the experiments dataframe. As a result cross validation could throw errors because of missing `CategoricalOutput`s when fitting. This PR addresses this by extending the `domain.aggregate_by_duplicates` to also aggregrate `CategoricalOutput`s, here it was chosen to aggregrate based on majority vote, or randomly when there is a tie. In both cases the user is warned that this is happening, as perhaps the user would not want to aggregrate at all if experiments that are duplicates in the inputs have different CategoricalOutputs.

Yes

Yes

## Test Plan

Added test case for majority votes with a clear winner.
Added test case for a tie in the majority votes triggering a random pick
Added additional test case for ContiniousInputs with slight variations to test previously untested precision parameter.

### DIFF
--- a/bofire/data_models/domain/domain.py
+++ b/bofire/data_models/domain/domain.py
@@ -189,10 +189,9 @@ class Domain(BaseModel):
         features. Continuous input features are rounded before identifying the
         duplicates. Continuous output features are aggregated by taking the
         average (or median) of the involved values; categorical output
-        features are aggregated by majority vote, with a warning emitted when
-        the duplicates carry conflicting categorical labels. Ties in the
-        majority vote are broken by a random pick (a separate warning is
-        emitted; pass ``random_state`` to make this reproducible).
+        features are aggregated by majority vote. Ties in the majority vote
+        are broken by a random pick; pass ``random_state`` to make this
+        reproducible.
 
         Args:
             experiments (pd.DataFrame): Dataframe containing experimental data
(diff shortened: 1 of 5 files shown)


Jimbo994:
@jduerholt ready for review.

bofire/data_models/domain/domain.py
```diff
@@ -227,6 +238,38 @@ def aggregate_by_duplicates(
         agg: Dict[str, Any] = dict.fromkeys(
             self.outputs.get_keys(ContinuousOutput), method
         )
+
+        rng = np.random.default_rng(random_state)
```
jduerholt: Can you rename `random_state` to `seed`? We use `seed` everywhere in BoFire.

jduerholt (changes requested):
Looks good to me. I would just recommend to remove the warnings, this will spoil the output. It is clear for a user when he calls the method that there can be contradicting labels. And he has always the possiblity to inspect them as the duplicated labcodes are also returned, so it is the same as for the continuous ones, there we also do not return any warnings.

bofire/data_models/domain/domain.py
Jimbo994: renamed

Jimbo994:
> Looks good to me. I would just recommend to remove the warnings, this will spoil the output. It is clear for a user when he calls the method that there can be contradicting labels. And he has always the possiblity to inspect them as the duplicated labcodes are also returned, so it is the same as for the continuous ones, there we also do not return any warnings.

warnings are removed.

jduerholt (approved):
Looks good. Thx!

PR #771, opened 2026-05-26
TabPFN Surrogate

jduerholt:
## Motivation

This PR adds a TabPFN surrogate and tests it for Hartmann optimzation, I was only running one repeat. But I see that it is really worse in comparison to just a default GP. 

@KislayaRavi : Can you have a look and maybe run the same thing with your implementation (have a look at the messy notebook in this PR) I would be really curious to see if you see the same. 

Yes.

Not yet.

## Test Plan

Unit tests


KislayaRavi:
Hi Johannes,
I will take a deeper look into your PR tomorrow. Meanwhile, I ran Hartmann 6-dimensional optimization problem on Vanilla GP and TabICL using my code. The acquition function is qLogNEI. (I played around too much with TabPFN code and now its broken, I will fix it when I get time, but not sure when). Here are my results from two different seeds:
image
image
image
image

So, in my opinion it is not a correct to say that foundation model is really worse than vanilla GP.  I have tested it on many other benchmarks like Rastrigin, Ackley, Bohachevsky, Droplet etc and drawing a straightforward superiority of one over the other is not right. On the contrary, foundation models outperform vanilla GP in many cases. 

The underlying idea of TabPFN is not that far away from TabICL. So, the difference should not be that much. But I cannot vouch for the similarity.

Some of my observations/conclusions :
- Foundation models are not well-suited for simple fitting tasks. For example, attempting to fit a quadratic relationship can reveal limitations, particularly when analyzing gradients.
- They are better applied when relationships are complex or when categorical variables are involved, as tabular foundation models are specifically built for structured data.
- Foundation models can sometimes exhibit overconfidence, leading to insufficient exploration. Exploring alternative acquisition functions—such as UCB or similar methods—could help encourage better exploration.
- A key benefit of foundation models is their out-of-the-box usability, as they do not require hyperparameter tuning or configuration decisions.
- They should not be treated as a universal solution. No single surrogate model works best for all scenarios; foundation models are most valuable in cases where Gaussian Processes are difficult to apply.
- Handling high-dimensional problems remains challenging for foundation models, although approaches like GIT-BO may help address this limitation.

jduerholt:
@KislayaRavi 

thanks for the explanation, I think, both of us had a bad seed. I was now running 45 repeats with Hartmann and got this: 

image


Which exactly reprodices the paper: https://arxiv.org/pdf/2402.02229

I will also implement tabicl in the same spirit as in this PR, then we should run large scale benchmarks with appropriate repeats to see if we can reproduce also the community versions. 

What do you think?

KislayaRavi:
I think that is a good idea. My proposal is to plot data profile plot (https://www.mcs.anl.gov/~wild/dfo/benchmarking/ceperf.pdf) to benchmark the algorithms. But without GPUs, foundation models will be difficult to run. I am planning to run the benchmarks using BoTorch and ditch BoFire all together because it does not support GPUs. Currently I am a bit overwhelmed, so I cannot commit on timeline. But I will update on this post

jduerholt:
> I think that is a good idea. My proposal is to plot data profile plot (https://www.mcs.anl.gov/~wild/dfo/benchmarking/ceperf.pdf) to benchmark the algorithms. But without GPUs, foundation models will be difficult to run. I am planning to run the benchmarks using BoTorch and ditch BoFire all together because it does not support GPUs. Currently I am a bit overwhelmed, so I cannot commit on timeline. But I will update on this post

No worries, no hurry, same for me. My plan would be to put the GPU support into BoFire ;) I will work on it.

PR #774, opened 2026-06-12
Add dimensionality-scaled Gamma priors (CHEN and dim-scaled threesix)

jduerholt:
## Context

BayBE 0.15.0 introduced dimension-aware Gamma hyperpriors in its GP presets — the **CHEN** preset and a new default that scales the lengthscale prior with the problem dimensionality. BoFire already expresses the Hvarfner dimension-scaled prior as a first-class, serializable data model (`DimensionalityScaledLogNormalPrior`) plus the mapper plumbing that passes the problem dimension `d` into lengthscale-prior mapping. This PR adds the Gamma analogues, reusing that same infrastructure — so these priors are serializable prior objects rather than imperative factory code.

## What

- **`DimensionalityScaledGammaPrior`** — Gamma analogue of `DimensionalityScaledLogNormalPrior`:
  ```
  concentration_eff = concentration + concentration_scaling * sqrt(d)
  rate_eff          = rate * d ** rate_power
  ```
  The asymmetric scaling (additive on concentration, power on rate) lets a single class express both target priors exactly.

- **Opt-in prior constants** (no surrogate/kernel defaults change):
  - `CHEN_LENGTHSCALE_PRIOR` = `Gamma(2m, 2)`, `CHEN_OUTPUTSCALE_PRIOR` = `Gamma(m, 1)` with `m = 0.4·√d + 4` — dimension-aware hyperpriors from Chen, Fleck & Stuyver, *"Leveraging Hidden-Space Representations Effectively in Bayesian Optimization for Experiment Design through Dimension-Aware Hyperpriors"*, ChemRxiv 2026, [doi:10.26434/chemrxiv.10001986/v2](https://doi.org/10.26434/chemrxiv.10001986/v2).
  - `DIMENSIONALITY_SCALED_THREESIX_LENGTHSCALE_PRIOR` — keeps the threesix concentration of 3 and scales the rate by `d**-0.5`, calibrated so the lengthscale mode matches the Hvarfner log-normal prior.

- **ScaleKernel plumbing**: `map_ScaleKernel` now forwards `d` (= number of active dims) to the outputscale prior mapping (mirroring `map_PolynomialFeatureInteractionKernel`), and `ScaleKernel.outputscale_prior` is widened from `AnyGeneralPrior` to `AnyPrior` so dimension-scaled priors are accepted there (registered in `priors._rebuild_dependent_models`).

## Testing

- Data-model layer (serialization roundtrip + spec validity) — `tests/bofire/data_models` passes
- Functional mapper + ScaleKernel `d`-threading — `tests/bofire/priors/test_mapper.py`, `tests/bofire/kernels/test_mapper.py`
- Registration coverage (`test_prior_rebuild_covers_all_anyprior_fields`) passes with the new `AnyPrior` field
- `ruff check`/`format` clean; `ty check bofire` reports no new errors

### DIFF
--- a/bofire/data_models/priors/api.py
+++ b/bofire/data_models/priors/api.py
@@ -52,11 +52,6 @@ _PRIOR_CONSTRAINT_TYPES: list[type] = [
 AnyPriorConstraint = tagged_union(*_PRIOR_CONSTRAINT_TYPES)
 
 
-# Deprecated alias kept for backwards compatibility. All prior map sites now receive the
-# problem dimensionality `d` (priors that do not need it ignore it), so there is no longer a
-# need for a separate union excluding the dimensionality-scaled priors. Use AnyPrior instead.
-AnyGeneralPrior = AnyPrior
-
 # default priors of interest
 # botorch defaults
 THREESIX_LENGTHSCALE_PRIOR = partial(GammaPrior, concentration=3.0, rate=6.0)
--- a/bofire/data_models/priors/api.py
+++ b/bofire/data_models/priors/api.py
@@ -52,11 +52,10 @@ _PRIOR_CONSTRAINT_TYPES: list[type] = [
 AnyPriorConstraint = tagged_union(*_PRIOR_CONSTRAINT_TYPES)
 
 
-# these are priors that are generally applicable
-# and do not depend on problem specific extra parameters
-AnyGeneralPrior = tagged_union(
-    GammaPrior, NormalPrior, LKJPrior, LogNormalPrior, SmoothedBoxPrior
-)
+# Deprecated alias kept for backwards compatibility. All prior map sites now receive the
+# problem dimensionality `d` (priors that do not need it ignore it), so there is no longer a
+# need for a separate union excluding the dimensionality-scaled priors. Use AnyPrior instead.
+AnyGeneralPrior = AnyPrior
 
 # default priors of interest
 # botorch defaults
--- a/bofire/data_models/priors/api.py
+++ b/bofire/data_models/priors/api.py
@@ -1,3 +1,4 @@
+import math
 from functools import partial
 
 from bofire.data_models.priors._register import (
@@ -10,7 +11,7 @@ from bofire.data_models.priors.constraint import (
     Positive,
     PriorConstraint,
 )
-from bofire.data_models.priors.gamma import GammaPrior
+from bofire.data_models.priors.gamma import DimensionalityScaledGammaPrior, GammaPrior
 from bofire.data_models.priors.interval import (
     Interval,
     LogTransformedInterval,
@@ -29,6 +30,7 @@ from bofire.data_models.unions import tagged_union
 
 _PRIOR_TYPES: list[type[Prior]] = [
     GammaPrior,
+    DimensionalityScaledGammaPrior,
     NormalPrior,
     LKJPrior,
     LogNormalPrior,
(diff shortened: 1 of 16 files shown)


tests/bofire/surrogates/test_gps.py
```diff
@@ -181,6 +182,22 @@ def test_SingleTaskGPModel(kernel, scaler, output_scaler):
     assert_frame_equal(preds, preds2)
 
 
+def test_SingleTaskGPModel_with_dimensionality_scaled_noise_prior():
```
Jimbo994: Is this test really necessary, we don't do this for any of the other priors?

bofire/data_models/priors/api.py
```diff
@@ -50,11 +52,10 @@
...
 
-# these are priors that are generally applicable
-# and do not depend on problem specific extra parameters
-AnyGeneralPrior = tagged_union(
-    GammaPrior, NormalPrior, LKJPrior, LogNormalPrior, SmoothedBoxPrior
-)
+# Deprecated alias kept for backwards compatibility. All prior map sites now receive the
+# problem dimensionality `d` (priors that do not need it ignore it), so there is no longer a
+# need for a separate union excluding the dimensionality-scaled priors. Use AnyPrior instead.
+AnyGeneralPrior = AnyPrior
```
Jimbo994: Can we not get rid of AnyGeneralPrior entirely then?

bofire/kernels/mapper.py
```diff
@@ -240,7 +240,10 @@ def map_ScaleKernel(
             **kwargs,
         ),
         outputscale_prior=(
-            priors.map(data_model.outputscale_prior)
+            priors.map(
+                data_model.outputscale_prior,
+                **({"d": len(active_dims)} if active_dims is not None else {}),
```
Jimbo994: Is this not a problem? You would pass {} if no active_dims, this would then cause problems in map_DimensionalityScaledGammaPrior() as d is required there.

Jimbo994 (changes requested):
Placed some comments, and I think there is a potential issue with the current PR that needs to be addressed.

bofire/kernels/mapper.py
Jimbo994: This would expose it:

import gpytorch
import torch
from bofire.data_models.kernels.api import ScaleKernel, RBFKernel
from bofire.data_models.priors.api import (
    DimensionalityScaledGammaPrior,
    GammaPrior,
)
import bofire.kernels.api as kernels


def test_scalekernel_dimscaled_outputscale_prior_without_active_dims():
    """ScaleKernel must thread d to a dimensionality-scaled outputscale prior
    even when active_dims is None. Currently it silently drops d and the
    scaled-prior mapper raises a bare TypeError."""
    data_model = ScaleKernel(
        base_kernel=RBFKernel(ard=False),
        outputscale_prior=DimensionalityScaledGammaPrior(
            concentration=3.0,
            concentration_scaling=0.0,
            rate=2.0,
            rate_power=-0.5,  # non-zero so d actually matters
        ),
    )
    mapped = kernels.map(
        data_model,
        batch_shape=torch.Size(),
        active_dims=None,  # the path the surrogate never takes
        features_to_idx_mapper=None,
    )
    assert isinstance(mapped.outputscale_prior, gpytorch.priors.GammaPrior)

bofire/data_models/priors/api.py
jduerholt: Correct, we should get completely rid of it.

tests/bofire/priors/test_mapper.py
```diff
@@ -94,3 +98,46 @@ def test_DimensionalityScaledLogNormalPrior_map(
...
+@pytest.mark.parametrize("d", [1, 5, 20])
+def test_chen_and_threesix_constants_map(d):
+    # CHEN: lengthscale Gamma(2m, 2), outputscale Gamma(m, 1), m = 0.4*sqrt(d) + 4
+    m = 0.4 * math.sqrt(d) + 4
+    ls = priors.map(CHEN_LENGTHSCALE_PRIOR(), d=d)
+    assert ls.concentration == pytest.approx(2 * m)
+    assert ls.rate == pytest.approx(2.0)
+    os = priors.map(CHEN_OUTPUTSCALE_PRIOR(), d=d)
+    assert os.concentration == pytest.approx(m)
+    assert os.rate == pytest.approx(1.0)
```
jduerholt: Why everywehre this pytest.approx? Not needed, or?

tests/bofire/surrogates/test_gps.py
jduerholt: Yep, we should remove it.

jduerholt:
@Jimbo994: I update the PR, can you have a look again?

tests/bofire/data_models/specs/priors.py
Jimbo994, line 1: why do we need random.random() and can we not just set something deterministically to test?

Jimbo994 (approved):
I approve, left one minor comment why we need stochasticity in the datamodels/specs/priors.py. Just wondering why.

Other than that, looks good!

tests/bofire/data_models/specs/priors.py
jduerholt: Somehow convention in these data model tests, at least for a few of them.

PR #781, opened 2026-06-23
Fix/doe nchoosek ipopt bounds

MrzvskK:
## Motivation

`test_ask[domain5-16]` in `test_space_filling.py` fails intermittently (~1 in 3 runs) with `ConstraintNotFulfilledError`. The root cause: NChooseK inactive dimensions are encoded as `[0, 0]` variable bounds, but IPOPT (and SLSQP) return O(1e-4) residuals on those pinned dimensions. With `max_count=1`, one leaked inactive feature plus the real active feature yields `sums=2`, which fails `NChooseKConstraint.is_fulfilled(tol=1e-4)` in `Strategy.ask`.

The fix adds `_project_to_equality_bounds()` in `utils.py`, which snaps `x[i] = lb` wherever `lb == ub` after the solver returns. This covers NChooseK pins, fixed experiments, and partially-fixed experiments — any dimension that is semantically an equality constraint on the solver variable.

Yes

yes

## Test Plan

Run the previously flaky test repeatedly to confirm it no longer fails


MrzvskK:
@jduerholt also something noticed during the nonlinear constraints PR review. I might be wrong on this.

MrzvskK:
@dlinzner-bcs was told to tag you.

MrzvskK:
@dlinzner-bcs quick reminder here

MrzvskK:
@jduerholt can you take a look please?

PR #786, opened 2026-06-26
implement convergence criteria scaffold and two simple examples

bertiqwerty:
## Motivation

Scaffold for convergence criteria. A strategy has converged if it is not expected provide better results in future requests. Even though it might overstate the math. term convergence I picked it over stopping criterion or termination criterion, because the latter can also just the experimental budget has been reached. But these convergence criteria are part of the strategy and have access to the internal state to assess the progress of the optimization. The orchestration layer, on the other hand, with, e.g., the number-of-experiments-condition or now the new has-converged-condition can stay in the step-wise strategy.

Y

y

## Test Plan

unit tests

### DIFF
--- a/bofire/data_models/strategies/convergence_criteria/convergence_criterion.py
+++ b/bofire/data_models/strategies/convergence_criteria/convergence_criterion.py
@@ -27,9 +27,10 @@ class ConvergenceCriterion(BaseModel):
       ``StrategyHasConvergedCondition`` bridge.
 
     The data model only holds the parameters; the actual convergence logic is
-    implemented in the functional layer (see ``bofire.convergence_criteria``).
+    implemented in the functional layer (see
+    ``bofire.strategies.convergence_criteria``).
     Custom criteria can be added via
-    :func:`bofire.convergence_criteria.api.register`.
+    :func:`bofire.strategies.convergence_criteria.api.register`.
     """
 
     type: Any
--- a/bofire/data_models/strategies/convergence_criteria/convergence_criterion.py
+++ b/bofire/data_models/strategies/convergence_criteria/convergence_criterion.py
@@ -12,14 +12,20 @@ class ConvergenceCriterion(BaseModel):
     (e.g. a maximum number of experiments or steps) is intentionally not covered
     here, as it is already handled by the conditions of the ``StepwiseStrategy``.
 
+    Convergence criteria only apply to ``PredictiveStrategy``s: they are an
+    intrinsic property of a single model-based strategy and are consumed by the
+    strategy itself via ``strategy.has_converged()``. The evaluator therefore has
+    full access to the strategy's internals, including its fitted surrogate
+    model(s), which it reads directly from the strategy when needed.
+
     Distinction from a stepwise ``Condition`` (see
     ``bofire.data_models.strategies.stepwise.conditions``):
 
     - A convergence criterion is an **intrinsic property of a single strategy**.
       It is consumed by the strategy itself via ``strategy.has_converged()`` and
-      therefore has full access to the strategy's internals, including its fitted
-      surrogate model(s) (see ``requires_surrogate``). This is the right home for
-      model-aware stopping logic that can assess the progress of the optimization.
+      therefore has full access to the strategy's internals. This is the right
+      home for model-aware stopping logic that can assess the progress of the
+      optimization.
     - A stepwise ``Condition`` is **orchestration** between strategies: it selects
       the active step of a ``StepwiseStrategy`` and only sees the strategy through
       the minimal ``StepStrategy`` protocol (i.e. ``has_converged()`` alone, no
@@ -34,15 +40,3 @@ class ConvergenceCriterion(BaseModel):
     """
 
     type: Any
-
-    @property
-    def requires_surrogate(self) -> bool:
-        """Whether the criterion needs the strategy's surrogate model(s).
-
-        Some convergence criteria are evaluated purely on the observed data
-        (e.g. the improvement of the best objective), while others rely on the
-        surrogate model(s) of the strategy (e.g. the acquisition value). The
-        functional layer uses this flag to provide the surrogates and to raise a
-        clear error if a required model is not available.
-        """
-        return False
--- a/bofire/data_models/strategies/convergence_criteria/convergence_criterion.py
+++ b/bofire/data_models/strategies/convergence_criteria/convergence_criterion.py
@@ -1,3 +1,4 @@
+from abc import abstractmethod
 from typing import Any
 
 from bofire.data_models.base import BaseModel
@@ -40,3 +41,38 @@ class ConvergenceCriterion(BaseModel):
     """
 
     type: Any
+
+    @classmethod
+    @abstractmethod
+    def is_applicable_to_multiobjective(cls) -> bool:
+        """Abstract method to check if the convergence criterion is multi-objective
+
+        Returns:
+            bool: True if the convergence criterion is multi-objective, False otherwise
+
+        """
+
+    @classmethod
+    @abstractmethod
+    def is_applicable_to_singleobjective(cls) -> bool:
+        """Abstract method to check if the convergence criterion is single-objective
+
+        Returns:
+            bool: True if the convergence criterion is multi-objective, False otherwise
+
+        """
+
+    @classmethod
+    @abstractmethod
+    def is_applicable_to_objective_free(cls) -> bool:
+        """Abstract method to check if the criterion needs no objective information.
+
+        This is True for criteria that assess convergence without relying on any
+        objective values (e.g. purely from the proposed input locations), so they
+        can be used by objective-free strategies such as active learning.
+
+        Returns:
+            bool: True if the convergence criterion can be evaluated without any
+            objective, False otherwise
+
+        """
--- a/bofire/data_models/strategies/stepwise/conditions.py
+++ b/bofire/data_models/strategies/stepwise/conditions.py
@@ -1,5 +1,5 @@
 from abc import abstractmethod
-from typing import Annotated, Any, List, Literal, Optional
+from typing import Annotated, Any, List, Literal, Optional, Protocol
 
 import pandas as pd
 from pydantic import Field, PositiveInt, field_validator
@@ -10,9 +10,27 @@ from bofire.data_models.objectives.api import ConstrainedObjective
 from bofire.data_models.unions import tagged_union
 
 
+class StepStrategy(Protocol):
+    """Minimal interface of the step strategy needed to evaluate conditions."""
+
+    def has_converged(self) -> bool: ...
+
+
 class EvaluateableCondition:
     @abstractmethod
-    def evaluate(self, domain: Domain, experiments: Optional[pd.DataFrame]) -> bool:
+    def evaluate(
+        self,
+        strategy: StepStrategy,
+        domain: Domain,
+        experiments: Optional[pd.DataFrame],
+    ) -> bool:
+        """Whether the step using ``strategy`` should remain the active step.
+
+        This is the polymorphic hook used by the ``StepwiseStrategy`` to select
+        the active step. Data-based conditions only look at ``domain`` and
+        ``experiments`` and ignore ``strategy``; conditions that depend on the
+        strategy itself (e.g. whether it is finished) inspect ``strategy``.
+        """
         pass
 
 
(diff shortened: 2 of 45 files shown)


bofire/strategies/convergence_criteria/mapper.py
jduerholt, line 1: As the convergence criteria are strategy specif datamodels, maybe we should move it under strategies? We should do this in a refactor maybe also for other stuff like priors and kernels ...

bofire/strategies/convergence_criteria/objective_improvement.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+if TYPE_CHECKING:
+    from bofire.strategies.strategy import Strategy
+    from bofire.surrogates.botorch_surrogates import BotorchSurrogates
+
+
+def evaluate_objective_improvement_criterion(
+    criterion: ObjectiveImprovementCriterion,
+    strategy: "Strategy",
```
jduerholt: should we not always assume that it is a surrogates based BO strategy, then strategy would be sufficient, or? Bacause strategy holds the surrogates object.

bofire/data_models/strategies/stepwise/conditions.py
```diff
@@ -10,9 +10,43 @@
 from bofire.data_models.unions import tagged_union
 
 
+class StepStrategy(Protocol):
```
jduerholt, line 13: Why is this needed?

bofire/strategies/stepwise/stepwise.py
```diff
@@ -45,10 +45,10 @@ def has_sufficient_experiments(self) -> bool:
     def get_step(self) -> Tuple[Strategy, Optional[Transform]]:
         """Returns the strategy at the current step and the corresponding transform if given."""
         for i, condition in enumerate(self.conditions):
-            if condition.evaluate(self.domain, experiments=self.experiments):
-                return self.strategies[i], self.transforms[
-                    i
-                ]  # ty: ignore[invalid-return-type]
+            if condition.evaluate(
```
jduerholt, line 48: Hmm, why this change?

jduerholt: because the condition fulfillment is now strategy specific?

bofire/strategies/strategy.py
```diff
@@ -124,6 +125,52 @@ def tell(
     def _tell(self) -> None:
         """Method to allow for customized tell functions in addition to self.tell()"""
 
+    def has_converged(self) -> bool:
```
jduerholt: I think this overly complex or, the converged conditions makes only sense for predictive strategies or? There we have everything and do need all the majic, or?

bofire/data_models/strategies/strategy.py
```diff
@@ -13,6 +14,7 @@ class Strategy(BaseModel):
     type: Any
     domain: Domain
     seed: Optional[Annotated[int, Field(ge=0)]] = None
+    convergence_criterion: Optional[AnyConvergenceCriterion] = None
```
jduerholt: Not sure if i would to base it on strategy, maybe predicitveStrategy, and then we can have something like `is_feature_implemented` to check if the criterion that we use here fits to the strategy to validate, what do you think?

tests/bofire/data_models/specs/strategies.py
jduerholt, line 1: Ideally we would also the conditions to the serialization testing with its own registry, so that it is always tested that they properly serialize and deserialize, on the other hand they are strategy specific things and not standalone objects, so maybe standakone tests are not necessary.

jduerholt (commented):
Hi Behrang,

thanks. This goes overall in exactly the direction in which I thought.

I let some comments. Let us discuss them.

Best,

Johannes

bofire/strategies/convergence_criteria/mapper.py
bertiqwerty: I also thought about this and decided to handle them like priors and kernels. We have to be careful when we refactor because the existing ones would get invalidated import paths, I think.

bofire/strategies/convergence_criteria/objective_improvement.py
bertiqwerty: Good point.

bofire/data_models/strategies/stepwise/conditions.py
bertiqwerty: Good point.

bofire/data_models/strategies/strategy.py
bertiqwerty: There are criteria that don't need surrogate models. On the other hand, we do not have "converging" algorithms in BoFire that do not have surrogate models, yet.

bofire/strategies/stepwise/stepwise.py
bertiqwerty: Yes. In case you want the `StrategyHasConvergedCondition`.

bofire/strategies/strategy.py
bertiqwerty: I tend to agree with you. As I said above, one could have converging strategies without surrogate models. But maybe this would not be BO anymore but just some kind of black-box optimization.

tests/bofire/data_models/specs/strategies.py
bertiqwerty: Agreed.

bofire/data_models/strategies/stepwise/conditions.py
bertiqwerty: Ah. If we want to rely on the existence of a `has_converged` in the strategy-base-class, we need to move the condition evaluation out of the data models part into a functional component part. Would be more effort but consistent. Of course, this is only necessary for the static type checker and autocomplete. During runtime it doesn't make a difference.

bofire/strategies/convergence_criteria/mapper.py
jduerholt: Question is if we should not already have the future picture in mind here?

bofire/data_models/strategies/strategy.py
jduerholt: I mean somethings that clearly only work for single objective strategies should not be allowed for multibojective ones?

bofire/strategies/convergence_criteria/mapper.py
bertiqwerty: Yes. Ok.

bofire/strategies/predictives/predictive.py
```diff
@@ -38,6 +43,24 @@ def is_fitted(self) -> bool:
...
+        criterion is configured, the strategy is never considered converged.
+
+        Returns:
+            bool: True if the strategy's convergence criterion is met, False otherwise.
+
+        """
+        criterion = self._data_model.convergence_criterion
+        if criterion is None:
+            return False
+        evaluator = convergence_criteria.map(criterion)
```
bertiqwerty, line 61: I am wondering if this is the correct place for the mapping or if this should happen in the constructor. @jduerholt, any opinion?

bertiqwerty: It has to happen during construction, such that everything is validated during construction, right? (I am aware that the validator is still missing)

bertiqwerty: ok, mapping there is fine. there is no validation happening during mapping.

bertiqwerty:
@jduerholt, please have another look.

bofire/data_models/strategies/convergence_criteria/convergence_criterion.py
```diff
@@ -0,0 +1,78 @@
...
+    implemented in the functional layer (see
+    ``bofire.strategies.convergence_criteria``).
+    Custom criteria can be added via
+    :func:`bofire.strategies.convergence_criteria.api.register`.
+    """
+
+    type: Any
+
+    @classmethod
+    @abstractmethod
```
jduerholt, line 46: Ah, you did it like this, in principle the current philosophy is that always the class who owns something checks the compatibility.

bofire/strategies/convergence_criteria/hypervolume_improvement.py
```diff
@@ -0,0 +1,92 @@
...
+        output_feature_keys=outputs.get_keys(),
+    )
+    n = len(valid)
+    # Need at least one experiment before the lookback window to form a baseline.
+    if n <= criterion.n_lookback:
+        return False
+
+    # A common reference point inferred from all valid experiments makes the
+    # hypervolumes of the two windows directly comparable.
+    ref_point = infer_ref_point(domain, valid)
```
jduerholt, line 83: We could take here the reference point from the strategy class if available.

bofire/strategies/convergence_criteria/objective_improvement.py
```diff
@@ -0,0 +1,70 @@
...
+    desirabilities = strategy.domain.outputs(experiments)
+    rewards = desirabilities.sum(axis=1).to_numpy(dtype=float)
+    rewards = rewards[~np.isnan(rewards)]
+
+    n = rewards.shape[0]
+    # Need at least one experiment before the lookback window to form a baseline.
+    if n <= criterion.n_lookback:
+        return False
+
+    best_now = float(np.max(rewards))
```
jduerholt, line 67: This is very much the same as for the hypervolume thingy, we could think about introducing another abstraction layer that definies the actual combined objective, which would be hypervolume for mobo and so forth,

bofire/strategies/convergence_criteria/objective_improvement.py
```diff
@@ -0,0 +1,70 @@
...
+        over the last ``n_lookback`` experiments, False otherwise (including when
+        there are not yet more than ``n_lookback`` experiments).
+    """
+    experiments = strategy.experiments
+    if experiments is None:
+        return False
+
+    # Map each experiment to a scalar reward via the domain objectives
+    # (higher is better) and sum the per-output desirabilities.
+    desirabilities = strategy.domain.outputs(experiments)
```
jduerholt, line 58: Not 100% correct here, as this is somehow now just doing additive scalarization. I think we need for the different sobo classes and possibilities to scalarizes there (Addtive Sobo and so forth) a new abstraction.

bofire/strategies/convergence_criteria/proposal_deviation.py
```diff
@@ -0,0 +1,80 @@
...
+    continuous_inputs = strategy.domain.inputs.get(ContinuousInput)
+    if len(continuous_inputs) == 0:
+        return False
+
+    n = len(experiments)
+    # Need n_consecutive + 1 proposals to form n_consecutive deviations.
+    if n < criterion.n_consecutive + 1:
+        return False
+
+    # Min-max normalize the continuous input locations to the unit cube.
```
jduerholt: What happens in case of categoricals?

tests/bofire/strategies/test_convergence.py
```diff
@@ -0,0 +1,413 @@
...
+    experiments["valid_y1"] = 1
+    experiments["valid_y2"] = 1
+    strategy = strategies.map(
+        MoboStrategy(domain=domain, convergence_criterion=criterion)
+    )
+    strategy.tell(experiments, retrain=False)
+    return strategy
+
+
+def test_convergence_criterion_serialization_roundtrip():
```
jduerholt: This can be removed, it is automatically tested via the specs machinerey as you added it to strategies and is tested there, if you want to test it as standalone, we should generate its own "tests/bofire/data_models/specs/convergence_criteriay.py" module register it properly and use this. This here is not the crrect location.

tests/bofire/strategies/test_make_strategy.py
```diff
@@ -13,7 +13,22 @@
 def remove_optional(anno):
     origin = get_origin(anno)
     if origin == typing.Union or origin is types.UnionType:
-        return sorted((a for a in anno.__args__ if a is not type(None)), key=hash)
+        result = []
```
jduerholt, line 16: Why is this needed?

jduerholt (approved):
Hi Behrang,

thank you very much, see my inline comments, most are minor ones that should not be adressed in this PR. I think we need to refactor the strategies (especially the SOBO) strategies a bit so that the convergence critieria can easily see how the objectives needs to combined.

Only thing that should change is the serialization test, this is the wrong location.

Best,

Johannes

tests/bofire/strategies/test_make_strategy.py
bertiqwerty: `ConvergenceCriterion` is the first thing that is both, a tagged union (i.e., `Annotated`) and optional.

tests/bofire/strategies/test_convergence.py
bertiqwerty: Right!

bofire/data_models/strategies/convergence_criteria/convergence_criterion.py
bertiqwerty: This is still the case. But the class needs to know, if the convergence criterion is applicable. This is what this is for.

bofire/strategies/convergence_criteria/hypervolume_improvement.py
bertiqwerty: I propose to improve the actual criteria after the merge. The main reason I added them was to "test" the scaffold.

bofire/strategies/convergence_criteria/proposal_deviation.py
bertiqwerty: Good point. In case of a purely categorical domain -> always false. In case of a mixed domain, convergence ignores categoricals.

bertiqwerty: Categoricals and discretes are now part of the check.

jduerholt:
Looks good to me, merge whenever you want!

PR #789, opened 2026-06-30
Unify categorical feature encodings (data on feature, encoding on surrogate)

jduerholt:
> **Scope note.** This started as "the categorical descriptor path only, continuous/discrete as a follow-up PR." It has since grown to cover **all three feature kinds** and to fully objectify the encoding subsystem — the original follow-up roadmap (items 1–3) now lands here. Only the optional "one unified `transforms` pipeline" step (old item 4) is deliberately left for later. The sections below describe the branch as it now stands, not the original narrower plan.

## The bigger picture

BoFire modelled the same conceptual operation — *"derive numeric model columns from a structured input"* — with a class explosion, because three orthogonal axes were fused into class identity:

| Axis | Values |
|---|---|
| **A. Feature kind** | categorical / continuous / discrete |
| **B. Descriptor source** | none / static table / generated-from-structure (SMILES) / composite |
| **C. Reduction op** | select-row (categorical) / amount-weighted-sum (continuous mixtures) |

Each `(A,B)` pair was its own class (`CategoricalDescriptorInput`, `CategoricalMolecularInput`, `ContinuousDescriptorInput`, `ContinuousMolecularInput`, …), so some combinations were simply impossible — e.g. **handcrafted *and* molecular descriptors on one feature**, or **descriptors on a discrete input** — and `to_/from_descriptor_encoding` was duplicated across feature classes.

The redesign separates **data from choice**:

- **Data, defined once → on the feature.** A `descriptors` table (numeric-only columns) plus an explicit `structure` field (SMILES), shared by all feature kinds via a `DescriptorsMixin`.
- **Encoding choice, per-surrogate → encoder objects.** A `DescriptorEncoding` references the feature's data and carries the modelling knobs (which `columns`, which `generators`, `filter_descriptors`, `correlation_cutoff`).
- **Runtime unchanged** — the existing `ChainedInputTransform` (`NumericToCategoricalEncoding`, `AppendFeatures`, scaler).

The litmus test for where something belongs: *would I retype it if I added a second surrogate?* If yes → data, on the feature. If it's a knob I'd tune per surrogate → choice, on the surrogate.

### Data on the feature — one shape for all kinds
- **`DescriptorsMixin`** mixed into `CategoricalInput`, `ContinuousInput`, `DiscreteInput`: a numeric-only `descriptors: Dict[str, List[float]]` table plus an explicit `structure: Optional[List[str]]` (SMILES, rdkit-lazy validated). Structures are a **schema-visible field**, not a magic `"smiles"` key in a dict disambiguated by a name registry — a typo like `reactant_smiles` can no longer silently become a broken numeric column.
- `descriptor_levels()` is single-component by default (`[self.key]`); `CategoricalInput` overrides with one row per category. This fixes `DiscreteInput` (a single numeric quantity, like `ContinuousInput` — not a set of per-value rows) and removes a duplicate override.

### Encoding subsystem fully objectified
- **New `bofire/data_models/encodings/`**. `CategoricalEncodingEnum` is **removed**; every encoding is a pydantic object under one base with a uniform `encode` / `decode` / `get_names` / `get_bounds` interface:
  - `OneHotEncoding(drop_first: bool = False)` — **DUMMY folded in** as `drop_first=True`
  - `OrdinalEncoding` — the pre-model passthrough that `NumericToCategoricalEncoding` must not expand (matched via `isinstance`)
  - `DescriptorEncoding` — carries a shared **`DescriptorSpec`** (`columns` + `generators: List[AnyDescriptorGenerator]` + `filter_descriptors` + `correlation_cutoff`). Static + generated columns are assembled into one block; **composite = both populated**, and correlation decorrelation is a *pure* function over the whole assembled block (no mutable state on generators).
- The `if spec == ONE_HOT … elif isinstance(spec, …)` chains in `domain.features` (`transform` / `inverse_transform` / `_get_transform_info` / `get_bounds`) and `torch_tools.get_categorical_encoder` collapse into **branch-free polymorphic calls**. `CategoricalInput` exposes **one** generic pandas entry point — `to_encoding` / `from_encoding` — replacing the six old per-encoding helpers.

### `MolFeatures` → `DescriptorGenerator`
- `bofire/data_models/molfeatures/` → `bofire/data_models/descriptor_generators/`; base `MolFeatures` → `DescriptorGenerator`, union `AnyMolFeatures` → `AnyDescriptorGenerator` (concrete `Fingerprints`/`Fragments`/`MordredDescriptors` names unchanged). They're now **pure, stateless transforms** — dropped `filter_descriptors`/`correlation_cutoff`/`_descriptors` state (lifted to `DescriptorSpec`); `CompositeMolFeatures` and the `FingerprintsFragments` factory are gone (fp+frag is just `generators=[Fingerprints(), Fragments()]`). Old import path kept as a warning shim; no serialization impact (type discriminators unchanged).

### Surrogate side
- **`categorical_encodings`** holds the encoder objects (1-arity, select-row), auto-defaulted from the feature's descriptor **data**: a `structure` column → fingerprint generator, numeric descriptor columns → static `DescriptorEncoding`, otherwise the surrogate's non-descriptor fallback (one-hot / ordinal); task inputs never descriptor-encode. Surrogates now only override `_default_plain_categorical_encodings`; descriptor resolution is shared in the base.
- **`engineered_features`** holds `WeightedSumFeature(columns/generators, normalize)` (N-arity, amount-weighted blend `Σ amountᵢ · rowᵢ` for mixtures/formulations) — the four `Weighted*`/`Molecular*` engineered features collapse into it, reading the shared descriptor data. **Two-layer split kept on purpose**; `scaler`/`output_scaler` untouched.
- **`input_preprocessing_specs` removed as a stored field.** It was a pre-in-model-transform vestige that every botorch surrogate force-set to ordinal — 100% derivable from `inputs`. It's now a read-only `@property` returning `{cat: OrdinalEncoding()}`; `categorical_encodings` is the single categorical knob on a surrogate. Legacy dumps/kwargs still load (dropped with a `DeprecationWarning`).

### New combinations this unlocks
```python
solvent = CategoricalInput(
    key="solvent",
    categories=["water", "ethanol", "thf"],
    descriptors={"logP": [-1.4, -0.3, 0.5], "MW": [18, 46, 72]},  # numeric-only
    structure=["O", "CCO", "C1CCOC1"],                            # explicit SMILES field
)
# same feature, different encodings per surrogate — incl. handcrafted + molecular together:
A = SingleTaskGPSurrogate(..., categorical_encodings={"solvent": DescriptorEncoding(columns=["logP"])})
B = SingleTaskGPSurrogate(..., categorical_encodings={
    "solvent": DescriptorEncoding(columns=[], generators=[Fingerprints(n_bits=1024)]),
})
```

## Compatibility & behaviour notes

- **Back-compat.** Old feature classes (`CategoricalDescriptorInput`, `CategoricalMolecularInput`, `ContinuousDescriptorInput`, `ContinuousMolecularInput`) keep working as deprecation shims that accept the legacy shape and re-emit the new one. Old serialized values migrate with a `DeprecationWarning`: enum strings (`"ONE_HOT"`/`"DUMMY"`/`"ORDINAL"`/`"DESCRIPTOR"`), bare molfeatures, `descriptors={"smiles": …}` → `structure=…`, generator `reads`/filter fields, and legacy `input_preprocessing_specs`. (Decision: accept-old / emit-new + warn.)
- **One behaviour delta.** `Inputs.get_bounds` now runs descriptor generation+filtering during scaler setup (the old bounds path didn't filter). Tests asserting exact unfiltered columns use `filter_descriptors=False` — the faithful new-API equivalent. `filter_descriptors` now defaults to **False**.

## What's deliberately left for later

The optional final consolidation — folding `categorical_encodings` + `engineered_features` + `scaler` into one ordered `transforms` pipeline — is **not** in this PR. The two-layer split (1-arity auto-defaulted encodings vs N-arity engineered features) is a fine resting point.

## Testing

- `tests/bofire/data_models` + `tests/bofire/surrogates` — green (1660 passed); new `encodings` spec collection wired into the serialization/deserialization harness; all categorical specs carry `descriptors`/`structure`; encoding-migration and drop-legacy-`input_preprocessing_specs` tests added.
- `tests/bofire/strategies` — green (478 passed); the only 2 failures (`doe/test_utils.py::test_n_zero_eigvals_*`) are **pre-existing** (fail on the clean base, unrelated to this work).
- `ruff` clean; `ty` introduces no new errors.


bofire/data_models/features/_descriptor_spec.py
```diff
@@ -0,0 +1,179 @@
...
+        duplicates = sorted({n for n in names if names.count(n) > 1})
+        if duplicates:
+            raise ValueError(
+                f"Duplicate descriptor names in descriptor spec: {duplicates}.",
+            )
+        return names
+
+    # -- table assembly (shared core) ----------------------------------------------
+
+    def _generated_frames(
```
jduerholt: Please add docstrings here, especially on the arguments.

bofire/data_models/features/_descriptor_spec.py
```diff
@@ -0,0 +1,179 @@
...
+            frames.append(gen_df)
+        return frames
+
+    def _assemble(
+        self,
+        index: List,
+        static: Optional[pd.DataFrame],
+        structures: Optional[pd.Series],
+    ) -> pd.DataFrame:
+        """Assemble a descriptor table from prepared parts (static ‖ generated).
```
jduerholt: Please improve the docstring, especially on the arguemtns, what is index, stratic and strucutures, and especially in which case, categorical vs continuous, also provide an example.

bofire/data_models/features/_descriptors.py
```diff
@@ -0,0 +1,121 @@
...
+        return
+    for value in values:
+        smiles2mol(value)
+
+
+class DescriptorsMixin(BaseModel):
+    """Mixin giving a feature per-level numeric ``descriptors`` and an optional
+    ``structure`` column.
+
+    A "level" is a category (categorical), a discrete value (discrete), or the single
```
jduerholt: Please also provide examples how it looks for the three different feature types.

bofire/data_models/features/categorical.py
```diff
@@ -34,7 +47,7 @@
 LLM_ENUM_SCHEMA_THRESHOLD = 32
 
 
-class CategoricalInput(Input):
+class CategoricalInput(Input, DescriptorsMixin):
     """Base class for all categorical input features.
 
     Attributes:
```
jduerholt, line 76: We should list here all attributes in addition to the ones inherited from Mixin as this is user facing.

bofire/data_models/features/continuous.py
```diff
@@ -31,6 +32,10 @@ class ContinuousInput(NumericalInput):
     order_id: ClassVar[int] = 1
 
     bounds: Bounds
+
```
jduerholt: Remove this.

bofire/data_models/features/continuous.py
```diff
@@ -6,13 +6,14 @@
...
 
+from bofire.data_models.features._descriptors import DescriptorsMixin
 from bofire.data_models.features.feature import Output, TTransform
 from bofire.data_models.features.numerical import NumericalInput
 from bofire.data_models.objectives.api import AnyObjective, MaximizeObjective
 from bofire.data_models.types import Bounds
 
 
-class ContinuousInput(NumericalInput):
+class ContinuousInput(NumericalInput, DescriptorsMixin):
```
jduerholt: Please also add to the docstring the inherided attributes, as this is user facing.

bofire/data_models/surrogates/botorch.py
```diff
@@ -42,53 +31,62 @@ class BotorchSurrogate(Surrogate):
...
+        source, and a plain categorical uses the surrogate's non-descriptor fallback.
+        """
+        fallbacks = cls._default_plain_categorical_encodings()
+        if not isinstance(feat, CategoricalTaskInput):
+            if feat.structure is not None:
+                # fingerprint-encode from the structure column, static columns excluded
+                return DescriptorEncoding(columns=[], generators=[Fingerprints()])
+            if feat.descriptor_columns():
+                return DescriptorEncoding()  # all numeric descriptor columns
+        for klass in type(feat).__mro__:
```
jduerholt: What is this?

bofire/surrogates/multi_task_gp.py
```diff
@@ -17,7 +17,7 @@
 from bofire.data_models.features.api import CategoricalTaskInput
 from bofire.data_models.priors.api import LKJPrior
 
-# from bofire.data_models.molfeatures.api import MolFeatures
+# from bofire.data_models.descriptor_generators.api import DescriptorGenerator
```
jduerholt: remove this comment

bofire/surrogates/robust_single_task_gp.py
```diff
@@ -4,7 +4,7 @@
 import torch
 from botorch.fit import fit_gpytorch_mll
 
-# from bofire.data_models.molfeatures.api import MolFeatures
+# from bofire.data_models.descriptor_generators.api import DescriptorGenerator
```
jduerholt: remove this comment

bofire/surrogates/single_task_gp.py
```diff
@@ -12,7 +12,7 @@
 import bofire.priors.api as priors
 from bofire.data_models.enum import OutputFilteringEnum
 
-# from bofire.data_models.molfeatures.api import MolFeatures
+# from bofire.data_models.descriptor_generators.api import DescriptorGenerator
```
jduerholt: remove this comment

jduerholt (commented):
I went over it again, mainly doc stuff.

jduerholt:
@bertiqwerty @TobyBoyne @LukasHebing:

I was cooking on this PR now for quite a while. It is now ready to review. Can you have a look? The idea is to tackle our feature task explosion and streamline it. Claude tried to code it in a backward compatible manner. Question is if we really want to have here backward compatibility? For me it would be ok to just go with it, which would also cut the number of lines of this PR massively (around 600). What do you think? 

Also please report if you have better ideas for streamlining and/or dislike the apporach from here ;)

bofire/data_models/surrogates/botorch.py
```diff
@@ -42,53 +31,67 @@ class BotorchSurrogate(Surrogate):
...
+        and ``TaskInput.validate_no_descriptor_data`` already guarantees they carry no
+        descriptor data. The check is kept explicit here so the intent is local.
+        """
+        is_task = isinstance(feat, CategoricalTaskInput)
+        if not is_task:
+            if feat.structure is not None:
+                # fingerprint-encode from the structure column, static columns excluded
+                return DescriptorEncoding(columns=[], generators=[Fingerprints()])
+            if feat.descriptor_columns():
+                return DescriptorEncoding()  # all numeric descriptor columns
```
TobyBoyne: Is this a reasonable default? If we provide the structure, do we automatically want to enable fingerprints?

TobyBoyne:
Wow, big PR! I will start a review now :) Here are a few thoughts I had from a brief glimpse:

I like the direction of this PR a lot! Having reviewed some other PRs where there has been confusion about the role of `EncodingEnum`s and `InputPreprocessingSpecs` etc etc, it is quite needed, and this PR will lead to a cleaner codebase. I wonder if this can go even further, and combine `EngineeredFeatures` and `Descriptors`. They are both deterministic functions of the inputs, so the fact that they are two distinct parts of the code seems like a missed opportunity to me?

I think that maintaining backwards compat is not worth it - either don't make this change, or make the change and don't have backwards compat. The whole point of this refactor is to reduce the number of ways of expressing the same thing, but as it currently stands it only increases! This constraint has also led to a number of code smells (eg. `DescriptorEncoding.validate_for_feature` just calling `self.validate_for`), which has made the first pass through the PR a little tricky to read. EDIT: not actually sure if backwards compat is the reason for the example I give, but the point still stands!

Let me know if you disagree, of course I am not coming at this from the industrial side so things-not-breaking is less important to my work!

I also don't like `_descriptor.py` and `_descriptor_spec.py`... there is something about them that has made the structure of this PR too convoluted for me. Maybe it's the `Mixin`s - I would prefer if this can be compositional, rather than inherited. I'll have a think if I can formalize my issue with them and maybe propose an alternative.

bertiqwerty:
Thanks Johannes. I have looked at it just very briefly. From my point of view, we can ditch backwards compatibility. I would focus on getting it clean without backwards compatibility.

bofire/data_models/features/continuous.py
```diff
@@ -6,15 +6,28 @@
...
+    descriptor level (the feature itself), so each ``descriptors`` column and the
+    ``structure`` column hold exactly one value. This is what lets a continuous
+    feature act as one component of a mixture, blended by a ``WeightedSumFeature``::
+
+        ethanol = ContinuousInput(
+            key="ethanol",
+            bounds=(0, 1),
+            descriptors={"logP": [-0.3], "MW": [46.0]},   # one value per column
+            structure=["CCO"],                            # one SMILES
+        )
```
TobyBoyne, line 31: Does this have any meaning by itself, outside of a `WeightedSumFeature`? If I had a domain with *just* this feature, would it make sense to have, when this input `==0.5`, then `logP==-0.15`?

If it *only* makes sense in the context of an engineered feature, I would move the descriptor logic to the `EngineeredFeature`.

bofire/data_models/encodings/naming.py
```diff
@@ -0,0 +1,13 @@
+"""Naming convention for encoded feature columns.
```
TobyBoyne, line 1: Feels like this should either be a static method on `CategoricalEncoding`, or a function in `utils/naming_conventions.py`

TobyBoyne (commented):
Okay, I have had a look through. There's a fair bit of code so I haven't been able to look at every change, but have left a few comments.

My only remaining major feedback (beyond my first comment) is that it's not clear to me what a `ContinuousFeature` with `descriptors` and `structure` is useful for, besides `EngineeredFeature`s. In which case, I think the descriptor should live with the `EngineeredFeature`.

jduerholt:
Thank you both, very much. I will then first ditch all the backward compatibility shim, which will make it easier to digest. And I agree with you Toby, that the current setup with  `_descriptor.py` and `_descriptor_spec.py` makes it convoluted. I try to find a better solution. 

Regarding the descriptor data in `ContinuousInput`, I disagree. It is the same as for `CategoricalInput`, there it is also only needed by `DescriptorEncoding`. The design pattern is that the domain holds all available data regarding individual features. Or what do you think @TobyBoyne ?

TobyBoyne:
> The design pattern is that the domain holds all available data regarding individual features

Yes, good point. Happy to concede this one! 

> the current setup with _descriptor.py and _descriptor_spec.py makes it convoluted

I do think the usage should look something more like this:
```python
# existing approach - `ContinuousInput` inherits `DescriptorsMixin`
feat = ContinuousInput(key="ethanol", bounds=(0, 1), structures=["CCO"], descriptors={"logP": [-0.3]})

# alternate approach - implementation logic in `Descriptors`
descriptor = Descriptor(
    structures=["CCO"] , # or maybe `{"ethanol": "CCO"}`
    descriptors={"logP": [-0.3]})  # or maybe `{"logP": {"ethanol": -0.3}}`
)  
feat = ContinuousInput(key="ethanol", bounds=(0, 1), descriptor=descriptor)
```

In the current implementation, the only time `DescriptorsMixin` needs access to other attributes is for `descriptor_levels`, for which I think there is likely an easy workaround. So here, I think composition is better (since we only really call-down from `Feature` to `DescriptorsMixing`).

Feel free to ping me if you want a second review!

jduerholt:
@TobyBoyne : What is the advantage of your proposal?

TobyBoyne:
@jduerholt Happy to leave it as is, mainly it's that the code is a bit tricky to read, to be honest. It took me a bit of time to understand the hierarchy, and I think having this deep inheritance and intertwining of responsibility is part of what triggered the need for a refactor in the first place. The more dependencies we have between the different classes, the more likely a future refactor will be! 

Of course, code-readability is perhaps less important in the age of LLMs (I'm not sure anyone is running `help(ContinuousInput)` to see what methods it has). But I think there is a benefit even to the LLMs if the responsibilities of each class are better delineated.

Very much your call - the tests pass, everything seems to work, and your approach has less-nested JSON serialization.

bertiqwerty:
I am with Toby on this one. I think composition is the better approach here. Especially, since we never really make use of inheritance in the object oriented sense with the `DescriptorMixin`, do we? I mean we don't have functions that accept the mixin-base-class, do we? That should work with all derivatives of `DescriptorMixin`? The goal is just to avoid code duplication. You might save a few lines with mixins but the composition-approach is more straightforward and simpler.

jduerholt:
@TobyBoyne and @bertiqwerty: I changed to compose. You are right. It is better! What do you think of the overall architecture now?

jduerholt:
@bertiqwerty  and @TobyBoyne  I tried to further improve readability and clearness of the PR, please have a look again ;)

bofire/data_models/features/_descriptor_spec.py
```diff
@@ -213,21 +156,48 @@ def _assemble(
...
+    def validate_for(self, descriptors: Optional[Descriptors], key: str) -> None:
+        """Ask every compatibility question about ``descriptors`` and this spec.
+
+        This is the gate: it is the only place that copes with a feature carrying no
+        block at all, and the only place that raises for an incompatible one. Everything
+        below it takes a `Descriptors` and assumes these checks have passed — which is
+        why `build` asserts rather than re-checking. It is reached from every entry point
+        that can lead to `build`: `Inputs._get_transform_info` / `transform` /
+        `inverse_transform` / `get_bounds` via `_validate_transform_specs`, and
+        `BotorchSurrogate` at construction.
```
TobyBoyne: (Nit) Do you have any strong feelings about the quantity/length of docstrings in this PR? I generally think they're a little verbose, and often have specific references outside the scope of the method that may become outdated. I also don't think they contribute to understanding of the code. But for the sake of getting this PR merged, this is not a top priority at all.

bofire/data_models/features/descriptors.py
```diff
@@ -1,19 +1,24 @@
-"""Shared per-feature descriptor data (categorical / continuous / discrete).
+"""Per-level descriptor data, as a value object carried by a feature.
```
TobyBoyne, line 1: No longer private - remove leading underscore in file name?

bofire/data_models/features/descriptors.py
```diff
@@ -148,24 +105,94 @@ def _validate_structure(cls, structure):
...
-    def _validate_descriptor_lengths(self):
-        n = len(self.descriptor_levels())
-        for name, column in self.descriptors.items():
-            if len(column) != n:
-                raise ValueError(
-                    f"descriptor column '{name}' must have {n} value(s) (one per level)",
-                )
-        if self.structure is not None and len(self.structure) != n:
-            raise ValueError(f"structure must have {n} value(s) (one per level)")
+    def _validate_rectangular(self):
```
TobyBoyne: (Nit) Pydantic validators aren't private methods (outside this PR, there are 158 public and 0 private Pydantic validators)

bofire/data_models/features/categorical.py
```diff
@@ -84,14 +87,46 @@ class CategoricalInput(Input, DescriptorsMixin):
...
+            >>> CategoricalInput.from_df("solvent", df)
+        """
+        return cls(
+            key=key,
+            categories=list(df.index),
+            allowed=[True for _ in range(len(df))],
+            descriptors=Descriptors(
+                columns={col: df[col].tolist() for col in df.columns}
+            ),
+        )
```
TobyBoyne: Maybe this should be `Descriptor.from_df`, then `CategoricalInput.from_descriptor`? Like:
```python
>> df = pd.DataFrame(
...     {"logP": [-1.4, -0.3], "MW": [18.0, 46.0]},
...     index=["water", "ethanol"],
... )
>>> descriptor = Descriptor.from_df(df)
>>> CategoricalInput.from_descriptor("ethanol", descriptor)
```

Not necessary, I mostly just find it odd that `CategoricalInput` is the only `Feature` to have a `from_df` method

TobyBoyne (approved):
Everything looks good from what I've read! Some small comments but nothing blocking a merge. Thanks for making the changes!

bofire/data_models/features/categorical.py
jduerholt: I removed it, you are completly right here!

bofire/data_models/features/descriptors.py
jduerholt: Good catch.

bofire/data_models/features/descriptors.py
jduerholt: Again a good catch.

bofire/data_models/encodings/naming.py
jduerholt: Good catch.

bofire/data_models/features/_descriptor_spec.py
jduerholt: Will fix it.

PR #790, opened 2026-07-02
callback function for GA, file-based progress tracking

LukasHebing:
The GA function now accepts a user-defined callback-function, in the systax given by pymoo:

```python
    def _callback(algorithm):
        callback_path.write_text(f"generation={algorithm.n_gen}\n", encoding="utf-8")

    x_opt, f_opt = run_ga(
        data_model=optimizer,
        domain=ga_domain,
        objective_callables=[ga_objective],
        q=1,
        callable_format="torch",
        callback=_callback,
    )
 ```
 
 For the high-level acquisition optimization, a default function is defined which exports basic stats to a user-specified csv file:
 
 ```python
optimizer = strategies_data_models.GeneticAlgorithmOptimizer(
    population_size=100,
    n_max_gen=100,
    verbose=False,
    # Write per-generation GA progress so it can be inspected while optimization runs.
    ga_progress_csv_path="_custom_ga_progress.csv",
)
```

This can then be tracked by other processes during execution and may be used for status updates in web-apps etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -114,6 +114,7 @@ venv/
 ENV/
 env.bak/
 venv.bak/
+uv.lock
 
 # Spyder project settings
 .spyderproject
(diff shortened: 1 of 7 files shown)


jduerholt (approved):
Looks good to me!

LukasHebing:
that was fast @jduerholt :)
Thanks!

bertiqwerty:
Nice!

Issue #777, opened 2026-06-17
[Bug]: Strategy.tell mutates caller DataFrame views, triggering repeated pandas SettingWithCopyWarning


rosonaeldred:
### What happened?

Passing a sliced pandas DataFrame into BoFire Strategy.tell triggers multiple SettingWithCopyWarning messages because BoFire validation/normalization writes into the provided frame instead of owning a copy first.

Request to deal with how people tend to pass data so don't constant generate SettingWithCopy warnings. 

Observed behavior

- Calling tell with a row-sliced DataFrame view emits many SettingWithCopyWarning warnings.
- In my local run, a single tell call emitted 15 warnings.

Expected behavior

- Strategy.tell should not mutate caller-owned DataFrames directly.
- Passing a slice/view should not produce SettingWithCopyWarning.
- BoFire should copy input at API boundary before in-place validation/column assignment.

Proposed fix
At BoFire boundary, copy input frames before validation/mutation in Strategy.tell (and defensively in set_experiments/add_experiments for direct callers). This removes the need for every downstream client to do data.copy() manually.

Optional workaround for users today
Call tell with an owned frame, for example using data.copy() or data.iloc[n:].copy().

Sketch solution:
```
diff --git a/bofire/strategies/strategy.py b/bofire/strategies/strategy.py
index <old>..<new> 100644
--- a/bofire/strategies/strategy.py
+++ b/bofire/strategies/strategy.py
@@ -104,6 +104,13 @@ class Strategy(BaseModel):
     def tell(
         self,
         experiments: pd.DataFrame,
         replace: bool = False,
     ) -> None:
@@
         if len(experiments) == 0:
             return
+        # BoFire mutates/normalizes frames during validation; own caller input
+        # to avoid mutating views and triggering pandas SettingWithCopyWarning.
+        experiments = experiments.copy(deep=True)
+
         if replace:
             self.set_experiments(experiments=experiments)
         else:
             self.add_experiments(experiments=experiments)
         self._tell()

@@ -306,6 +313,10 @@ class Strategy(BaseModel):
     def set_experiments(self, experiments: pd.DataFrame):
         """Set experiments of the strategy. Overwrites existing ones."""
+        # Defensive copy for callers invoking set_experiments directly.
+        experiments = experiments.copy(deep=True)
         experiments = self.domain.validate_experiments(experiments)
         self._experiments = experiments

@@ -316,6 +327,10 @@ class Strategy(BaseModel):
[...]

PR #791, opened 2026-07-13
Update dependency from pandas to 3.0.0 to implement standardized WriteOnCopy-Behaviour on Pandas Dataframes. Refers to issue #777.

kademirci:
## Motivation

In Issue #777 pandas of version prior to 3.0.0 can cause SettingToCopyWarning. In even older pandas version evoking Strategy.tell with a spliced view could change the original dataframe in theory.

To mitigate this we can increase the dependency of pandas to atleast 3.0.0 to implement the standardized CopyOnWrite-Behaviour, where a pandas.Dataframe automaticly delivers a copy (instead of a view) when doing a filter-operation or something similar to it. The SettingWithCopyWarning would also not appear anymore.

I don't believe that would cause any new issues since it would restrict the version and not expand it. Please correct me if I'm wrong.

Y

No

## Test Plan

Unit tests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.11.0, <3.15"
 dependencies = [
     "numpy<2.5.0", # we pin due to some incompatibility with shap under uv dependency resolution
-    "pandas",
+    "pandas>=3.0.0",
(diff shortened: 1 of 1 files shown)


jduerholt (approved):
Thanks! Looks good to me! Closes https://github.com/experimental-design/bofire/issues/777

PR #796, opened 2026-07-30
Fix allow_zero + categorical crashing optimize_acqf_mixed

### LINKED ISSUE #795: [Bug]: allow_zero flag + BoTorch 0.18.x combines to disaster :   allow_zero=True on multiple ContinuousInputs causes optimize_acqf_mixed failure with BoTorch >=0.18 ("stack expects each tensor to be equal size")
### What happened?

I might end up opening multiple tickets on this, as there are multiple issues we're seeing with this combination. 
Could we possibly set it to optional instead of default false? 

Issue for this particular ticket: allow_zero=True on multiple ContinuousInputs causes optimize_acqf_mixed failure with BoTorch >=0.18 (`"stack expects each tensor to be equal size"`)

When a BoFire domain contains multiple ContinuousInputs with allow_zero=True, optimization can fail during ask() with a runtime error originating from BoTorch's mixed optimization path.

The issue appears when more than one input has allow_zero=True. BoFire internally treats these features as having optional zero values, which results in a mixed-search-space optimization. In our case, a domain with three continuous inputs and allow_zero=True on each input leads to 2³ = 8 categorical combinations.
With BoTorch 0.18.x, candidate generation fails in `optimize_acqf_mixed` with:

`RuntimeError: stack expects each tensor to be equal size, but got [1] at entry 0 and [3] at entry 1`

The same workflow worked previously with older BoTorch versions (e.g. 0.13.x development versions used by our stack).

Interestingly, requests may succeed when only a single proposal is requested or when specific optimizer paths are taken, but fail for other proposal requests, making the behavior appear inconsistent.

## Expected Behavior: 
Domains using multiple continuous inputs with allow_zero=True should successfully generate candidates.
BoTorch's mixed optimizer should handle all discrete combinations generated by BoFire and return acquisition values with consistent tensor shapes.

##Actual Behavior
`ask()` fails inside BoTorch's mixed acquisition optimization:
`     RuntimeError: stack expects each tensor to be equal size, but got [1] at entry 0 and [3] at entry 1`

The exception originates from:

`     botorch.optim.optimize.optimize_acqf_mixed`

specifically during:
`     ff_acq_values = torch.stack(ff_acq_value_list)`

where acquisition values returned for different fixed-feature configurations appear to have different tensor shapes.

## Additional Observations
-	The issue disappears when allow_zero=False is used for the inputs. (but this causes other issues, expect other tickets...)
-	The actual lower bound value does not appear to matter. We reproduced the issue both with bounds starting at 0 and bounds starting above 0.
-	The problematic condition appears to be having multiple allow_zero=True continuous inputs, which causes BoFire to expose multiple fixed-feature combinations to BoTorch.
-	The issue seems to have appeared after upgrading to BoTorch 0.18.x.
-	Older environments using BoTorch 0.13.x did not exhibit this behavior.

## Proposed Fix
Remove the flag, or find a way to make it optional. 
Otherwise, might need to escalate to a Botorch issue. 

It looks like `optimize_acqf_mixed() `assumes acquisition values returned from all fixed-feature configurations have identical shapes. A defensive normalization (or handling of scalar vs batched acquisition values) before the final torch.stack() may resolve the issue.

At minimum, it would be helpful if BoTorch could either:
1.	Normalize acquisition value shapes prior to stacking, or
2.	Raise a more informative error identifying which fixed-feature configurations produced inconsistent outputs.

## History of allow_zero, and BoTorch's role in this:
Note: allow_zero was introduced in the May Hackathon for the BONSAI pruning methods (f563e534, @jduerholt  + claude).

### Recent Code Changes (Timeline)

**May 20, 2026 (commit f563e534): BONSAI pruning added**
- Introduced `relax_allow_zero` parameter to `get_torch_bounds_from_domain()`
- When pruning is applicable (semi-continuous features exist), bounds are relaxed for botorch
- Pruning post-processes botorch's output to enforce the gaps

**June 12, 2026 (commit 396f629d): BoTorch version bumped**
- Upgraded from `botorch>=0.18.0` to `botorch[fully_bayesian]>=0.18.1`
- Happened AFTER the relax_allow_zero logic was introduced

### Please provide a minimal, reproducible example of the unexpected behavior.
```shell
1     import pandas as pd
2     
3     from bofire.data_models.domain.api import Domain
4     from bofire.data_models.features.api import (
5         ContinuousInput,
6         ContinuousOutput,
7     )
8     from bofire.data_models.strategies.api import PolytopeSampler
9     from bofire.strategies.api import SoboStrategy
10     from bofire.data_models.strategies.predictives.sobo import SoboStrategy as SoboDM
11     
12     
13     domain = Domain.from_lists(
14         inputs=[
15             ContinuousInput(
16                 key="x1",
17                 bounds=(0.0, 20.0),
18                 allow_zero=True,
19             ),
20             ContinuousInput(
21                 key="x2",
22                 bounds=(1.0, 20.0),
23                 allow_zero=True,
24             ),
25             ContinuousInput(
26                 key="x3",
27                 bounds=(1.0, 20.0),
28                 allow_zero=True,
29             ),
30         ],
31         outputs=[
32             ContinuousOutput(key="y"),
33         ],
34     )
35     
36     experiments = pd.DataFrame(
37         {
38             "x1": [5.0, 10.0, 15.0, 18.0],
39             "x2": [5.0, 10.0, 15.0, 8.0],
40             "x3": [5.0, 10.0, 15.0, 12.0],
41             "y": [1.0, 2.0, 3.0, 4.0],
42             "valid_y": [1, 1, 1, 1],
43         }
44     )
45     
46     strategy = SoboStrategy(
47         data_model=SoboDM(
48             domain=domain,
49         )
50     )
51     
52     strategy.tell(experiments)
53     
54     strategy.ask(candidate_count=2)
```

### Please paste any relevant traceback/logs produced by the example provided.

```shell
1     File ".../botorch/optim/optimize.py", line 1332, in optimize_acqf_mixed
2         ff_acq_values = torch.stack(ff_acq_value_list)
3     RuntimeError: stack expects each tensor to be equal size, but got [1] at entry 0 and [3] at entry 1
```

Full stack trace indicates the failure occurs along the following path:
1     BoFire
2      -> AcqfOptimizer
3      -> optimize_acqf_list
4      -> optimize_acqf_mixed
5      -> torch.stack(ff_acq_value_list)

### BoFire Version

0.4x

### Python Version

3.12

### Operating System

Linux

jduerholt:
Closes #795

## Problem

`f563e534` added an `include_semicontinuous` flag to `Inputs.get_number_of_categorical_combinations`, and `BotorchOptimizer._determine_optimizer` passes `not is_pruning_applicable(domain)` — so that semi-continuous features (`allow_zero=True` with `lb > 0`) are resolved by the post-AF BONSAI pruning step rather than by enumerating their on/off states.

The sibling `Inputs.get_categorical_combinations` — which builds the actual `fixed_features_list` — never got the flag. So the routing decision was contradicted at list-construction time. Whenever routing lands on `OPTIMIZE_ACQF_MIXED` / `OPTIMIZE_ACQF_LIST` for some *other* reason (a categorical or discrete input, or multiple acqfs), the semi-continuous states were still enumerated, and the features got handled twice: once by enumeration, once by pruning.

Beyond the wasted work, this crashes. The all-features-off combination pins *every* tensor column, and botorch short-circuits that case in `_optimize_acqf_all_features_fixed`, which under `return_best_only=False` collapses the restart dimension to shape `(1,)`. Combinations that leave a column free return `(num_restarts,)`. `optimize_acqf_mixed` stacks these unguarded:

```
RuntimeError: stack expects each tensor to be equal size, but got [1] at entry 0 and [20] at entry 1
```

Reproduced on `main` with three `allow_zero` inputs plus one `CategoricalInput`:

```
routed: OPTIMIZE_ACQF_MIXED  combos=16  cols=4  all-fixed combos=2  -> FAILED
```

Pure categorical/discrete domains are unaffected — there *every* combination pins every column, so the list stays homogeneous and stacks fine. Only the semi-continuous enumeration mixes the two kinds:

```
2 DiscreteInput only              combos=6   cols=2  all-fixed=6/6   ok
DiscreteInput + CategoricalInput  combos=6   cols=2  all-fixed=6/6   ok
3x allow_zero + categorical       combos=16  cols=4  all-fixed=2/16  FAILED
```

## Fix

- Add `include_semicontinuous: bool = True` to `Inputs.get_categorical_combinations`, mirroring the counting function.
- `BotorchOptimizer.get_categorical_combinations` passes `not is_pruning_applicable(domain)`.
- The `n_combos` guard in the `OPTIMIZE_ACQF_LIST` branch was also calling the counting function without the flag, and would have disagreed with the list it guards; it now passes the same value.

After the fix the same domain yields `combos=2  all-fixed=0` and `ask` succeeds, with pruning still zeroing the semi-continuous features as intended.

## Tests

Three regression tests, each verified to fail without the source change:

- `test_inputs_get_categorical_combinations_exclude_semicontinuous` — data-model level; asserts the enumeration drops semi-continuous features, and that both functions agree for **both** values of the flag (the invariant that was broken).
- `test_get_categorical_combinations_excludes_semicontinuous_when_pruning` — 16 → 2 combinations, and no combination pins every tensor column.
- `test_ask_with_semicontinuous_and_categorical` — end-to-end `ask(candidate_count=2)` on the issue's domain plus a categorical, asserting each `allow_zero` feature lands on zero or inside `[lb, ub]`, never in the forbidden gap.

`tests/bofire/data_models` 1368 passed; `tests/bofire/strategies` 506 passed. The two `doe/test_utils.py::test_n_zero_eigvals_*` failures are pre-existing on a clean tree and unrelated. `ty check bofire` reports 146 diagnostics before and after.

## Follow-up (not in this PR)

The botorch side is a genuine upstream bug: `optimize_acqf_mixed` assumes every entry of `fixed_features_list` yields the same acqf-value shape. This PR stops BoFire from constructing the heterogeneous list that triggers it, but the latent issue should be filed upstream so it can't resurface from another direction.

### DIFF
--- a/bofire/data_models/domain/features.py
+++ b/bofire/data_models/domain/features.py
@@ -569,6 +569,7 @@ class Inputs(_BaseFeatures[AnyInput]):
         exclude: Union[
             Type, List[Type]
         ] = None,  # ty: ignore[invalid-parameter-default]
+        include_semicontinuous: bool = True,
     ) -> list[tuple[tuple[str, float] | tuple[str, str], ...]]:
         """Get a list of tuples pairing the feature keys with a list of valid categories
 
@@ -576,6 +577,16 @@ class Inputs(_BaseFeatures[AnyInput]):
             include (Feature, optional): Features to be included. Defaults to Input.
             exclude (Feature, optional): Features to be excluded, e.g. subclasses
                 of the included features. Defaults to None.
+            include_semicontinuous (bool, optional): When True (default), each
+                semi-continuous feature (`ContinuousInput` with `allow_zero=True`
+                and a positive lower bound, un-fixed) doubles the number of
+                combinations via its on/off enumeration. When False,
+                semi-continuous features are excluded from the enumeration --
+                useful when the caller handles them via a post-optimisation
+                pruning step rather than by AF-time enumeration. Mirrors the
+                flag of the same name on
+                `get_number_of_categorical_combinations`; the two must be
+                passed the same value to stay consistent.
 
         Returns:
             List[(str, List[str])]: Returns a list of tuples pairing the feature
@@ -602,11 +613,14 @@ class Inputs(_BaseFeatures[AnyInput]):
         cat_and_discrete_values = cat_values + discrete_values
         all_combos = list(itertools.product(*cat_and_discrete_values))
 
-        conditional_conts = [
-            f
-            for f in self.get(includes=include, excludes=exclude)
-            if (isinstance(f, ContinuousInput) and f.is_semicontinuous)
-        ]
+        if include_semicontinuous:
+            conditional_conts = [
+                f
+                for f in self.get(includes=include, excludes=exclude)
+                if (isinstance(f, ContinuousInput) and f.is_semicontinuous)
+            ]
+        else:
+            conditional_conts = []
 
         conditional_values = [[(d.key, 0.0), (d.key, None)] for d in conditional_conts]
 
--- a/bofire/strategies/predictives/acqf_optimization.py
+++ b/bofire/strategies/predictives/acqf_optimization.py
@@ -729,7 +729,9 @@ class BotorchOptimizer(AcquisitionOptimizer):
                 fixed_features_list=self.get_categorical_combinations(domain),
             )
         elif optimizer == OptimizerEnum.OPTIMIZE_ACQF_LIST:
-            n_combos = domain.inputs.get_number_of_categorical_combinations()
+            n_combos = domain.inputs.get_number_of_categorical_combinations(
+                include_semicontinuous=not is_pruning_applicable(domain),
+            )
             return _OptimizeAcqfListInput(
                 acq_function_list=acqfs,
                 bounds=bounds,
@@ -786,7 +788,17 @@ class BotorchOptimizer(AcquisitionOptimizer):
         """
         fixed_basis = self.get_fixed_features(domain=domain)
 
-        combos = domain.inputs.get_categorical_combinations()
+        # Must use the same `include_semicontinuous` value as
+        # `_determine_optimizer`: when pruning is applicable, semi-continuous
+        # features are resolved by the post-AF pruning step, so enumerating
+        # their on/off states here would handle them twice. It would also
+        # produce a `fixed_features_list` in which some entries pin *every*
+        # tensor column while others do not -- `optimize_acqf_mixed` stacks
+        # the per-entry acqf values and the all-fixed shortcut in botorch
+        # collapses the restart dimension, so the stack fails.
+        combos = domain.inputs.get_categorical_combinations(
+            include_semicontinuous=not is_pruning_applicable(domain),
+        )
         # now build up the fixed feature list
         if len(combos) == 1:
             return [fixed_basis]
--- a/tests/bofire/data_models/domain/test_inputs.py
+++ b/tests/bofire/data_models/domain/test_inputs.py
@@ -130,6 +130,41 @@ def test_inputs_get_categorical_combinations_conditional():
     )
 
 
+def test_inputs_get_categorical_combinations_exclude_semicontinuous():
+    """`include_semicontinuous=False` must drop the on/off enumeration of
+    semi-continuous features, staying consistent with the flag of the same
+    name on `get_number_of_categorical_combinations`.
+    """
+    inputs = Inputs(
+        features=[
+            CategoricalInput(key="f1", categories=["c11", "c12"]),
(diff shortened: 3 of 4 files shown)


jduerholt:
It looks good for me, @Jimbo994 can you review it?

Jimbo994:
It looks good to me. I added one additional test to verify the correct handling of semi-continuous inputs when multiple acquisition functions are used. Maybe have a look if you deem it necessary, or not.

Jimbo994 (approved):

jduerholt:
@Jimbo994: Thanks for the review and the test! Looks good to me.

PR #797, opened 2026-07-30
fix: make registered engineered features visible to Features.get()

### LINKED ISSUE #794: [Bug]: Custom engineered features registered via `register_engineered_feature` are silently dropped by `Features.get()` / `get_keys()`
### What happened?

**BoFire version:** 0.4.1

#### Summary
Registering a custom engineered-feature data model through the public
registration API makes Pydantic *accept* it (it is stored in
`EngineeredFeatures.features`), but the container's `get()` / `get_keys()`
methods silently omit it. The default `includes` filter of these methods is
`AnyFeature`, and `register_engineered_feature` updates
`AnyEngineeredFeature` (and the `EngineeredFeatures.features` field union) but
**not** `AnyFeature`. As a result, a newly registered feature type is retained
but invisible to the primary accessor.

#### Root cause
- `bofire/data_models/domain/features.py` — `_BaseFeatures.get()` and
  `get_keys()` both default `includes: ... = AnyFeature`. This default is a
  module-level union bound at function-definition (import) time.
- `bofire/data_models/features/_register.py` — `register_engineered_feature()`
  appends to `_ENGINEERED_FEATURE_TYPES`, rebinds
  `features_api.AnyEngineeredFeature`, and extends the `EngineeredFeatures`
  `features` field union, but does **not** touch `AnyFeature`.
- `bofire/data_models/features/api.py` — `AnyFeature` is a separate,
  independent `tagged_union(...)`. Because it is never updated, the custom
  type is not a subclass of any member, so `filter_by_class` (which does
  `isinstance(x, tuple(to_list(AnyFeature)))`) excludes it.

### Please provide a minimal, reproducible example of the unexpected behavior.

```python
from typing import ClassVar, Literal

import bofire.surrogates.engineered_features as ef
import bofire.data_models.features.api as fapi
from bofire.data_models.features.engineered_feature import EngineeredFeature
from bofire.data_models.domain.features import EngineeredFeatures
from bofire.data_models.unions import to_list


class FreshFeature(EngineeredFeature):
    type: Literal["FreshFeature"] = "FreshFeature"
    order_id: ClassVar[int] = 99
    features: list[str]

    @property
    def n_transformed_inputs(self) -> int:
        return 1


def _map(inputs, transform_specs, feature):  # dummy aggregate map
    return None


ef.register(FreshFeature, _map)

feat = FreshFeature(key="fresh", features=["a", "b"])
container = EngineeredFeatures(features=[feat])

print(container.features)                              # [FreshFeature(...)]  -> stored
print(container.get_keys())                            # []                   -> DROPPED
print(container.get_keys(includes=EngineeredFeature))  # ['fresh']            -> workaround
print(FreshFeature in to_list(fapi.AnyFeature))            # False
print(FreshFeature in to_list(fapi.AnyEngineeredFeature))  # True
```

### Please paste any relevant traceback/logs produced by the example provided.

```shell

```

### BoFire Version

0.4.1

### Python Version

_No response_

### Operating System

_No response_

jduerholt:
Fixes #794.

## The bug

Registering a custom engineered feature via `register_engineered_feature` makes Pydantic accept it, but `get()` / `get_keys()` silently omit it.

This is not cosmetic. Every accessor that relies on the default filter degrades, and the failure is silent end-to-end — a surrogate trains **without** the feature and **without** any error or warning:

```
engineered_features.get_keys() : []
n_transformed_inputs           : 0
get_features2idx()             : {}
model train_inputs shape       : torch.Size([10, 3])   ← expected [10, 4]
input_transform                : Normalize()           ← AppendFeatures never attached
```

`validate_inputs()` also becomes a no-op, so a feature referencing a nonexistent input key passes validation.

## Root cause

Three independent defects, all the same underlying mistake: **binding a mutable registry at import time**.

**1. Def-time default binding.** `_BaseFeatures.get()`/`get_keys()` bound `AnyFeature` as a default argument. Python evaluates that once at import and stores it in `__defaults__`:

```python
_BaseFeatures.get.__defaults__[0] is fapi.AnyFeature  # True
```

`tagged_union()` returns a *new* object each call, so rebinding the module global can never reach the default. Worth stating plainly: **the "just rebuild `AnyFeature`" fix alone does not work** — I verified it leaves `get_keys()` returning `[]`.

**2. Stale cross-module import.** `domain/features.py` does `from ...features.api import AnyEngineeredFeature`, so `__add__`'s `is_engineeredfeats` reads a permanently stale union. `EngineeredFeatures(...) + [custom_feat]` raised `union_tag_invalid`.

**3. Registry drift.** `register_engineered_feature` updated `AnyEngineeredFeature` but not `AnyFeature`, so `Features(features=[custom_feat])` raised. `AnyFeature` already lists all nine built-in engineered types, so it was clearly meant to cover them.

## The fix

- **`get()`/`get_keys()`** default to a `_Unset` sentinel and resolve the filter at call time from `model_fields["features"].annotation` — the annotation registration patches in place. Reading the class's generic parameter instead would be equally stale (`__orig_bases__` is frozen at class-creation time).

  A sentinel is required rather than `None`: `includes=None` already means "no include filter" and must keep that meaning. Collapsing the two states would have broken `sobo.py:73` and three existing assertions in `test_features.py`.

- **`__add__`** filters on the `EngineeredFeature` base class, mirroring how `is_infeats`/`is_outfeats` already use `Input`/`Output`.

- **`register_engineered_feature`** keeps `AnyFeature` and `Features.features` in sync, via a new `_FEATURE_TYPES` registry mirroring the existing `_ENGINEERED_FEATURE_TYPES` pattern. It sits inside the block already guarded by `register_into`, inheriting idempotency and duplicate-discriminator handling.

`_Unset` is a class rather than `object()` so it satisfies the `Type` member of the parameter annotation — this lets two now-dead `# ty: ignore[invalid-parameter-default]` directives be dropped without introducing new errors.

## Reviewer note

`get()`'s default now resolves per container: `Inputs.get()` filters by `AnyInput` rather than `AnyFeature`. This is a no-op today — Pydantic already guarantees an `Inputs` container holds only `AnyInput` members — but there are ~59 no-arg `.get()`/`.get_keys()` call sites, so it deserves a deliberate look rather than being buried in the diff.

## Tests

The existing registration tests asserted only `.features` and `len(.features)` — precisely the two things that already worked. That gap is what let this ship.

16 new tests. Verified they fail without the fix: reverting only the source, **8 of 13** failed; reverting only Part A (keeping the rest), **exactly the 3** `AnyFeature` tests failed. The tests that pass in both states are the `includes=None` semantics guards, which is intended — they exist to prove nothing moved.

The end-to-end test fits a real `SingleTaskGPSurrogate` and asserts `train_inputs[0].shape[-1] == 4` (3 raw + 1 appended). That is the silent model corruption, caught directly.

## Validation

| | Result |
|---|---|
| `tests/bofire/data_models` + `test_register.py` | 1423 passed, 5 skipped |
| `tests/bofire/surrogates` + `strategies` | 789 passed, 62 skipped, 2 failed |
| `utils`, `kernels`, `priors`, `outlier_detection`, `plot` | 284 passed |
| `ruff check` / `format` | clean |
| `ty check bofire` | 144 diagnostics, down from 146 on baseline |

The 2 failures are `test_n_zero_eigvals_unconstrained` / `_constrained` in `tests/bofire/strategies/doe/test_utils.py`. I confirmed they fail identically on a clean tree with this change reverted — pre-existing, DoE eigenvalue utilities, unrelated to this PR.

## Known limitations, not addressed here

- `data_models/api.py:70` computes `AnyThing` at import time, so it never sees post-import registrations. This affects **every** registrar (kernels, priors, strategies), not just features. Its only consumer is a schema test that parametrizes at collection, so it is inert today. Worth a separate issue.
- The now-unused `AnyEngineeredFeature` import in `domain/features.py` is still stale after registration. Its remaining uses are a runtime-noop `cast()` and the class generic parameter (which Pydantic resolves into `model_fields`, where registration patches it), so it is harmless — but it is a trap for the next reader. Left alone to keep this diff focused.

### DIFF
--- a/bofire/data_models/domain/features.py
+++ b/bofire/data_models/domain/features.py
@@ -18,8 +18,6 @@ from typing import (
     TypeVar,
     Union,
     cast,
-    get_args,
-    get_origin,
 )
 
 import numpy as np
@@ -62,38 +60,6 @@ from bofire.data_models.unions import to_list
 
 F = TypeVar("F", bound=AnyFeature)
 FeatureSequence = Sequence[F]
-
-
-class _Unset:
-    """Sentinel for the default ``includes`` filter of ``get``/``get_keys``.
-
-    A distinct sentinel is needed because ``includes=None`` already means "do
-    not filter by class", so it cannot double as "not supplied". This is a
-    class rather than a plain ``object()`` so that it satisfies the ``Type``
-    part of the parameter annotation.
-    """
-
-
-def _element_union(cls: Type["_BaseFeatures"]):
-    """Return the union of feature types *cls* currently accepts.
-
-    Resolved from ``model_fields`` instead of the generic parameter of the
-    class: ``register_engineered_feature`` patches the field annotation in
-    place, so this reflects what the container actually validates, while
-    ``__orig_bases__`` is frozen at class-creation time. Binding the union at
-    import time (as a default argument would) makes newly registered feature
-    types invisible to the filter.
-    """
-    annotation = cls.model_fields["features"].annotation
-    if get_origin(annotation) in (list, tuple, Sequence):
-        args = get_args(annotation)
-        if args:
-            annotation = args[0]
-    if isinstance(annotation, TypeVar):
-        # ``_BaseFeatures`` itself is unparametrized, so its annotation is a
-        # bare TypeVar, which ``isinstance`` cannot consume.
-        return Feature
-    return annotation
 
 
 class _BaseFeatures(BaseModel, Generic[F]):
@@ -216,7 +182,7 @@ class _BaseFeatures(BaseModel, Generic[F]):
 
     def get(
         self,
-        includes: Union[Type, List[Type], None] = _Unset,
+        includes: Union[Type, List[Type], None] = Feature,
         excludes: Union[Type, List[Type], None] = None,
         exact: bool = False,
     ) -> Self:
@@ -224,9 +190,8 @@ class _BaseFeatures(BaseModel, Generic[F]):
 
         Args:
             includes: All features in this container that are instances of an
-                include are returned. If not provided, the feature types this
-                container accepts are used. If None, the include filter is not
-                active.
+                include are returned. Defaults to the ``Feature`` base class,
+                i.e. everything. If None, the include filter is not active.
             excludes: All features in this container that are not instances of
                 an exclude are returned. If None, the exclude filter is not active.
             exact: Boolean to distinguish if only the exact class listed in
@@ -237,8 +202,6 @@ class _BaseFeatures(BaseModel, Generic[F]):
             List of features in the domain fitting to the passed requirements.
 
         """
-        if includes is _Unset:
-            includes = _element_union(type(self))
         return self.__class__(
             features=sorted(
                 filter_by_class(
@@ -252,7 +215,7 @@ class _BaseFeatures(BaseModel, Generic[F]):
 
     def get_keys(
         self,
-        includes: Union[Type, List[Type], None] = _Unset,
+        includes: Union[Type, List[Type], None] = Feature,
         excludes: Union[Type, List[Type], None] = None,
         exact: bool = False,
     ) -> List[str]:
@@ -260,9 +223,8 @@ class _BaseFeatures(BaseModel, Generic[F]):
 
         Args:
             includes: All features in this container that are instances of an
-                include are returned. If not provided, the feature types this
-                container accepts are used. If None, the include filter is not
-                active.
+                include are returned. Defaults to the ``Feature`` base class,
+                i.e. everything. If None, the include filter is not active.
             excludes: All features in this container that are not instances of
                 an exclude are returned. If None, the exclude filter is not active.
             exact: Boolean to distinguish if only the exact class listed in
--- a/bofire/data_models/domain/features.py
+++ b/bofire/data_models/domain/features.py
@@ -18,6 +18,8 @@ from typing import (
     TypeVar,
     Union,
     cast,
+    get_args,
+    get_origin,
 )
 
 import numpy as np
@@ -41,6 +43,7 @@ from bofire.data_models.features.api import (
     ContinuousInput,
     ContinuousOutput,
     DiscreteInput,
+    EngineeredFeature,
     Feature,
     Input,
     Output,
@@ -59,6 +62,38 @@ from bofire.data_models.unions import to_list
 
 F = TypeVar("F", bound=AnyFeature)
 FeatureSequence = Sequence[F]
+
+
+class _Unset:
+    """Sentinel for the default ``includes`` filter of ``get``/``get_keys``.
+
+    A distinct sentinel is needed because ``includes=None`` already means "do
+    not filter by class", so it cannot double as "not supplied". This is a
+    class rather than a plain ``object()`` so that it satisfies the ``Type``
+    part of the parameter annotation.
+    """
+
+
+def _element_union(cls: Type["_BaseFeatures"]):
+    """Return the union of feature types *cls* currently accepts.
+
+    Resolved from ``model_fields`` instead of the generic parameter of the
+    class: ``register_engineered_feature`` patches the field annotation in
+    place, so this reflects what the container actually validates, while
+    ``__orig_bases__`` is frozen at class-creation time. Binding the union at
+    import time (as a default argument would) makes newly registered feature
+    types invisible to the filter.
+    """
+    annotation = cls.model_fields["features"].annotation
+    if get_origin(annotation) in (list, tuple, Sequence):
+        args = get_args(annotation)
+        if args:
+            annotation = args[0]
+    if isinstance(annotation, TypeVar):
+        # ``_BaseFeatures`` itself is unparametrized, so its annotation is a
+        # bare TypeVar, which ``isinstance`` cannot consume.
+        return Feature
+    return annotation
 
 
 class _BaseFeatures(BaseModel, Generic[F]):
@@ -115,7 +150,10 @@ class _BaseFeatures(BaseModel, Generic[F]):
             return is_feats_of_type(feats, Outputs, Output)
 
         def is_engineeredfeats(feats):
-            return is_feats_of_type(feats, EngineeredFeatures, AnyEngineeredFeature)
+            # Filter on the base class, mirroring ``is_infeats``/``is_outfeats``.
+            # The ``AnyEngineeredFeature`` union imported here is bound at
+            # import time and would not see registered feature types.
+            return is_feats_of_type(feats, EngineeredFeatures, EngineeredFeature)
 
         if is_infeats(self) and is_infeats(other):
             return Inputs(features=cast(Tuple[AnyInput, ...], new_feature_seq))
@@ -178,9 +216,7 @@ class _BaseFeatures(BaseModel, Generic[F]):
 
     def get(
         self,
-        includes: Union[
-            Type, List[Type], None
-        ] = AnyFeature,  # ty: ignore[invalid-parameter-default]
+        includes: Union[Type, List[Type], None] = _Unset,
         excludes: Union[Type, List[Type], None] = None,
         exact: bool = False,
     ) -> Self:
@@ -188,7 +224,9 @@ class _BaseFeatures(BaseModel, Generic[F]):
 
         Args:
             includes: All features in this container that are instances of an
-                include are returned. If None, the include filter is not active.
+                include are returned. If not provided, the feature types this
+                container accepts are used. If None, the include filter is not
+                active.
             excludes: All features in this container that are not instances of
                 an exclude are returned. If None, the exclude filter is not active.
             exact: Boolean to distinguish if only the exact class listed in
@@ -199,6 +237,8 @@ class _BaseFeatures(BaseModel, Generic[F]):
             List of features in the domain fitting to the passed requirements.
 
         """
+        if includes is _Unset:
+            includes = _element_union(type(self))
         return self.__class__(
             features=sorted(
                 filter_by_class(
@@ -212,9 +252,7 @@ class _BaseFeatures(BaseModel, Generic[F]):
 
     def get_keys(
         self,
-        includes: Union[
-            Type, List[Type], None
-        ] = AnyFeature,  # ty: ignore[invalid-parameter-default]
+        includes: Union[Type, List[Type], None] = _Unset,
         excludes: Union[Type, List[Type], None] = None,
         exact: bool = False,
     ) -> List[str]:
@@ -222,7 +260,9 @@ class _BaseFeatures(BaseModel, Generic[F]):
 
         Args:
             includes: All features in this container that are instances of an
-                include are returned. If None, the include filter is not active.
+                include are returned. If not provided, the feature types this
+                container accepts are used. If None, the include filter is not
+                active.
             excludes: All features in this container that are not instances of
                 an exclude are returned. If None, the exclude filter is not active.
             exact: Boolean to distinguish if only the exact class listed in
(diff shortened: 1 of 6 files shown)


tests/bofire/test_register.py
```diff
@@ -918,6 +918,192 @@ def n_transformed_inputs(self) -> int:
...
+
+    type: Literal["_VisibleEngineered"] = "_VisibleEngineered"
+    order_id = 102
+
+    @property
+    def n_transformed_inputs(self) -> int:
+        return 1
+
+
+def _map_visible_feature(inputs, transform_specs, feature):
```
jduerholt: We do not need this map for the test, or?

tests/bofire/test_register.py
```diff
@@ -918,6 +918,192 @@ def n_transformed_inputs(self) -> int:
...
+    def test_get_sees_registered_feature(self, container):
+        assert [f.key for f in container.get()] == ["engineered"]
+
+    def test_n_transformed_inputs_counts_registered_feature(self, container):
+        assert container.n_transformed_inputs == 1
+
+    def test_features2idx_maps_registered_feature(self, container):
+        assert container.get_features2idx() == {"engineered": (0,)}
+
+    def test_validate_inputs_is_not_skipped(self, container):
```
jduerholt: Not needed here this test.

tests/bofire/test_register.py
```diff
@@ -918,6 +918,192 @@ def n_transformed_inputs(self) -> int:
...
+        """The default filter fed ``validate_inputs``, so it silently passed."""
+        from bofire.data_models.domain.api import Inputs
+
+        inputs = Inputs(features=[ContinuousInput(key="a", bounds=(0, 1))])
+        with pytest.raises(ValueError, match="missing in inputs"):
+            container.validate_inputs(inputs)
+
+    def test_add_returns_engineered_features(self, container):
```
jduerholt: not needed.

tests/bofire/test_register.py
```diff
@@ -918,6 +918,192 @@ def n_transformed_inputs(self) -> int:
...
+        """``__add__`` filtered on a union bound at import time."""
+        from bofire.data_models.domain.features import EngineeredFeatures
+
+        combined = EngineeredFeatures(features=[]) + list(container.features)
+        assert isinstance(combined, EngineeredFeatures)
+        assert combined.get_keys() == ["engineered"]
+
+    def test_any_feature_union_is_kept_in_sync(self):
```
jduerholt: not needed

tests/bofire/test_register.py
```diff
@@ -918,6 +918,192 @@ def n_transformed_inputs(self) -> int:
...
+            AnyEngineeredFeature,
+            AnyFeature,
+        )
+        from bofire.data_models.unions import to_list
+
+        assert _VisibleEngineeredFeature in _FEATURE_TYPES
+        assert _VisibleEngineeredFeature in to_list(AnyFeature)
+        assert _VisibleEngineeredFeature in to_list(AnyEngineeredFeature)
+
+    def test_generic_features_container_accepts_registered_feature(self, container):
```
jduerholt, line 962: not needed

tests/bofire/test_register.py
```diff
@@ -918,6 +918,192 @@ def n_transformed_inputs(self) -> int:
...
+        """``Features.features`` is annotated with ``AnyFeature``."""
+        from bofire.data_models.domain.features import Features
+
+        generic = Features(features=list(container.features))
+        assert generic.get_keys() == ["engineered"]
+
+    def test_registered_feature_roundtrips_through_any_feature(self, container):
```
jduerholt: not needed

tests/bofire/test_register.py
```diff
@@ -918,6 +918,192 @@ def n_transformed_inputs(self) -> int:
...
+
+        import bofire.data_models.features.api as features_api
+
+        feature = container.features[0]
+        restored = TypeAdapter(features_api.AnyFeature).validate_python(
+            feature.model_dump()
+        )
+        assert restored == feature
+
+    def test_surrogate_appends_the_engineered_dimension(self, container):
```
jduerholt: not needed

tests/bofire/test_register.py
```diff
@@ -918,6 +918,192 @@ def n_transformed_inputs(self) -> int:
...
+        experiments = inputs.sample(10)
+        experiments["y"] = experiments.sum(axis=1)
+        experiments["valid_y"] = 1
+        surrogate.fit(experiments)
+
+        # 3 raw inputs + 1 appended engineered feature
+        assert surrogate.model.train_inputs[0].shape[-1] == 4
+
+
+class TestGetFilterDefaults:
```
jduerholt: These tests should not be in tests_register, mabye go in test modules for the features.

jduerholt (commented):
The test suite is to elaborate for this feature and we should also put the tests on the correct files.

tests/bofire/test_register.py
jduerholt: Correct — removed. The fixture now calls `register_engineered_feature` directly; none of the remaining tests touch the functional mapper.

tests/bofire/test_register.py
jduerholt: Removed.

tests/bofire/test_register.py
jduerholt: Removed.

tests/bofire/test_register.py
jduerholt: Removed.

tests/bofire/test_register.py
jduerholt: Removed.

tests/bofire/test_register.py
jduerholt: Removed.

tests/bofire/test_register.py
jduerholt: I kept one test here rather than dropping it outright, because this line is the only coverage for the third defect (`register_engineered_feature` never updated `AnyFeature`/`Features.features`) — the other four tests all exercise `EngineeredFeatures`, which is a different union. I did rewrite it to assert the behaviour instead of introspecting the union:

```python
def test_generic_features_container_accepts_registered_feature(self, container):
    assert Features(features=list(container.get())).get_keys() == ["engineered"]
```

Verified it is load-bearing: reverting only the `_register.py`/`api.py` part of the fix fails exactly this one test. Happy to drop it if you'd rather not cover that path.

tests/bofire/test_register.py
jduerholt: Moved. `test_exclude_include` in `tests/bofire/data_models/domain/test_features.py` already covers the `includes=None` / `excludes` combinations including the "no filter provided" error, so most of that class was redundant — I dropped it and added one small parametrized test next to it pinning that the no-argument default stays equivalent to `AnyFeature` for the built-in types.

jduerholt:
Hi @TobyBoyne,

could you review this PR? Maybe, you also have a better idea to solve the issue ;) 

Best,

Johannes

bofire/data_models/domain/features.py
```diff
@@ -61,6 +64,38 @@
...
+
+    Resolved from ``model_fields`` instead of the generic parameter of the
+    class: ``register_engineered_feature`` patches the field annotation in
+    place, so this reflects what the container actually validates, while
+    ``__orig_bases__`` is frozen at class-creation time. Binding the union at
+    import time (as a default argument would) makes newly registered feature
+    types invisible to the filter.
+    """
+    annotation = cls.model_fields["features"].annotation
+    if get_origin(annotation) in (list, tuple, Sequence):
```
TobyBoyne: This doesn't match all possible Sequences. Better as 
```python
if issubclass(get_origin(annotation), Sequence):
   ...
```

bofire/data_models/domain/features.py
```diff
@@ -178,17 +216,17 @@ def get_by_keys(self, keys: Sequence[str], include: bool = True) -> Self:
 
     def get(
         self,
-        includes: Union[
-            Type, List[Type], None
-        ] = AnyFeature,  # ty: ignore[invalid-parameter-default]
+        includes: Union[Type, List[Type], None] = _Unset,
```
TobyBoyne: Could the default here just be `Feature`? This would avoid many of the issues solved in this PR, no?

tests/bofire/test_register.py
```diff
@@ -918,6 +918,54 @@ def n_transformed_inputs(self) -> int:
...
+class _VisibleEngineeredFeature(EngineeredFeature):
+    type: Literal["_VisibleEngineered"] = "_VisibleEngineered"
+    order_id = 102
+
+    @property
+    def n_transformed_inputs(self) -> int:
+        return 1
+
+
+class TestRegisteredFeatureIsVisible:
```
TobyBoyne, line 937: Most of the tests (~80%) in BoFire are functions, not classes. Maybe worth adding to `CLAUDE.md` that functions are preferable over classes, if that is the case?

TobyBoyne (commented):
My only real comment is that `_Unset` can be replaced by `Feature`, which avoids a fair bit of the changes here. I tried making the change and all the tests in `test_features` and `test_register` still pass. It does change the behaviour if `exact==True`, but I'm not sure anyone is setting `exact` without also specifying `includes`.

I think the changes to registering `AnyFeature` are still required for Pydantic.

Otherwise, LGTM :)

bofire/data_models/domain/features.py
jduerholt: Yes — done, thanks, this is much better. `filter_by_class` matches with `isinstance`, so a base class picks up subclasses registered later without any rebinding, which is exactly the property the sentinel was hand-rolling. It also matches the existing idiom: `get_by_objective` already defaults to the `Objective` base class, and `get_reps_df` already calls `self.get(Feature)`.

This removes `_Unset` and `_element_union` entirely — `features.py` goes from +58/−9 down to +14/−10.

On the `exact=True` caveat you flagged: I grepped every `exact` call site in `bofire/` and `tests/` and all of them pass `includes` explicitly (`multi_fidelity.py`, `predictives/botorch.py`, `botorch.py`, `mixed_single_task_gp.py`, and the test modules). So nothing in-tree relies on the old behaviour. It is still technically a public behaviour change — `get(exact=True)` with no `includes` now returns nothing instead of everything — so I called it out in the commit message.

Also agreed the `AnyFeature` registration part is still needed; `Features.features` is annotated with that union and Pydantic rejects the registered type without it.

bofire/data_models/domain/features.py
jduerholt: Moot now — `_element_union` is gone with the `Feature` default, so there is no `get_origin` check left. (For the record you were right: `get_origin(Sequence[X])` is `collections.abc.Sequence`, and my `in (list, tuple, Sequence)` tuple would have missed any other sequence type.)

tests/bofire/test_register.py
jduerholt: Good suggestion — added a **Test Style** section to `CLAUDE.md`.

I counted to check the ratio: **815** module-level `def test_*` vs **152** methods inside test classes, so ~84% functions. Only four files use classes at all: `test_register.py`, `strategies/test_utils.py`, `strategies/test_nchoosek_pruning.py`, `utils/test_domain_repair.py`.

One nuance, which is why I did *not* convert this particular class: `test_register.py` is one of those four, and it is 100% class-based — 11 test classes, zero module-level test functions. So `TestRegisteredFeatureIsVisible` matches the local convention; converting just it would make it the odd one out. I wrote the guideline to say functions are the default, use a class when you need a shared fixture, and follow the local convention when adding to an already class-based file.

I also folded in Johannes' point about test placement (tests for feature containers belong in `data_models/domain/test_features.py`, not `test_register.py`), since that was the other half of the review.

jduerholt:
Hi @TobyBoyne,

many thanks for the review. This was really, really good. Your solution was much better than the one by Claude!

Can you review again? I think it is fine now.

Best,

Johannes

TobyBoyne (approved):
Thanks for making the change! All good now

PR #800, opened 2026-08-15
Add qEUBO preference strategy

### LINKED ISSUE #799: Add qEUBO-based preferential Bayesian optimization
## Motivation

BoFire now provides `PairwiseGPSurrogate`, which can infer a latent utility
function from pairwise comparisons. However, it does not yet provide a predictive
strategy that proposes the next options to compare. Users therefore cannot run an
end-to-end preferential Bayesian optimization loop through BoFire's strategy API.

BoTorch provides `qExpectedUtilityOfBestOption` (qEUBO) for this purpose. qEUBO
jointly selects a batch of options by maximizing the expected utility of the best
option under the preference model posterior.

## Proposal

Add first-class Monte Carlo qEUBO support to BoFire:

- add a serializable `qEUBO` acquisition-function data model, including the
  number of Monte Carlo samples;
- add a preference-aware predictive strategy backed by `PairwiseGPSurrogate`;
- support storing and validating pairwise feedback alongside the referenced
  experiments;
- fit the preference surrogate and use BoTorch's
  `qExpectedUtilityOfBestOption` to generate the next comparison batch;
- integrate candidate generation with BoFire's existing domain transformations,
  constraints, pending candidates, and acquisition optimizer where supported;
- add data-model, mapping, functional, serialization, and end-to-end tests;
- document the preferential BO ask/compare/tell loop.

## Initial scope

- Monte Carlo qEUBO only; analytic EUBO is intentionally out of scope.
- One latent continuous utility output.
- Pairwise feedback represented by `labcode_A`, `labcode_B`, and signed
  `preference`, matching `PairwiseGPSurrogate`.
- The common `q=2` comparison workflow must be supported. The design should not
  unnecessarily prevent larger qEUBO batches.

## Acceptance criteria

- A user can configure and serialize a preference strategy using
  `PairwiseGPSurrogate` and `qEUBO`.
- A user can provide designs plus pairwise comparisons and request the next
  comparison batch.
- Returned candidates satisfy supported domain constraints and preprocessing
  rules.
- Pairwise preference data is validated consistently with
  `PairwiseGPSurrogate`.
- Relevant focused tests and serialization tests pass.

## Open API question

The normal `Strategy.tell` method accepts a single experiments table, while
pairwise learning requires a separate preferences table. The implementation
should introduce a clear preference-feedback API without weakening the existing
`tell` contract.

Jimbo994:
## Summary

- add a serializable Monte Carlo qEUBO acquisition-function configuration
- add a preference strategy backed by `PairwiseGPSurrogate`
- support preference-aware `tell`, joint qEUBO optimization, pending candidates, and latent-utility predictions
- extend the pairwise-GP tutorial and add data-model and end-to-end strategy tests

## Why

This completes the existing pairwise-GP building block into an end-to-end preferential Bayesian optimization loop. The implementation deliberately focuses on Monte Carlo qEUBO and does not add AnalyticEUBO.

Closes #799

## Validation

- `ruff check` on changed Python files
- focused `ty check` on the new strategy modules
- `pytest -q tests/bofire/data_models tests/bofire/surrogates/test_pairwise_gp.py tests/bofire/strategies/test_preference.py tests/bofire/test_register.py` — 1441 passed, 5 skipped
- `quarto render docs/tutorials/advanced_examples/pairwise_gp.qmd` — all 9 executable cells completed


bofire/strategies/predictives/preference.py
```diff
@@ -0,0 +1,331 @@
...
+        missing = set(self.PREFERENCE_COLUMNS) - set(preferences.columns)
+        if missing:
+            raise ValueError(
+                f"`preferences` is missing required columns: {sorted(missing)}."
+            )
+        preferences = preferences[list(self.PREFERENCE_COLUMNS)]
+        preferences["preference"] = pd.to_numeric(
+            preferences["preference"], errors="raise"
+        )
+        if not np.isfinite(preferences["preference"].to_numpy()).all():
```
Jimbo994: do we not have duplicate checks here?

bofire/strategies/predictives/preference.py
```diff
@@ -0,0 +1,331 @@
...
+        new_experiments = self._validate_new_experiments(experiments)
+        if replace or self.experiments is None:
+            combined_experiments = new_experiments.reset_index(drop=True)
+        else:
+            combined_experiments = pd.concat(
+                [self.experiments, new_experiments], ignore_index=True
+            )
+        if len(combined_experiments) == 0:
+            raise ValueError("No preference experiments have been provided.")
+        if combined_experiments["labcode"].duplicated().any():
```
Jimbo994: seems some repeated code with above, should we have a separate validate function?

Jimbo994: we can use validate_preferences?

Jimbo994:
During the previous hackathon, we implemented `PairwiseGP`, but we did not yet explore acquisition functions or a strategy around it. I experimented with this in the current PR. The [`.qmd` tutorial](https://github.com/experimental-design/bofire/blob/agent/qeubo-preference-strategy/docs/tutorials/advanced_examples/pairwise_gp.qmd) is probably the easiest place to see the proposed workflow.

I am not yet certain whether this is how we ultimately want to structure the implementation, because preference BO introduces two related but distinct questions.

First: **which pair should we compare next?**

For this, the PR adds qEUBO. It jointly selects a pair by maximizing the expected utility of the best option in that pair. We could run a complete preference-optimization campaign by repeatedly selecting comparisons with qEUBO, collecting preference labels, and refitting the model. At the end, we could recommend the candidate with the highest posterior utility. This is the approach proposed by [[Astudillo et al.](https://proceedings.mlr.press/v206/astudillo23a.html)](https://proceedings.mlr.press/v206/astudillo23a.html).

Second: **given a fitted `PairwiseGP`, which candidate should we select next?**

Since `PairwiseGP` provides a posterior over latent utility, some acquisition functions from standard BO can also operate on it. This PR currently supports `qLogNEI` for that purpose.

An acquisition function such as qEI requires an explicit `best_f`. Because latent utility is not directly observed, we would need to derive this from the model—for example, by evaluating the posterior mean at the historical designs and taking its maximum. Whether that is the right definition deserves some thought. The qEUBO paper also shows that directly applying qEI in preferential BO can fail to achieve vanishing simple regret.

qLogNEI does not require an explicit `best_f`. Instead, we provide the historical designs as `X_baseline`. It integrates over uncertainty in their latent utilities and compares a new candidate against the sample-wise best baseline candidate.

These two operations could also be combined in one campaign. A standard acquisition function could select the next experiment, after which we generate additional comparisons—randomly when comparisons are inexpensive, or using qEUBO when the comparison budget is limited. This is conceptually similar to [[LILO](https://arxiv.org/abs/2510.17671v2)](https://arxiv.org/abs/2510.17671v2), which alternates standard BO candidate selection with preference labeling and fitting a pairwise GP.

More broadly, since `PairwiseGP` exposes a regular posterior, one could imagine using it within a `SoboStrategy`. A more niche extension would combine a conventionally measured objective and a preference-derived latent objective in a `MoboStrategy`. That would introduce further questions around mixed observation types, separate training data, and normalization of the latent utility.

For now, I have deliberately kept the implementation narrow: `PreferenceStrategy` supports qEUBO for proposing a comparison pair and qLogNEI for proposing a single candidate.

@jduerholt I would especially appreciate thoughts on:

- Does the distinction between collecting preferences and selecting a candidate make sense?
- Should standard acquisition functions remain in `PreferenceStrategy`, or should `PairwiseGP` eventually be supported by the existing SOBO/MOBO strategies?
- How should we define the incumbent for acquisition functions such as qEI that require `best_f`?
- Is the deliberately narrow scope of qEUBO and qLogNEI a reasonable starting point?

bofire/strategies/predictives/preference.py
```diff
@@ -0,0 +1,272 @@
...
+
+    def __init__(self, data_model: DataModel, **kwargs):
+        super().__init__(data_model=data_model, **kwargs)
+        self.acquisition_function = data_model.acquisition_function
+        self.acqf_optimizer: AcquisitionOptimizer = get_optimizer(
+            data_model.acquisition_optimizer
+        )
+        assert data_model.surrogate_spec is not None
+        surrogate = map_surrogate(data_model.surrogate_spec)
+        if not isinstance(surrogate, PairwiseGPSurrogate):
```
jduerholt: Why this, it is impossible to get to this error, using the PreferenceStrategy data model.

PR #802, opened 2026-08-18
Document data model fields in Field(description=...), starting with constraints

jduerholt:
## Why

BoFire is configured entirely through its pydantic data models, which makes those models the API surface for users and for LLM agents alike. That surface was undocumented in the place that matters:

| | before |
|---|---:|
| data model classes | 245 |
| fields declared on them | 465 |
| fields with `Field(description=...)` | **14** |
| classes with a docstring | 123 |

Only `Field(description=...)` reaches `model_json_schema()`, which is what an agent reads. Prose that did exist sat in Google `Attributes:` blocks — invisible to the schema, and free to drift from the fields it describes.

This PR establishes the convention, installs a ratchet so the gap cannot grow, and migrates one package as the pilot. The remaining 14 packages follow as independent PRs.

## The convention

- Field prose goes in `Field(description=...)`, never an `Attributes:` block.
- The class docstring says what the model is and **when to pick it over its siblings in the union** — the part no generated table can supply.
- `Examples:` blocks are encouraged on concrete types, but optional and not executed (no doctest config is added).
- The `type` discriminator is exempt; it is self-documenting.

Written into `CLAUDE.md` (Code Style, a new "Documentation Ratchet" section, and both data-model checklists) and `CONTRIBUTING.md`.

## The ratchet

`tests/bofire/data_models/test_documentation.py` enumerates all **245** models by walking `bofire.data_models` with `pkgutil`, so a new package is covered the moment it exists. Import errors are deliberately uncaught — a module that fails to import would otherwise drop out of the parametrization and leave a whole package silently unenforced.

Two details worth review:

- It checks **own** fields (those in `cls.__annotations__`), not all of `cls.model_fields`. Without this, one undocumented field on `Strategy` fails all 56 strategy subclasses and the allowlist grows from ~450 entries to thousands.
- The two `UNDOCUMENTED_*` sets record what predates the convention and **may only shrink**. `test_field_allowlist_has_no_stale_entries` and its class counterpart fail once an entry is documented (or removed) without being deleted here, so documenting a package and pruning its entries happen in the same commit.

Seeded at **430 fields / 120 classes** — down from 451/122 by the constraints work in this PR. `git diff` on that file is the migration progress bar.

## The pilot: `constraints`

21 classes, 21 fields, **zero allowlist entries left**. 17 descriptions harvested from `Attributes:` blocks (all blocks removed), 4 written new (`Constraint.features`, `Constraint.context`, `InterpointEqualityConstraint.features`, `NonlinearConstraint.hessian_expression`), and docstrings added to `EqualityConstraint` and `InequalityConstraint`.

One rule surfaced that will apply to every later package: **prose for an inherited field is dropped, not moved.** `NChooseKConstraint` and `NonlinearConstraint` each documented `features` — a field owned by `Constraint` — with different wording. Both were dropped in favour of one canonical description on the base, which now propagates into every constraint's schema.

## Verification

- `pytest tests/bofire -k constraint` — 240 passed, 9 skipped (repo-wide, includes the strategy and DoE consumers)
- ruff 0.6.5 (the pinned pre-commit version) — clean
- `ty==0.0.16` — 151 diagnostics before and after, i.e. none introduced
- The ratchet was mutation-tested rather than trusted on a green run: a new class with a bare field fails, a class without a docstring fails, and both stale-entry tests fire when an allowlisted entry is documented without being pruned.

Descriptions do not affect `model_dump()`, so the serialization roundtrip contract in `tests/bofire/data_models/serialization/` is untouched.

## Two things reviewers should know

**Nothing changes on the website yet.** quartodoc drops pydantic fields entirely — it renders `DemoModel()` and only whatever sits in an `Attributes:` block. Making these descriptions visible needs a griffe extension wired into the docs build, which is separate work. Removing the `Attributes:` blocks therefore *reduces* rendered output in the short term; the content is in the JSON schema instead.

**`bofire.data_models.domain.constraints.Constraints`** (the container) is still allowlisted. It lives in the `domain` package and belongs to that PR, not this one.


bofire/data_models/constraints/constraint.py
```diff
@@ -90,6 +108,13 @@ def is_fulfilled(self, experiments: pd.DataFrame, tol: float = 1e-6) -> pd.Serie
 
 
 class InequalityConstraint(IntrapointConstraint):
+    """Abstract base class for constraints that must evaluate to zero or less.
+
+    A candidate fulfills the constraint if the constraint evaluation does not exceed
```
jduerholt: This is wrong, the rhs depends on the actually implementing constraint and the is_fulfilled depends on self.__call__. I would remove it from here. Maybe also applies to EqualityConstraint.

bofire/data_models/constraints/constraint.py
```diff
@@ -80,6 +91,13 @@ class IntrapointConstraint(Constraint):
 
 
 class EqualityConstraint(IntrapointConstraint):
+    """Abstract base class for constraints that must evaluate to zero.
+
+    A candidate fulfills the constraint if the constraint evaluation is within `tol`
```
jduerholt: This is also wrong, have a look at my comment at InequalityConstraint.

jduerholt (commented):
I think there are two errors on the equality and inequality docstrings.

jduerholt:
@bertiqwerty: what do you think about this? Should I integrate the correct rendering of in quarto already in this PR? Or do we want do it when this series of PRs for all classes is done. Motivation is that we move in the direction where agents can compose BoFire and via this change the data models get more self-explanatory.

bofire/data_models/constraints/constraint.py
jduerholt: You're right, fixed in 79b455b2. `rhs` is a field of `LinearConstraint` and `ProductConstraint`, not of these bases — and a `NonlinearEqualityConstraint` has no `rhs` at all, so my "budget that must not be overspent" example described a subclass this class knows nothing about.

Rather than dropping the docstring entirely (the new ratchet test requires one), I narrowed it to what this class actually fixes — the interpretation of the value returned by `__call__`:

> Abstract base class for constraints fulfilled at a non-positive constraint value.
>
> What is evaluated is defined by the `__call__` of the implementing subclass. This class only fixes how that value is interpreted: a candidate fulfills the constraint if the value does not exceed `tol`.

The "when to pick this over its sibling" guidance now lives only on the concrete classes (`LinearEqualityConstraint` vs `LinearInequalityConstraint` etc.), where the choice is actually made.

bofire/data_models/constraints/constraint.py
jduerholt: Same fix applied here in 79b455b2 — it did apply to `EqualityConstraint` too. It now reads:

> Abstract base class for constraints fulfilled at a constraint value of zero.
>
> What is evaluated is defined by the `__call__` of the implementing subclass. This class only fixes how that value is interpreted: a candidate fulfills the constraint if the value is within `tol` of zero.

bertiqwerty (approved):
Thanks, Johannes.

PR #803, opened 2026-08-20
Document the objectives and priors data models

jduerholt:
Second package batch of the migration started in #802. Drains the `objectives` and `priors` allowlists to zero — **49 field descriptions and one class docstring**, taking the ratchet from 551 to 501 remaining entries.

The two packages go together because they sit next to each other alphabetically in the allowlist; splitting them would have produced a merge conflict between the two PRs for no reviewer benefit.

## Mostly harvest, three judgment calls

41 of the 49 descriptions came from existing `Attributes:` blocks, which are now removed. The rest needed more than a move:

**`LKJPrior`'s docstring was stale.** It documented `n` and `eta`; the fields have been `n_tasks` and `shape` since the gpytorch mapping was written (`bofire/priors/mapper.py:83-87` maps `n_tasks`→`n`, `shape`→`eta`). The prose was recovered under the current names rather than discarded — worth a look to confirm I mapped them the way you'd expect.

**`DimensionalityScaledLogNormalPrior` had no prose anywhere.** Its semantics came from `mapper.py:97-107`, and the effective-parameter formula is now stated in the class docstring the same way its Gamma sibling already states its own:

```
loc_eff   = loc + log(d) * loc_scaling
scale_eff = sqrt(scale**2 + log(d) * scale_scaling)
```

**Several subclasses documented fields they don't declare** — `clip`, `w` and `bounds` throughout `desirabilities.py`, and all three interval fields in `NonTransformedInterval` / `LogTransformedInterval`. Per the rule established in #802 those entries were dropped, with one canonical description on the declaring class. Where a subclass genuinely *reinterprets* an inherited field, that goes in its class docstring instead: `MovingMaximizeSigmoidObjective` treats `tp` as a relative turning point, which the shared `SigmoidObjective.tp` description cannot say.

## Verification

- `pytest tests/bofire -k "objective or prior or desirab"` — 382 passed, 32 skipped
- ruff 0.6.5 clean; `ty==0.0.16` at 151 diagnostics, unchanged from main
- Zero `objectives` / `priors` entries left in either allowlist

No CHANGELOG entry: #802 announced the migration, and a per-package entry would conflict on every subsequent PR.

As with #802, none of this is visible on the website yet — quartodoc drops pydantic fields until the rendering work lands. The payoff is in `model_json_schema()`.


bofire/data_models/objectives/desirabilities.py
```diff
@@ -112,21 +117,20 @@ class DecreasingDesirabilityObjective(DesirabilityObjective):
...
+    `bounds[1]` it is 0 (or negative). Use `IncreasingDesirabilityObjective` for the
+    opposite direction.
+
+    Examples:
+        >>> DecreasingDesirabilityObjective(bounds=[0.0, 10.0])
     """
 
     type: Literal["DecreasingDesirabilityObjective"] = "DecreasingDesirabilityObjective"
-    log_shape_factor: float = 0.0
+    log_shape_factor: float = Field(
```
jduerholt, line 132: Can we not get this by desirability?

bofire/data_models/objectives/desirabilities.py
jduerholt, line 1: We should somehow mentioned that these desirbailities in here are not differentiable ...

bofire/data_models/objectives/identity.py
```diff
@@ -11,16 +11,18 @@
...
-    Attributes:
-        w (float): float between zero and one for weighting the objective
-        bounds (Tuple[float], optional): Bound for normalizing the objective between zero and one. Defaults to (0,1).
-
     """
 
     type: Literal["IdentityObjective"] = "IdentityObjective"
-    w: TWeight = 1
-    bounds: Bounds = [0, 1]
+    w: TWeight = Field(
```
jduerholt: Maybe we should also mention that the weight is only used in specific strategies like AdditiveSobo, MultiplicativeSobo and ignored in others.

bofire/data_models/objectives/identity.py
```diff
@@ -11,16 +11,18 @@
...
+        default=1,
+        description="Relative weight of this objective, between zero and one.",
+    )
+    bounds: Bounds = Field(
+        default=[0, 1],
+        description="Lower and upper bound used to normalize the objective between "
```
jduerholt: And Zero and one means no normalization.

bofire/data_models/objectives/identity.py
```diff
@@ -74,10 +76,12 @@ def __call__(
 class MaximizeObjective(IdentityObjective):
     """Child class from the identity function without modifications, since the parent class is already defined as maximization
```
jduerholt: Does it makes sense to call it here child class, info is irrelevant for the agent or, only interesting for coders, and they just see it. We should look for this kind of descripotionsl

bofire/data_models/objectives/identity.py
```diff
@@ -74,10 +76,12 @@ def __call__(
...
 
-    Attributes:
-        w (float): float between zero and one for weighting the objective
-        bounds (Tuple[float], optional): Bound for normalizing the objective between zero and one. Defaults to (0,1).
+    The default objective for an output that should be as large as possible. Use
+    `MinimizeObjective` for the opposite, or `CloseToTargetObjective` to aim at a
+    specific value.
 
+    Examples:
+        >>> MaximizeObjective(bounds=[0.0, 100.0])
```
jduerholt: We should not have the bounds in here, these bounds are some kind of special case not often needed. Most cases, bounds should be just kept as default.

jduerholt: Applies everywhere.

bofire/data_models/objectives/identity.py
```diff
@@ -89,10 +93,11 @@ def to_description(self) -> str:
...
     """Class returning the negative identity as reward.
 
-    Attributes:
-        w (float): float between zero and one for weighting the objective
-        bounds (Tuple[float], optional): Bound for normalizing the objective between zero and one. Defaults to (0,1).
+    For an output that should be as small as possible. Use `MaximizeObjective` for the
+    opposite, or `CloseToTargetObjective` to aim at a specific value.
 
+    Examples:
+        >>> MinimizeObjective(bounds=[0.0, 100.0])
```
jduerholt: No bounds in example.

bofire/data_models/objectives/sigmoid.py
```diff
@@ -12,28 +13,28 @@
...
+    )
+    tp: float = Field(description="Turning point of the sigmoid.")
+    w: TWeight = Field(
+        default=1,
+        description="Relative weight of this objective, between zero and one.",
+    )
 
 
 class MaximizeSigmoidObjective(SigmoidObjective):
     """Class for a maximizing sigmoid objective
```
jduerholt: Also not really descriptive, maybe we should also explain what a sigmoid objective is, how it is behaved and what is desriable ...

jduerholt: This is in principle, maximize to a threshhold, and above the thresh it does not count anymore ...

bofire/data_models/objectives/sigmoid.py
```diff
@@ -110,11 +112,11 @@ def __call__(
 class MinimizeSigmoidObjective(SigmoidObjective):
     """Class for a minimizing a sigmoid objective
```
jduerholt: Hmm, also not sure about the description here.

jduerholt: should maybe also mention that often the strategies use the sigmoid objectives as output constraints?

bofire/data_models/objectives/target.py
```diff
@@ -42,19 +49,27 @@ def __call__(
...
+    Examples:
+        >>> TargetObjective(target_value=7.0, tolerance=0.5, steepness=10.0)
     """
 
     type: Literal["TargetObjective"] = "TargetObjective"
-    w: TWeight = 1
-    target_value: float
-    tolerance: TGe0
-    steepness: TGt0
+    w: TWeight = Field(
```
jduerholt: can we not get it by inheritance? or already get it?

bofire/data_models/objectives/target.py
```diff
@@ -42,19 +49,27 @@ def __call__(
...
+        default=1,
+        description="Relative weight of this objective, between zero and one.",
+    )
+    target_value: float = Field(description="Target value that should be reached.")
+    tolerance: TGe0 = Field(
+        description="Half-width of the accepted window around the target value.",
+    )
+    steepness: TGt0 = Field(
```
jduerholt, line 60: Do we not also get thsi via inheritance?

bofire/data_models/priors/constraint.py
```diff
@@ -14,31 +14,36 @@ class PriorConstraint(BaseModel):
 class Positive(PriorConstraint):
     """Class for constraints that enforce a prior to be positive.
```
jduerholt: is this nice prose?

bofire/data_models/priors/constraint.py
```diff
@@ -14,31 +14,36 @@ class PriorConstraint(BaseModel):
...
+    Use when only the sign matters; use `GreaterThan` or `Interval` to bound the
+    parameter more tightly.
     """
 
     type: Literal["Positive"] = "Positive"
-    initial_value: Optional[PositiveFloat] = None
+    initial_value: Optional[PositiveFloat] = Field(
+        default=None,
+        description="Optional warm-start value used when registering the constraint "
+        "on a gpytorch parameter. If not provided, the consuming code may supply a "
```
jduerholt: In principle nice, but it does not make sense to mention gpytorch here, or?

bofire/data_models/priors/constraint.py
```diff
@@ -14,31 +14,36 @@ class PriorConstraint(BaseModel):
...
+    Use `Interval` instead if the parameter needs an upper bound as well.
     """
 
     type: Literal["GreaterThan"] = "GreaterThan"
-    lower_bound: float
-    initial_value: Optional[PositiveFloat] = None
+    lower_bound: float = Field(
+        description="Lower bound enforced on the constrained parameter.",
+    )
+    initial_value: Optional[PositiveFloat] = Field(
```
jduerholt: can we get this by inheritance or change the code so that we get by inheritance?

bofire/data_models/priors/constraint.py
```diff
@@ -52,17 +57,20 @@ def validate_initial_value(self):
...
+    Use `Interval` instead if the parameter needs a lower bound as well.
     """
 
     type: Literal["LessThan"] = "LessThan"
-    upper_bound: float
-    initial_value: Optional[PositiveFloat] = None
+    upper_bound: float = Field(
+        description="Upper bound enforced on the constrained parameter.",
+    )
+    initial_value: Optional[PositiveFloat] = Field(
```
jduerholt: Inheritance.

bofire/data_models/priors/gamma.py
```diff
@@ -36,17 +38,27 @@ class DimensionalityScaledGammaPrior(Prior):
...
     ``DIMENSIONALITY_SCALED_THREESIX_LENGTHSCALE_PRIOR``).
 
-    Attributes:
-        concentration(PositiveFloat): base concentration of the gamma distribution.
-        concentration_scaling(float): factor multiplying ``sqrt(d)`` that is added to
-            the base concentration.
-        rate(PositiveFloat): base rate of the gamma distribution.
-        rate_power(float): exponent of ``d`` that the base rate is multiplied by.
-
+    Use instead of `GammaPrior` when the same prior should be reusable across problems
```
jduerholt: Remove this sentence.

bofire/data_models/priors/interval.py
```diff
@@ -10,19 +10,19 @@ class Interval(PriorConstraint):
 
     It is used to define interval constraints on GP hyperparameters.
```
jduerholt: A bit doubled, or?

bofire/data_models/priors/interval.py
```diff
@@ -10,19 +10,19 @@ class Interval(PriorConstraint):
...
+    `Positive`.
     """
 
     type: Literal["Interval"] = "Interval"
-    lower_bound: PositiveFloat
-    upper_bound: PositiveFloat
-    initial_value: Optional[PositiveFloat] = None
+    lower_bound: PositiveFloat = Field(description="Lower bound of the interval.")
+    upper_bound: PositiveFloat = Field(description="Upper bound of the interval.")
+    initial_value: Optional[PositiveFloat] = Field(
```
jduerholt: Can we not have this by inheritance? In prinviple it is the same as for the others and it is quite obvious that it should lie in between.

bofire/data_models/priors/normal.py
```diff
@@ -1,49 +1,76 @@
...
     It was introduced by Hvarfner et al. in their paper https://arxiv.org/abs/2402.02229. More can be read in
     this excellent blogpost: https://www.miguelgondu.com/blogposts/2024-03-16/when-does-vanilla-gpr-fail/
+
+    The effective log-normal distribution used at mapping time, given the
+    dimensionality ``d``, is::
+
+        loc_eff   = loc + log(d) * loc_scaling
+        scale_eff = sqrt(scale**2 + log(d) * scale_scaling)
+
+    Use instead of `LogNormalPrior` when the same prior should be reusable across
```
jduerholt: Remove this sentence starting with `use instead`.

jduerholt (commented):
I added some comments, mainly focusing on inheritance and some other wordings, maybe one can also investigate if one can use more inheritance in general for the things here.

jduerholt:
Pushed as 253c9545. Summary of what changed, and the two places I did *not* follow the suggestion.

## Inheritance

I mapped which fields are actually shared before moving anything:

| field | declared on | on every subclass? | action |
|---|---|---|---|
| `w` | 5 objective classes | **yes**, all 12 objectives | moved to `Objective` |
| `initial_value` | 4 prior constraints | **yes**, all 6 | moved to `PriorConstraint` |
| `target_value` | TargetObjective, CloseToTargetObjective | no — only those two | left duplicated |
| `log_shape_factor` | Increasing, Decreasing, Peak | **no** — InRangeDesirability lacks it | left duplicated |
| `lower_bound` / `upper_bound` | Interval, GreaterThan, LessThan | no — asymmetric by design | left as is |

The first two are pure deduplication: no schema change, no spec updates, nine redeclarations gone.

This turned out to matter more than tidiness. Pydantic **silently drops the inherited description when a subclass redeclares a field** — `Redeclares.model_fields["w"].description` is `None` — so every duplicate declaration needed its own copy of the same prose, and any of them could drift. Hoisting fixes the cause rather than the symptom.

**On `log_shape_factor` (your comment at desirabilities.py:129):** I left it. `InRangeDesirability` has no shape factor, so putting it on `DesirabilityObjective` would give that class a field with no effect on its output — a schema change plus spec updates. The alternative is a new intermediate base for the three shaped variants. Happy to do either if you'd prefer; it just seemed like more hierarchy than the duplication costs.

**On `target_value` (target.py:69):** same reasoning — only two classes share it, and their only common ancestor is `Objective`, so it would need a new intermediate class.

## Wording

Docstrings now lead with behaviour. The "use instead of X" pointers are gone everywhere except between `TargetObjective` and `CloseToTargetObjective`, which are genuinely easy to confuse and where the names don't reveal that one defines a feasible window and the other doesn't.

Specifically: the sigmoid objectives now explain that they express a *threshold* — reward saturates above `tp`, so no further credit — and that strategies commonly use them as output constraints; the desirability objectives note they aren't differentiable at the bounds; `bounds` records that the `[0, 1]` default leaves values unscaled; `w` records that only the additive, multiplicative and custom SOBO strategies read it (verified against `torch_tools.py:637,650,692,761`); the gpytorch reference is out of `initial_value`; the "child class" framing and similar source-reader notes are gone; and examples no longer set `bounds`.

**One thing to check:** you said no bounds in examples, applies everywhere — so I removed it from `InRangeDesirability` too, leaving `>>> InRangeDesirability()`. That's the one class where the range *is* the entire semantic content, so the example is now fairly empty. Say the word and I'll put it back there alone.

`CLAUDE.md` and `CONTRIBUTING.md` are updated so later packages follow the revised convention, including the rule about declaring shared fields once. Per your call, the constraints package from #802 keeps its existing comparative sentences.

## Verification

- `pytest tests/bofire -k "objective or prior or desirab"` — 382 passed, 32 skipped
- ruff 0.6.5 clean; `ty==0.0.16` at 151 diagnostics, unchanged from main
- Confirmed the descriptions arrive by inheritance: `MaximizeObjective.w` and `TargetObjective.w` both resolve to the base description, likewise `Interval.initial_value` and `GreaterThan.initial_value`

Jimbo994 (approved):
Very nice. this makes our documentation cleaner, but should be very nice for LLM development and optimization!

Issue #804, opened 2026-08-21
[FEATURE REQUEST]: Progressive-disclosure digest for data model JSON schemas


jduerholt:
## Motivation

BoFire is configured entirely through its pydantic data models, which makes `model_json_schema()` the API surface an LLM agent reads. For leaf types that works well. For strategies it does not:

| model | schema size | `$defs` | own fields |
|---|---:|---:|---:|
| `TargetObjective` | 1.4 KB / ~350 tokens | 0 | 5 |
| `LinearEqualityConstraint` | 1.6 KB / ~410 tokens | 0 | 5 |
| **`SoboStrategy`** | **189 KB / ~47,000 tokens** | **120** | **10** |

The bloat is transitive rather than intrinsic. `SoboStrategy` has ten knobs of its own, but `domain` drags in every feature, constraint and objective type, and `surrogate_specs` drags in every surrogate, kernel and prior:

```
Kernel 18 · Surrogate 14 · Objective 12 · Constraint 11 · Prior 7 · Input 6 · other 35
```

So the schema is complete and correct and unusable as a prompt: 47k tokens to learn that a strategy has ten settings. It cannot simply be truncated, because the nesting *is* the information.

Note this gets worse, not better, as #802 and its follow-ups add `Field(description=...)` across the data models — `SoboStrategy`'s schema grew from 182 KB to 189 KB as the `objectives` and `priors` descriptions landed. The descriptions are worth having; they just make a navigation aid more necessary.

## Describe the solution you'd like to see implemented in BoFire

Progressive disclosure over the schema graph — roughly `ls` / `cd` / `cat` — rather than compression:

```python
digest(SoboStrategy)                          # ~139 tokens: the ten knobs
digest(SoboStrategy, "acquisition_function")  # ~93 tokens: what can go here
qUCB.model_json_schema()                      # ~200 tokens: the chosen leaf
```

Three round trips totalling roughly 430 tokens replace the 47,000-token schema, and the caller only pays for the branch it takes.

I prototyped both levels. Level 1:

```
# SoboStrategy
  domain*: Domain
  seed: integer | null = None
  convergence_criterion: one of (3): HypervolumeImprovementCriterion, ObjectiveImprovementCriterion, ProposalDeviationCriterion
  acquisition_optimizer: BotorchOptimizer | GeneticAlgorithmOptimizer
  surrogate_specs: BotorchSurrogates
  frequency_hyperopt: integer = 0
  folds: integer = 5
[...]

PR #806, opened 2026-08-23
Document the features data models

jduerholt:
Third package batch of the migration started in #802 and continued in #803. Drains the `features` allowlist to zero — **41 field descriptions and one class docstring** (`CategoricalOutput`), taking the ratchet from 501 to 459 remaining entries.

## A counting correction worth knowing for the remaining packages

30 of the 41 were harvestable, not the 15 I estimated when picking this package. `engineered_feature.py` documents its fields under Google **`Args:`** blocks rather than `Attributes:`, which an `Attributes:`-only scan misses entirely. Both block types are harvested and removed here, and the scan for the remaining packages should look for both.

## Inheritance: nothing hoisted, and that is the finding

#803 moved `w` to `Objective` and `initial_value` to `PriorConstraint`. This package is the opposite case — I checked every redeclaration and all of them are deliberate type changes that must stay:

| field | why it stays duplicated |
|---|---|
| `unit` | on the numerical inputs and `ContinuousOutput`, but not on the categoricals or any engineered feature — hoisting to `Feature` would add it to nine classes that lack it |
| `objective` | on both outputs and nowhere else, but required `AnyCategoricalObjective` vs `Optional[AnyObjective]` |
| `descriptors` | narrowed to `None` on both task inputs on purpose; the descriptions now say *why* it is disabled rather than restating the type |
| `features` | `ProductFeature` widens it (plain `List[str]`, dropping the uniqueness validator in `FeatureKeys` so `['x', 'x']` gives a quadratic term) while `CloneFeature` narrows it to `OneFeatureKeys` |

## Three docstring defects fixed rather than harvested

- **`CloneFeature`'s `Args:` block was copy-pasted from `ProductFeature`** — it said the features are "used to compute the product" and that repeating one is allowed. Neither is true: `CloneFeature` copies them, and its `OneFeatureKeys` type forbids duplicates. Harvesting mechanically would have published that into the schema.
- **`Output` documented `key`**, which `Feature` declares.
- **`SumFeature` and `MeanFeature` documented only inherited fields** (`features`, `keep_features`).

## Rename candidates

As agreed, collected rather than acted on — fields whose description exists mainly because the name does not carry it:

- **`DiscreteInput.rtol`** — relative tolerance for matching a provided value against `values`. The name says "some tolerance"; nothing indicates what it is compared against.
- **`ContinuousInput.allow_zero`** — makes the feature semi-continuous. The name reads like a validation flag rather than a statement about the feature's domain; `semicontinuous` would carry it.
- **`InterpolateFeature.normalize_y`** — a *divisor*, not a boolean, while its neighbour `normalize_x` is a boolean. The asymmetry is invisible from the names.

All three are breaking schema changes, so nothing was renamed here.

## Verification

- `pytest tests/bofire -k "feature or input or output"` — 668 passed, 5 skipped
- ruff 0.6.5 clean; `ty==0.0.16` at 151 diagnostics, unchanged from main
- Confirmed `Feature.key` and `Feature.context` reach subclass schemas by inheritance, e.g. via `ContinuousInput.model_json_schema()`
- No `Attributes:` or `Args:` field blocks left in the package

No CHANGELOG entry: #802 announced the migration.


bofire/data_models/features/_descriptor_spec.py
```diff
@@ -60,23 +60,32 @@ def filter_correlated(df: pd.DataFrame, cutoff: float) -> pd.DataFrame:
...
+    """Mixin declaring how to build a descriptor table for a feature."""
+
+    columns: Optional[List[str]] = Field(
+        default=None,
+        description="Which of the feature's stored numeric descriptor columns to use. "
+        "Null means all of them; an empty list means none, leaving only the generated "
+        "columns.",
+    )
+    generators: List[AnyDescriptorGenerator] = Field(
+        default_factory=list,
```
jduerholt: Why not just `=[]` instead of using `default_factory`?

bofire/data_models/features/categorical.py
```diff
@@ -73,26 +73,32 @@ class CategoricalInput(Input):
...
-        categories (List[str]): Names of the categories.
-        allowed (List[bool]): List of bools indicating if a category is allowed within the optimization.
-        descriptors (Descriptors, optional): Per-category descriptor data — numeric
-            columns and/or a SMILES structure column. Defaults to None.
-        key (str, inherited from `Feature`): The unique name of the feature.
-        context (str, optional, inherited from `Feature`): Free-text context for the
-            feature. Defaults to None.
-
+    Examples:
+        >>> CategoricalInput(key="solvent", categories=["water", "ethanol", "thf"])
```
jduerholt, line 68: Please update the example also with descriptors.

bofire/data_models/features/categorical.py
```diff
@@ -73,26 +73,32 @@ class CategoricalInput(Input):
...
     # order_id: ClassVar[int] = 5
     order_id: ClassVar[int] = 7
 
-    categories: CategoryVals
-    descriptors: Optional[Descriptors] = None
+    categories: CategoryVals = Field(
+        description="Names of the categories the feature can take. They are unordered; "
+        "use a `DiscreteInput` when the numeric order of the values is meaningful.",
+    )
+    descriptors: Optional[Descriptors] = Field(
```
jduerholt, line 91: Why do we list discriptors here? Is it not from inheritance?

bofire/data_models/features/continuous.py
```diff
@@ -30,36 +30,37 @@ class ContinuousInput(NumericalInput):
...
-            Consumed by N-arity engineered features, not by this feature's own encoding.
-            Defaults to None.
-        unit (str, optional, inherited from `NumericalInput`): The unit of the feature.
-            Defaults to None.
-        key (str, inherited from `Feature`): The unique name of the feature.
-        context (str, optional, inherited from `Feature`): Free-text context for the
-            feature. Defaults to None.
-
+    Examples:
+        >>> ContinuousInput(key="temperature", bounds=(20, 80))
```
jduerholt, line 25: Why do you regenerate an example and not just use the one from above?

bofire/data_models/features/descriptors.py
jduerholt: Is the correct way of writing examples?

jduerholt: Correct way of writing examples?

bofire/data_models/features/discrete.py
```diff
@@ -18,24 +18,24 @@ class DiscreteInput(NumericalInput):
...
-        descriptors (Descriptors, optional): Descriptor data for the single component —
-            numeric columns holding one value each, and/or a one-element SMILES structure.
-            Defaults to None.
-        unit (str, optional, inherited from `NumericalInput`): The unit of the feature.
-            Defaults to None.
-        context (str, optional, inherited from `Feature`): Free-text context for the
-            feature. Defaults to None.
-
+    Examples:
+        >>> DiscreteInput(key="n_layers", values=[1.0, 2.0, 4.0, 8.0])
```
jduerholt, line 22: Add a second example with the descritptors.

bofire/data_models/features/engineered_feature.py
```diff
@@ -172,33 +175,62 @@ class InterpolateFeature(EngineeredFeature):
...
-        y_keys: Feature keys used as y-coordinates for interpolation.
-        interpolation_range: (lower, upper) bounds for the interpolation x-grid.
-        n_interpolation_points: Number of evenly spaced points in the interpolation grid.
-        prepend_x: Extra x-values to prepend before the feature x-values.
-        append_x: Extra x-values to append after the feature x-values.
-        prepend_y: Extra y-values to prepend before the feature y-values.
-        append_y: Extra y-values to append after the feature y-values.
-        normalize_y: Divisor for y-values before interpolation.
-        normalize_x: Whether to normalize x-values to [0, 1] before interpolation.
+    The optimized inputs are the coordinates of a curve; this feature evaluates that
```
jduerholt: Change to "this feature interpolates that curve" and "which is how profiles such as a temperature ramp can be optimized".

bofire/data_models/features/engineered_feature.py
```diff
@@ -229,18 +261,15 @@ class CloneFeature(EngineeredFeature):
...
-            It is allowed to state a feature more than once to for example
-            an quadratic term.
-        keep_features: Whether to keep the original features after
-            creating the engineered feature in surrogate creation.
     """
 
     type: Literal["CloneFeature"] = "CloneFeature"
     order_id: ClassVar[int] = 5
-    features: OneFeatureKeys
+    features: OneFeatureKeys = Field(
```
jduerholt, line 338: Do we need this OneFeatureKeys? Or can we just get it by inheritance?

bofire/data_models/features/task.py
```diff
@@ -34,8 +34,19 @@ class CategoricalTaskInput(TaskInput, CategoricalInput):
...
-    fidelities: list[int] = []
+    descriptors: None = Field(
+        default=None,
+        description="Always null. A task input carries no descriptor data: the "
+        "relationship between tasks is learned as the inter-task covariance of the "
+        "surrogate, not read off descriptor columns.",
+    )
+    fidelities: list[int] = Field(
+        default=[],
+        description="Fidelity level of each task, one entry per category in the same "
```
jduerholt, line 55: This is wrong, the higher the better the more expensive.

jduerholt (commented):
A few things to adapt and some questions.

jduerholt:
Pushed as 8bae702a. Four of your ten comments were questions, so here are the answers alongside what changed.

## Answered by measuring

**`default_factory` vs `= []`** — your suggestion is better than a style preference:

```
Field(default_factory=list)  → 'default' in schema: False
Field(default=[])            → 'default' in schema: True   []
```

Pydantic deep-copies defaults, so the literal form is safe from the mutable-default trap. And `default_factory` hiding the default from `model_json_schema()` is exactly the gap recorded in #804 — an agent reading the schema cannot see it at all. Changed here and in `InterpolateFeature`'s four `prepend_*`/`append_*` fields; `prepend_x` and `generators` now report `[]` to a schema reader. Rule added to `CLAUDE.md` and `CONTRIBUTING.md`.

**Fidelity direction (`task.py:45`)** — I think your review comment had it backwards, and I'd rather flag that than quietly write the reverse into a published schema. The code says fidelity **0 is the target**, the accurate and expensive task:

- `multi_fidelity.py:63` — `task_feature.allowed = [fidelity == 0 for ...]`, under the comment *"only optimize the input x on the target fidelity"*
- `multi_fidelity.py:91-93` — `argsort(...)[::-1]` then `[-1]`, so the **smallest** value is the target
- `docs/tutorials/advanced_examples/multifidelity_bo.qmd:80` and `tests/.../specs/strategies.py:1021` both use `categories=["task_hf", "task_lf"], fidelities=[0, 1]`

You picked "you're right, I misread it" in my questionnaire, though the option text underneath said the opposite, so I went with the label plus the evidence. The description now spells the inversion out — *"the numbering runs opposite to the everyday sense of the word: 0 is the target task, the accurate and expensive one"* — since that is precisely what tripped us both up. Happy to flip it if you disagree; it is one field.

## Examples

Converted every `::` literal block in the package to Google `Examples:` with `>>>`, including the four `Example::` blocks in method docstrings, so the package is consistent. The descriptor illustrations that were floating in the class prose of `ContinuousInput`, `CategoricalInput` and `Descriptors` now live in those sections rather than being restated — that answers your "why regenerate an example" point directly. `DiscreteInput` gained the descriptor example it was missing.

The convention is now written into `CLAUDE.md` and `CONTRIBUTING.md`, with the reason: the docstring is copied verbatim into the schema `description`, so a bare `::` ships to agents as a stray artifact while `>>>` reads as plain text.

## Left as they are, per your answers

- **`descriptors` on `CategoricalInput`** — it could be hoisted to `Input` (every `AnyInput` member has the field), but the semantics differ: one row per category there, exactly one row on `NumericalInput`. A shared description would have to be vague about the row count, which is the thing users get wrong.
- **`CloneFeature.features: OneFeatureKeys`** — needed because `EngineeredFeature.features` is `min_length=2`. Inheriting would mean relaxing the base to 1, which would let `SumFeature`, `MeanFeature`, `WeightedSumFeature` and `InterpolateFeature` accept a single feature.

Plus the `InterpolateFeature` wording change you asked for.

## Verification

- `pytest tests/bofire -k "feature or input or output or encoding"` — 730 passed, 5 skipped (widened to cover `encodings`, since `DescriptorEncoding` inherits the changed `DescriptorSpec` defaults)
- ruff 0.6.5 clean; `ty==0.0.16` at 151 diagnostics, unchanged from main

jduerholt:
Hi @TobyBoyne,

can you review this? 

Best,

Johannes

bofire/data_models/features/categorical.py
```diff
@@ -57,42 +57,51 @@ class CategoricalInput(Input):
...
+        default=None,
+        description="Per-category descriptor data: numeric columns holding one value "
+        "per category and/or a SMILES structure per category, in the same order as "
+        "`categories`. How it becomes model columns is chosen per surrogate via "
+        "`categorical_encodings`, not here.",
+    )
     allowed: Optional[Annotated[List[bool], Field(min_length=2)]] = Field(
         default=None,
         validate_default=True,
```
TobyBoyne, line 100: Not necessary to have `validate_default=True` for `default=None`, I think

bofire/data_models/features/continuous.py
```diff
@@ -314,26 +317,30 @@ def __str__(self) -> str:
...
+    unit: Optional[str] = Field(
+        default=None,
+        description="Unit the output is measured in, for example 'g/l'. Recorded for "
+        "documentation; it is not used in any computation.",
+    )
 
     objective: Optional[AnyObjective] = Field(
         default_factory=lambda: MaximizeObjective(w=1.0),
+        description="How this output should be optimized. Defaults to maximization. "
+        "Set to null for an output that is modelled and reported but not optimized.",
```
TobyBoyne: Should this be `"Set to None"`? Or is this for JSON?

bofire/data_models/features/descriptors.py
```diff
@@ -52,38 +52,36 @@ class Descriptors(BaseModel):
...
-                columns={"logP": [-1.4, -0.3, 0.5], "MW": [18.0, 46.0, 72.0]},
-                structure=["O", "CCO", "C1CCOC1"],
-            ),
-        )
+    block is attached to, and is the feature's business, not this class's. A
+    ``CategoricalInput`` has **one level per category**, so the block picks the row of
+    the chosen category. A ``ContinuousInput`` or ``DiscreteInput`` has **a single
+    level**, the feature itself: it is one component of a mixture whose amount weights
+    its row, so each column holds one value. For a discrete input the allowed ``values``
+    are not levels — a restricted amount of a substance still describes one substance.
```
TobyBoyne: This is a bit too verbose, and couples the docstring with the implementation of the `Input`s. I think you should just let the examples explain it. I would maybe rewrite this paragraph as:

> A "level" is a row of the block. Each level corresponds provides the descriptor data for a unique input (e.g., each unique molecule will have its own level). See Examples for usage with different features.

bofire/data_models/features/engineered_feature.py
```diff
@@ -172,33 +211,75 @@ class InterpolateFeature(EngineeredFeature):
...
-        prepend_x: Extra x-values to prepend before the feature x-values.
-        append_x: Extra x-values to append after the feature x-values.
-        prepend_y: Extra y-values to prepend before the feature y-values.
-        append_y: Extra y-values to append after the feature y-values.
-        normalize_y: Divisor for y-values before interpolation.
-        normalize_x: Whether to normalize x-values to [0, 1] before interpolation.
+    The optimized inputs are the coordinates of a curve; this feature interpolates that
+    curve on a fixed grid and appends one column per grid point. That gives the model a
+    consistent representation of the curve even though the coordinates move, which is
+    how profiles such as a temperature ramp can be optimized.
```
TobyBoyne: Maybe this is just me, but I cannot work out how this feature works from this docstring.
- What are "the optimized inputs"?
- "Appends one column" to what? If I set `n_interpolation_points=1000`, will the dimension of my domain increase by 1000? I would have thought no, and that it would just improve the quality of the interpolation, but the docstring implies otherwise.
- Are `y_keys` output features? 
- I can't make much sense of the example. If `T1` is an input. then the Domain already includes temperature so no need to interpolate it. If `T1` is an output, then do I need to wire up more machinery so that this feature directly gives the output?

It is very possible I'm just missing something completely here. But I think this paragraph needs to be rewritten, and the example needs definition of the features.

bofire/data_models/features/engineered_feature.py
```diff
@@ -230,17 +311,17 @@ class CloneFeature(EngineeredFeature):
...
     """
 
     type: Literal["CloneFeature"] = "CloneFeature"
     order_id: ClassVar[int] = 5
-    features: OneFeatureKeys
+    features: OneFeatureKeys = Field(
+        description="Keys of the input features to copy. Each appears once; the copy "
+        "gets its own column so it can be scaled or fed to a different kernel than the "
+        "original.",
+    )
```
TobyBoyne, line 341: Is cloning multiple features in a single `CloneFeature` possible? If no, the descriptions need to be rewritten.

If yes, how does this work with more than one feature? If `CloneFeature(key=temperature_copy", features=["temperature", "time"])`, then how would I access the copy of `time`? Or would I only be able to access the *pair* of cloned features, and not access either one of them independently? The example should probably include an example with cloning two features.

Also "OneFeatureKeys" implies, to me, that `len(features)==1`, but from its definition it looks like `len(features)>=1`. Maybe worth renaming?

bofire/data_models/features/task.py
```diff
@@ -30,12 +30,35 @@ class CategoricalTaskInput(TaskInput, CategoricalInput):
...
+        "order. Note the numbering runs opposite to the everyday sense of the word: 0 "
+        "is the target task, the accurate and expensive one that is being optimized "
+        "for, and larger values are progressively cheaper, less accurate "
+        "approximations. Must start at 0 and increase in steps of one. Defaults to "
+        "every task being at fidelity 0.",
```
TobyBoyne: (Nit) This reads slightly unclear, since there is seemingly a contradiction between
> Must start at 0 and increase in steps of one

and 

> Defaults to every task being at fidelity 0

The first quote, to me, suggests that each is unique, which is not necessarily the case. The `levels` increase in steps of one, but each level can be assigned to multiple tasks.

Also I don't like the phrasing of "Note the numbering runs opposite to the everyday sense of the word", I would just remove that bit.

TobyBoyne (commented):
Always nice to see some work on documentation :)

Please have a look at my comments on the Engineered Features. Otherwise, LGTM!

bofire/data_models/features/continuous.py
jduerholt: Yeah, I was also stubling about this. Honestly this was AI generated under my guidance (I want to properly document the data models). And thhis also propagates to json schemas. But I think None is better here, because it is python code docimentation.

bofire/data_models/features/engineered_feature.py
jduerholt: Yeah, this is a very hard to explain super edgy feature. I would not give too much about it.

bofire/data_models/features/engineered_feature.py
jduerholt: Very good point. And I think you are right and it was already wrong in the beginning with OneFeatureKeys. It should be just one key with len(features)==1

jduerholt:
Thanks @TobyBoyne — pushed as f7c54759. One of your points I did not apply, with evidence; the rest are in.

## `CloneFeature` now takes exactly one feature

You were right that the descriptions were written for one feature while the type allowed many, and @jduerholt confirmed the intent. `OneFeatureKeys` is gone — nothing else used it (three references in the whole repo: its definition, the import, and this field) — replaced by an inline `Annotated[List[str], Field(min_length=1, max_length=1)]`.

Worth flagging: **I was wrong when I said no call site passes more than one key.** I had grepped for `CloneFeature(` constructor calls and missed that the spec suite carried a two-key `CloneFeature` as an `add_valid`. It is now an `add_invalid`, and since this is a breaking validation change it gets a CHANGELOG entry. Cloning several features means one `CloneFeature` each, which the docstring now says.

## `validate_default=True` is required — not removed

Your suggestion would break `CategoricalInput`. Without the flag the `generate_allowed` validator never runs on the default, so `allowed` stays `None` instead of becoming `[True, ...]`:

```
with    validate_default: [True, True]
without validate_default: None
```

## `InterpolateFeature`

Your four questions were all fair, and three had concrete answers I have now written down:

- **`y_keys` are input features**, not outputs — `validate_features` requires every key in `features` to be a `ContinuousInput`.
- **`n_interpolation_points=1000` does not add 1000 dimensions to your domain.** Engineered features are configured on the *surrogate*, via `BotorchSurrogate.engineered_features`, so the columns are appended to what the model sees while the optimization problem stays the control points. That answer applies to all six engineered features, so it now lives once on the `EngineeredFeature` base class.
- **"the optimized inputs"** is replaced by a statement of when to reach for it: when several continuous inputs are really the coordinates of a curve rather than independent settings, so that two candidates with different control points but similar curves look similar to the model.
- The example now says what `t1`/`t2`/`T1`/`T2` are.

@jduerholt asked me not to spend too long on this one, so I have aimed at answering your questions rather than at a full treatment.

## Also applied

- **`Descriptors`** — your rewrite, near enough verbatim. The level explanation no longer enumerates how each `Input` type uses the block; the two examples carry that instead.
- **`fidelities`** — dropped the "opposite to the everyday sense" aside you disliked, and fixed the contradiction you spotted: it now says the levels used must run from 0 upwards without gaps, while several tasks may share a level. That matches the validator, which checks `set(fidelities) == range(max+1)`.
- **"null" → "None"** throughout, per @jduerholt's reply on `continuous.py` — these are Python docstrings.

## Verification

- `pytest tests/bofire/data_models` — 1853 passed, 4 skipped
- `pytest tests/bofire -k "clone or engineered or feature or input or output"` — 666 passed, 5 skipped
- ruff 0.6.5 clean; `ty==0.0.16` at 151 diagnostics, unchanged from main

TobyBoyne (approved):
Thanks for making the changes, happy with them all now!

I hadn't realised that `validate_default` also does the conversion logic, thanks for pointing that out.

jduerholt:
Correction on `CloneFeature` — the restriction in f7c54759 was wrong and is reverted in d782bf62.

Multi-key cloning is deliberate, not an oversight in the type. `CloneFeature` copies one *or more* features under a single engineered key, producing one column each, named `{key}_{feature}`:

```python
>>> c = CloneFeature(key="rates_copy", features=["flow_a", "flow_b"])
>>> c.get_names(inputs)
['rates_copy_flow_a', 'rates_copy_flow_b']
```

The consequence is the part that was genuinely missing from the docs, and it answers @TobyBoyne's question about how you would reach an individual clone. `BotorchSurrogate.get_feature_indices` (`bofire/surrogates/botorch.py:105`) resolves an engineered feature key to *every* column that feature produced, so a `FeatureSpecificKernel` naming `rates_copy` gets both columns and cannot pick one. Cloning features separately — one `CloneFeature` each — is what makes them individually addressable, and that is the actual difference between the two shapes.

So:

- `OneFeatureKeys` is restored, and its `min_length=1` semantics are correct as they stand
- the two-key `add_valid` spec is restored and the `add_invalid` I added is gone
- the CHANGELOG entry is dropped — there is no behaviour change here after all; `CHANGELOG.md` and `types.py` are now byte-identical to main
- the docstring states the grouping and its consequence, with examples of both a single clone and a two-feature group showing the column names it yields

My apologies for the churn — I took the review thread as settling the intent and restricted the type before checking what the implementation actually supported, which is backwards.

Everything else from the review round stands unchanged.

Verification: `pytest tests/bofire/data_models` — 1856 passed, 4 skipped (back to the pre-restriction count); `pytest tests/bofire -k "clone or engineered or feature"` — 512 passed, 2 skipped; ruff 0.6.5 clean.

PR #807, opened 2026-08-27
Document the acquisition function data models

jduerholt:
Fourth package batch of the migration started in #802 (#803 objectives/priors, #806 features). Drains the `acquisition_functions` allowlist to zero — **28 field descriptions and 16 class docstrings**, taking the ratchet from 459 to 415 remaining entries.

Almost none of it was harvestable: only `qLogPF` carried an `Attributes:` block, so this is close to 44 pieces of new prose. It is also the package that most needed it — the level-2 digest prototype in #804 is literally this union, and it currently reports `qUCB beta=0.2` with nothing about what `beta` does.

## Deduplication

`n_mc_samples` was declared on **13 of the 14** concrete acquisition functions and `alpha` on **all 5** multi-objective ones. Since redeclaring a field drops the inherited description, each copy would have needed its own duplicate of the same sentence.

A new `MCAcquisitionFunction` base carries `n_mc_samples`; `alpha` moves onto the existing `MultiObjectiveAcquisitionFunction`. `pTS` keeps inheriting `SingleObjectiveAcquisitionFunction` alone — it is sample-path based and takes no Monte Carlo sample count, which is exactly the distinction the new base draws. `qMFHVKG` redeclares `n_mc_samples` because its default is 32 rather than 512.

Verified additive — no accepted input changes, no spec needed touching:

```
pTS has n_mc_samples : False
qEI  n_mc_samples    : 512 (inherited, described)
qMFHVKG n_mc_samples : 32  (own default and description)
qMFHVKG alpha        : inherited description
qUCB model_dump      : {'type': 'qUCB', 'n_mc_samples': 512, 'beta': 0.2}
```

`prune_baseline` stays duplicated: its four classes straddle both branches, so their only common ancestor is one that `pTS` shares too.

## Claims I derived rather than read

Flagging these explicitly, because the semantics here are BoTorch's rather than BoFire's and three of my inferences were caught in review on the last two PRs. BoFire only forwards these parameters (`sobo.py:90-105`, `mobo.py:80-94`), so the meanings below come from BoTorch, not from anything in this repo:

- **`alpha`** — "tolerance for approximating the partitioning of the objective space; 0 partitions exactly, larger trades accuracy for speed"
- **`tau`** — "temperature smoothing the improvement indicator so it can be differentiated"
- **`beta`** — "how strongly uncertainty is rewarded relative to the predicted mean". Directionally right; I have not spelled out BoTorch's internal scaling
- **`prune_baseline`** — "drops already-observed points unlikely to be the best / on the Pareto front"
- **The recommendations** — that `qLogNEI` and `qLogNEHVI` are the usual defaults, and that the log variants keep a gradient where the plain ones underflow

`num_fantasies` and `num_pareto` I read from `multi_fidelity_knowledge_gradient.py:221-276` rather than inferring.

## One thing to decide separately

**`qMFHVKG.alpha` appears to be inert.** Its strategy reads `n_mc_samples`, `num_fantasies` and `num_pareto` (`multi_fidelity_knowledge_gradient.py:132-134`) and never `alpha`, so a caller can set it and nothing happens. That predates this PR — `qMFHVKG` already declared the field, it now inherits it instead — and I have deliberately **not** documented it as intended behaviour, since it looks more like something to fix than to describe. Worth its own look.

## Also

- `qNegIntPosVar.weights` used `Field(default_factory=lambda: None)` — a factory returning a constant, which kept the default out of `model_json_schema()` (the #804 defect). Now a literal default, and it appears there.
- `CLAUDE.md` and `CONTRIBUTING.md` record the rule from the #806 review: describe only what a caller can set or observe, never a BoFire-internal artefact; state the consequence instead.

## Rename candidates

Carried forward, not acted on: **`alpha`** and **`tau`** are BoTorch names whose meaning the field name does not convey at all — both descriptions exist purely to compensate.

## Verification

- `pytest tests/bofire/data_models` — 1880 passed, 4 skipped
- `pytest tests/bofire -k "sobo or mobo or acquisition or active_learning or qparego or knowledge"` — 207 passed, 8 skipped
- ruff 0.6.5 clean
- `ty==0.0.16` — 158 diagnostics, identical to the merged-main baseline (the count rose from 151 with #724, not with this PR)


bofire/data_models/acquisition_functions/acquisition_function.py
```diff
@@ -7,103 +7,274 @@
...
+    the model's belief about the already-observed points, which is the right choice when
+    the measurements are noisy. Prefer `qLogNEI`, whose reformulation optimizes more
+    reliably.
+    """
 
-class qNEI(SingleObjectiveAcquisitionFunction):
     type: Literal["qNEI"] = "qNEI"
-    prune_baseline: bool = True
-    n_mc_samples: IntPowerOfTwo = 512
+    prune_baseline: bool = Field(
```
jduerholt: Would it make sense to also have a noisy mixin class? for the prune_baseline?

bofire/data_models/acquisition_functions/acquisition_function.py
```diff
@@ -7,103 +7,274 @@
...
+    accurate everywhere rather than a single optimum.
+    """
 
-class qNegIntPosVar(SingleObjectiveAcquisitionFunction):
     type: Literal["qNegIntPosVar"] = "qNegIntPosVar"
-    n_mc_samples: IntPowerOfTwo = 512
-    weights: Optional[Dict[str, PositiveFloat]] = Field(default_factory=lambda: None)
+    weights: Optional[Dict[str, PositiveFloat]] = Field(
+        default=None,
+        description="How much reducing the uncertainty of each output counts, keyed by "
```
jduerholt, line 302: We should mention here that it is additive.

bofire/data_models/acquisition_functions/acquisition_function.py
```diff
@@ -112,11 +283,8 @@ class qLogPF(SingleObjectiveAcquisitionFunction):
     and is especially useful in combination with the FeasibleExperimentCondition
     within the StepwiseStrategy.
 
-    Attributes:
-        n_mc_samples: Number of Monte Carlo samples to use to
-            approximate the probability of feasibility.
-
+    It optimizes for feasibility alone and ignores the objectives entirely, so it is a
```
jduerholt: This is misleading, the feasibility is also defined by the obnjective classes in bofire namely the constrainedobnective stiuff, the sigmoid objectives.

jduerholt (commented):
Some comments.

jduerholt:
Pushed as 73818eb7. All three applied, and checking one of them turned up a further error of mine.

## Noisy mixin — yes, good call

`NoisyAcquisitionFunction` now carries `prune_baseline`, mirroring what `MCAcquisitionFunction` does for `n_mc_samples`. I had said in the PR body that hoisting was impossible because `qNEI`/`qLogNEI`/`qNEHVI`/`qLogNEHVI` straddle both branches — true for a shared *ancestor*, but a mixin sidesteps it, exactly as the MC one does. It also names a real distinction: these are precisely the variants that do not take the observed results at face value.

```
qNEI       {'type': 'qNEI', 'prune_baseline': True, 'n_mc_samples': 512}
qLogNEHVI  {'type': 'qLogNEHVI', 'alpha': 0.0, 'prune_baseline': True, 'n_mc_samples': 512}
qEI        {'type': 'qEI', 'n_mc_samples': 512}          # correctly no prune_baseline
pTS        {'type': 'pTS'}                                # still field-free
```

The two descriptions I had written differed — "unlikely to be the best one" versus "unlikely to lie on the Pareto front" — so the shared one is phrased neutrally as points that "cannot plausibly be optimal", which covers both without going vague.

## `qNegIntPosVar.weights` — additive, and I had the normalization wrong

Adding "additive" sent me back to `active_learning.py:58-70`, where I found my own description was wrong. **Only the default branch normalizes** (`torch.ones(ny) / ny`); weights you supply are passed through exactly as given. I had claimed "the values are normalized", which is false.

Since they feed a `ScalarizedPosteriorTransform`, the outputs are combined **additively** as a weighted sum of variances — so scaling every weight is a no-op for the argmax and it is the ratios that decide the winner. That is what it now says, without the false normalization claim.

## `qLogPF` — you're right, and it was the reverse of the truth

I wrote that it "ignores the objectives entirely". Feasibility in BoFire *is* defined by objectives — those subclassing `ConstrainedObjective`, i.e. the sigmoid objectives, `TargetObjective` and `ConstrainedCategoricalObjective`. `sobo.py:167` swaps in an `IdentityMCObjective` while still passing the constraint callables, so what `qLogPF` drops is the objective being *optimized*, not the constraint-defining ones. The docstring now names those classes explicitly, since that is how a caller controls what feasible means.

This was one of the five claims I flagged as inferred rather than read — the flag worked, though I would rather have caught it myself.

## Verification

- `pytest tests/bofire/data_models` — 1882 passed, 4 skipped
- `pytest tests/bofire -k "sobo or mobo or acquisition or active_learning or qparego or knowledge"` — 209 passed, 8 skipped
- ruff 0.6.5 clean; `ty==0.0.16` at 158, matching the merged-main baseline

Still open from the PR body: **`qMFHVKG.alpha` looks inert** — its strategy reads only `n_mc_samples`, `num_fantasies` and `num_pareto` (`multi_fidelity_knowledge_gradient.py:132-134`). I have left it undocumented rather than describing a field that does nothing.

jduerholt:
@TobyBoyne: can you review again? I really like your reviews and affects also some stuff added by you, I think.

TobyBoyne:
@jduerholt Of course! I will try to review today :)

bofire/data_models/acquisition_functions/acquisition_function.py
```diff
@@ -7,103 +7,273 @@
...
+        "optimization step. Must be a power of two.",
+    )
+
+
+class NoisyAcquisitionFunction(AcquisitionFunction):
+    """Acquisition function that treats the observed results as noisy.
+
+    Rather than taking the best result so far at face value, these integrate over the
+    model's belief about the points already observed. That is the right choice whenever
+    a repeated experiment would not return exactly the same number.
```
TobyBoyne: I'm not actually too sure on the state of the art here, but if there is expected to be a *small* amount of noise, should one still use a `NoisyAcquisitionFunction`? Or should you only use it if you expect the noise to be significant, and otherwise the price to pay for generating fantasies might be greater than the gain from accounting for noise.

If the latter, the docstring should be updated to reflect that. But again, not too sure on actual performance - is there any downside to always using `NoisyAcquisitionFunction`?

bofire/data_models/acquisition_functions/acquisition_function.py
```diff
@@ -7,103 +7,273 @@
...
+
+    type: Any
+
+
+class MCAcquisitionFunction(AcquisitionFunction):
+    """Acquisition function whose value is approximated by Monte Carlo sampling.
+
+    Sampling is what lets these score a whole batch of candidates jointly rather than one
+    point at a time, at the cost of an approximation whose noise falls as the sample
+    count rises.
```
TobyBoyne: (Really tiny nit) I think docstrings tend to be singular, e.g., "This uses sampling to score a whole batch ...." instead of "These". Applies to all the docstrings.

bofire/data_models/acquisition_functions/acquisition_function.py
```diff
@@ -7,103 +7,273 @@
...
+        default=0.0,
+        description="Tolerance for approximating the partitioning of the objective "
+        "space used in the hypervolume computation. The default of 0 partitions "
+        "exactly; larger values trade accuracy for speed, which matters as the number "
+        "of objectives grows.",
+    )
+
+
+class qNEI(
+    MCAcquisitionFunction, NoisyAcquisitionFunction, SingleObjectiveAcquisitionFunction
```
TobyBoyne: (Small nit) The inheritance is a bit funny here, since this inherits from `AcquisitionFunction` three times (once through each parent). I would maybe change `NoisyAcquisitionFunction` and `SingleObjectiveAcquisitionFunction` to Mixins that inherit directly from base model.

bofire/data_models/acquisition_functions/acquisition_function.py
```diff
@@ -7,103 +7,273 @@
...
 
 
 class pTS(SingleObjectiveAcquisitionFunction):
-    """Pathwise Thompson Sampling acquisition function."""
+    """Pathwise Thompson Sampling acquisition function.
+
+    Draws a sample function from the model and proposes its optimum, so exploration
+    comes from the randomness of the draw rather than from an explicit trade-off
+    parameter. It takes no settings, and being sample-path based it needs no Monte Carlo
+    sample count.
```
TobyBoyne: Remove "It takes no settings, and being sample-path based it needs no Monte Carlo sample count"

bofire/data_models/acquisition_functions/acquisition_function.py
```diff
@@ -7,103 +7,273 @@
...
 
 
-class qSR(SingleObjectiveAcquisitionFunction):
+class qSR(MCAcquisitionFunction, SingleObjectiveAcquisitionFunction):
+    """Simple regret: the expected value of the output itself.
+
+    Purely exploitative — it ignores uncertainty and heads for wherever the model
+    predicts the best mean, which makes it a reasonable final-step choice but prone to
+    getting stuck if used throughout a campaign. It cannot handle output constraints
+    separately and falls back to a constrained Monte Carlo objective.
```
TobyBoyne: Not too sure what this last sentence means - it sounds like it can handle constraints?

Should the `Strategy` validation prevent output constraints with AFs that don't support them?

bofire/data_models/acquisition_functions/acquisition_function.py
```diff
@@ -7,103 +7,273 @@
...
     type: Literal["qSR"] = "qSR"
-    n_mc_samples: IntPowerOfTwo = 512
 
 
-class qUCB(SingleObjectiveAcquisitionFunction):
+class qUCB(MCAcquisitionFunction, SingleObjectiveAcquisitionFunction):
+    """Upper confidence bound: predicted mean plus a multiple of the uncertainty.
+
+    The one acquisition function whose exploration is set explicitly rather than implied,
+    which makes it the natural choice when you want direct control over that trade-off.
```
TobyBoyne: (Nit) This docstring implicitly couples to other AFs (by being "The one..."). Should instead read "The exploration of this acquisition function is set explicitly by the `beta` parameter.

I also am a bit unsure of "...when you want direct control over that trade-off", since it is very hard to tune the parameter. But happy to leave it in.

bofire/data_models/acquisition_functions/acquisition_function.py
```diff
@@ -7,103 +7,273 @@
...
+    comes from the randomness of the draw rather than from an explicit trade-off
+    parameter. It takes no settings, and being sample-path based it needs no Monte Carlo
+    sample count.
+    """
 
     type: Literal["pTS"] = "pTS"
 
 
-class qEI(SingleObjectiveAcquisitionFunction):
+class qEI(MCAcquisitionFunction, SingleObjectiveAcquisitionFunction):
```
TobyBoyne: Is this PR a good time to start deprecating the non-log variants of the acquisition functions? I don't know of any good reason to keep them, I feel like it's just asking for users to use the wrong one. Maybe outside the scope of this PR (also not sure we have an official deprecation process).

bofire/data_models/acquisition_functions/acquisition_function.py
```diff
@@ -7,103 +7,273 @@
...
-    num_fantasies: int = 8
-    num_pareto: int = 10
-    n_mc_samples: IntPowerOfTwo = 32
+class qMFHVKG(MCAcquisitionFunction, MultiObjectiveAcquisitionFunction):
+    """Multi-fidelity hypervolume knowledge gradient.
 
+    Scores how much running a candidate would improve the Pareto front you could report
+    afterwards, rather than how good the candidate itself looks. That look-ahead is what
+    lets it decide to spend a cheap low-fidelity evaluation for information alone, so it
+    is the multi-objective choice for a domain with a task or fidelity input.
```
TobyBoyne: This is a little bit untrue, and I don't think the wording is very clear. For reference, see Chapter 7 of Bayesian Optimization by Roman Garnett.

EI is a one-step lookahead function on the *simple* reward. Its maximizer $ x^* $ gives the optimal experiment assuming that, after I evaluate $ f(x^*) $, I will take the best observed value as my best experiment.

KG is a one-step lookahead function on the *global* reward. Its maximizer $ x^* $ gives the optimal experiment assuming that, after I evaluate $ f(x^*) $, I will take the optimum of the posterior mean as my best experiment.
 
Suggested rewording:

> Knowledge gradient proposes a candidate that maximizes the *global reward* - the maximum of the surrogate's mean prediction after observing the experiment at the candidate. This encourages proposing cheap, low-fidelity candidates at the current iteration to give more information about the best experiment at the next iteration. This multi-objective acquisition function should be used for a domain with a task or fidelity input.

bofire/data_models/acquisition_functions/acquisition_function.py
```diff
@@ -112,11 +282,12 @@ class qLogPF(SingleObjectiveAcquisitionFunction):
...
-    Attributes:
-        n_mc_samples: Number of Monte Carlo samples to use to
-            approximate the probability of feasibility.
-
+    What counts as feasible comes from the outputs carrying one of the objectives that
+    define a constraint, namely `MaximizeSigmoidObjective`, `MinimizeSigmoidObjective`,
+    `MovingMaximizeSigmoidObjective`, `TargetObjective` and
+    `ConstrainedCategoricalObjective`. Those still shape the score; what this drops is
+    the objective being optimized, so it is a way out of an infeasible region rather
+    than a way to a good result.
```
TobyBoyne: The "what"s here hurt readability. Also I would avoid enumerating objectives, to prevent coupling docstrings.

Maybe rewrite the whole thing as

> This acquisition function is evaluated against all of the outputs with *constrained* objectives. It does not account for the optimization objectives; it provides way out of an infeasible region rather than a high objective value.

To me, this raises a separate issue - it isn't clear that the listed objectives are constrained (`TargetObjective` especially, but also `MaximizeSigmoidObjective`). The fact that `CloseToTargetObjective` also exists, but is the version that doesn't apply constraints, is a bit of a codesmell to me. Maybe worth changing all of those class names to `...ConstrainedObejctive`? Definitely outside the scope of this PR, but the point stands that if the constrained objectives need to be enumerated in a docstring, then something has maybe gone wrong.

TobyBoyne (commented):
Happy to see even *more* documentation!

Most of my comments are pretty nitpicky here, sorry! Mostly all looks good.

I do think the point about `ConstrainedObjectives` is an issue. Probably one for a different time, but worth flagging!

TobyBoyne:
> qMFHVKG.alpha appears to be inert

I think it is meant to be that way. I could only find one place where the partitioning is used in the `qMFHVKG` AF ([here](https://github.com/meta-pytorch/botorch/blob/917535f0f2892ebecdf58c9c90e4f2feb96fd44b/botorch/acquisition/multi_objective/hypervolume_knowledge_gradient.py#L559-L566)), and it's an empty partitioning so I don't think `alpha` would do anything.

bofire/data_models/acquisition_functions/acquisition_function.py
jduerholt: Downside is only that it is computationally less efficient.

bofire/data_models/acquisition_functions/acquisition_function.py
jduerholt: Yep, we should do this.

bofire/data_models/acquisition_functions/acquisition_function.py
jduerholt: You are right, it can handlie constraints but uses a different mechanism for it. We should remove it.

bofire/data_models/acquisition_functions/acquisition_function.py
jduerholt: Yeah, we should introduce deperaction warnings here and point people to the log equivalents.

bofire/data_models/acquisition_functions/acquisition_function.py
jduerholt: Yeah, we take the enum out and maybe rename in a sperate PR .

jduerholt:
Thanks @TobyBoyne — all applied across bab130d8 and 207266cf.

## Hierarchy (your line 81)

`MCAcquisitionFunction` and `NoisyAcquisitionFunction` are now plain `BaseModel` mixins, so `AcquisitionFunction` appears **exactly once** in every concrete MRO instead of two or three times.

One deviation from your suggestion, because it does not quite work: making `SingleObjectiveAcquisitionFunction` a mixin would leave `pTS` — which inherits only from it — not being an `AcquisitionFunction` at all. So I made the *capability* classes mixins and kept the *taxonomy* in the hierarchy, which reaches the same single-path result. Verified every concrete class is still `isinstance(..., AcquisitionFunction)` and no serialized output changed.

## qMFHVKG (your line 235)

You were right that it was wrong, and I have taken your wording almost verbatim. I had described it as scoring the improvement to the *reportable Pareto front*, which conflates it with EI's simple reward; knowledge gradient is a one-step lookahead on the **global reward**, the maximum of the surrogate's mean prediction after observing the candidate. Thanks for the Garnett reference.

## qLogPF (your line 290)

Enumeration removed, rewritten close to your suggestion. Worth noting the history: the list was added at @jduerholt's request in the previous round, so this reverses that — going with the newer decision on this thread.

Your wider point stands and is worth its own issue. Needing to enumerate which objectives are "constrained" is a symptom: `TargetObjective` constrains while `CloseToTargetObjective` does not, and nothing in either name says so.

## Also applied

- **Singular voice** throughout — no more "these"/"they".
- **`pTS`** — dropped the sentence about taking no settings.
- **`qSR` / `qUCB`** — the claim that they cannot handle output constraints is gone. Per @jduerholt they can, by a different mechanism, so the sentence was simply wrong.
- **`qUCB`** — no longer defines itself by reference to the other acquisition functions, and now admits `beta` is hard to tune: *"The exploration of this acquisition function is set explicitly, by `beta` … That parameter is not easy to tune, so the explicitness is worth having mainly when a particular trade-off is being imposed deliberately."*
- **`NoisyAcquisitionFunction`** — answers your question directly, using @jduerholt's reply: integrating over the observed points is never *wrong*, it just costs more compute on noise-free data, so the choice is efficiency rather than correctness.

## Deprecation (your line 117)

Done in this PR, per @jduerholt. `qEI`, `qNEI`, `qEHVI` and `qNEHVI` emit a `DeprecationWarning` on construction naming their log replacement, with a CHANGELOG entry under a new **Deprecated** heading.

A `SupersededByLogVariant` mixin carries the warning and each class names its replacement in a `ClassVar`, so the pointer is declared once per class rather than duplicated into four near-identical validators. Being a `ClassVar` it stays out of `model_fields`, so neither `model_dump()` nor the JSON schema changes:

```
qEI    [DeprecationWarning] qEI is deprecated ... Use qLogEI instead
qEI    fields: ['type', 'n_mc_samples']        # log_variant does not leak
qLogEI, qLogNEI, qLogEHVI, qLogNEHVI, qUCB, pTS — 0 warnings
```

## Verification

- `pytest tests/bofire/data_models` — 1884 passed, 4 skipped
- `pytest tests/bofire -k "sobo or mobo or acquisition or active_learning or qparego or knowledge"` — 211 passed, 8 skipped
- ruff 0.6.5 clean; `ty==0.0.16` at 158, matching the merged-main baseline

Still open and deliberately untouched: **`qMFHVKG.alpha` appears to be inert** — its strategy reads only `n_mc_samples`, `num_fantasies` and `num_pareto`.

TobyBoyne (approved):
Thanks for making the changes!

PR #809, opened 2026-08-31
Document the kernel data models

jduerholt:
Fifth package batch of the migration started in #802 (#803 objectives/priors, #806 features, #807 acquisition functions). Drains the `kernels` allowlist to zero — **69 field descriptions and 18 class docstrings**, taking the ratchet from 415 to 328 remaining entries.

## Deduplication

26 of the 69 fields were the same few concepts restated: `lengthscale_prior` and `lengthscale_constraint` on **seven** kernels, `ard` on **six**.

The groupings do not nest — `TanimotoKernel` has `ard` but no lengthscale fields, and the two Wasserstein kernels have the lengthscale fields but no `ard` — so two mixins carry them, `ARDKernel` and `LengthscaleKernel`. Both are plain `BaseModel` mixins rather than `Kernel` subclasses, which is the shape @TobyBoyne asked for in #807: `Kernel` now appears exactly once in every concrete MRO.

Twenty declarations become three descriptions, with no change to serialized output:

```
RBFKernel           ard=True   ls_prior=True   Kernel-in-MRO=1
TanimotoKernel      ard=True   ls_prior=False  Kernel-in-MRO=1
WassersteinKernel   ard=False  ls_prior=True   Kernel-in-MRO=1
LinearKernel        ard=False  ls_prior=False  Kernel-in-MRO=1
```

**Left duplicated on purpose:** `num_categories`, `rank`, `prior` and `var_constraint` on `IndexKernel` and `PositiveIndexKernel` — those two are *siblings* under `CategoricalKernel`, not parent and child, so sharing them would mean adding a class to the categorical taxonomy. Same reasoning for `outputscale_prior`, `kernels` and `offset_prior`, each spanning classes with no suitable common ancestor.

## Claims I derived rather than read

Same flagging as #807, since this is again GPyTorch semantics rather than BoFire's. `bofire/kernels/mapper.py` shows *that* a parameter is forwarded, not what it means, so these came from GPyTorch:

- **`nu`** — that only 0.5/1.5/2.5 are supported is read from the validator, but "0.5 gives a rough, non-differentiable surface, 2.5 a nearly smooth one, larger approaches RBF" is not
- **`rank`** on the index kernels — "a low rank forces the categories onto a few shared dimensions, which regularizes the fit"
- **`power`** on `PolynomialKernel` and `offset_prior` — "how much weight the lower-order terms carry relative to the highest one"
- **`WedgeKernel.angle_prior` / `radius_prior`** — "how different an active candidate is from an inactive one" and "how much the value of an active feature matters relative to whether it is active at all". These are my weakest inferences in the PR; the mapper (`mapper.py:453`) shows only that they are forwarded
- **`DownsamplingKernel`** — that the kernel assumes the approximation improves monotonically with fidelity
- **`SphericalLinearKernel`** — that projecting onto the sphere makes it respond to direction rather than magnitude, hence its suitability for compositions

What I *did* read: `ard` becomes `ard_num_dims=len(active_dims) if ard else None` (`mapper.py:90,115`); priors are mapped with the problem dimensionality via `priors.map(..., d=len(active_dims))` (`mapper.py:92-94`), which is what makes the dimensionality-scaled priors from #803 meaningful here; and `FeatureSpecificKernel.features` resolves through `get_feature_indices` (`bofire/surrogates/botorch.py:105`), the same mechanism documented for `CloneFeature` in #806 — so naming an engineered feature selects every input it contributes.

## Two things to look at separately

- **`InfiniteWidthBNNKernel` redeclares `features`** (`continuous.py:91`) as plain `Optional[List[str]]` where `FeatureSpecificKernel` uses `Optional[NonRestrictedFeatureKeys]`, silently dropping the uniqueness validator. Nothing says whether that is deliberate. Following the `CloneFeature` lesson from #806 I have **not** changed the validation — it gets its own description and this flag.
- **`AdditiveKernel` declared `type` twice**, once before and once after its `kernels` field. Removed, since it is unambiguously redundant.

Still open from #807: **`qMFHVKG.alpha` appears inert**.

## Also

Includes the guard test offered during #807 review: every deprecated acquisition function's `log_variant` must name a real member of `AnyAcquisitionFunction`, so a typo in that string fails the suite rather than surfacing in a warning a user eventually reads.

## Rename candidates

Carried forward: **`ard`** (the acronym means nothing without the description), **`squared`** on the Wasserstein kernels (squared *what*), and **`nu`**.

## Verification

- `pytest tests/bofire/data_models` — 1889 passed, 4 skipped
- `pytest tests/bofire -k "kernel or surrogate"` — 647 passed, 2 skipped
- ruff 0.6.5 clean; `ty==0.0.16` at 158, matching the merged-main baseline

### DIFF
--- a/bofire/data_models/kernels/aggregation.py
+++ b/bofire/data_models/kernels/aggregation.py
@@ -1,6 +1,8 @@
 from collections.abc import Sequence
 from typing import Literal, Optional, Union
 
+from pydantic import Field
+
 from bofire.data_models.kernels.categorical import (
     HammingDistanceKernel,
     IndexKernel,
@@ -23,6 +25,12 @@ from bofire.data_models.priors.api import AnyPrior, AnyPriorConstraint
 
 
 class AdditiveKernel(AggregationKernel):
+    """Sum of several kernels.
+
+    Two candidates count as similar if they are similar under any one of the summed
+    kernels, which suits a response made of separable contributions.
+    """
+
     type: Literal["AdditiveKernel"] = "AdditiveKernel"
     kernels: Sequence[
         Union[
@@ -42,11 +50,16 @@ class AdditiveKernel(AggregationKernel):
             "MultiplicativeKernel",
             "ScaleKernel",
         ]
-    ]
-    type: Literal["AdditiveKernel"] = "AdditiveKernel"
+    ] = Field(description="The kernels to sum.")
 
 
 class MultiplicativeKernel(AggregationKernel):
+    """Product of several kernels.
+
+    Two candidates count as similar only if they are similar under every one of the
+    multiplied kernels, which is how an interaction between input types is expressed.
+    """
+
     type: Literal["MultiplicativeKernel"] = "MultiplicativeKernel"
     kernels: Sequence[
         Union[
@@ -66,10 +79,16 @@ class MultiplicativeKernel(AggregationKernel):
             "MultiplicativeKernel",
             "ScaleKernel",
         ]
-    ]
+    ] = Field(description="The kernels to multiply.")
 
 
 class ScaleKernel(AggregationKernel):
+    """Wraps another kernel with a fitted output scale.
+
+    The base kernel decides which candidates are similar; this decides how much the
+    response actually varies, which is why most kernels are wrapped in one before use.
+    """
+
     type: Literal["ScaleKernel"] = "ScaleKernel"
     base_kernel: Union[
         RBFKernel,
@@ -87,11 +106,19 @@ class ScaleKernel(AggregationKernel):
         "ScaleKernel",
         WassersteinKernel,
         ExactWassersteinKernel,
-    ]
+    ] = Field(description="The kernel whose output is scaled.")
     # the ScaleKernel mapper forwards the dimensionality d to the outputscale prior, so
     # dimensionality-scaled priors are supported here.
-    outputscale_prior: Optional[AnyPrior] = None
-    outputscale_constraint: Optional[AnyPriorConstraint] = None
+    outputscale_prior: Optional[AnyPrior] = Field(
+        default=None,
+        description="Prior over the output scale, which sets how far the response is "
+        "expected to vary. If not provided, the surrogate's default is used.",
+    )
+    outputscale_constraint: Optional[AnyPriorConstraint] = Field(
+        default=None,
+        description="Hard bounds on the output scale, enforced during fitting rather "
+        "than merely preferred as a prior is.",
+    )
 
 
 class PolynomialFeatureInteractionKernel(AggregationKernel):
@@ -127,11 +154,6 @@ class PolynomialFeatureInteractionKernel(AggregationKernel):
     ])
     ```
 
-    Attributes:
-        kernels: The base kernels that should interact.
-        max_degree: Maximum degree of interactions computed.
-        include_self_interactions: Whether a kernel is allowed to interact with itself.
-        outputscale_prior: The prior used to scale each interaction term before summing.
     """
 
     type: Literal["PolynomialFeatureInteractionKernel"] = (
@@ -153,10 +175,24 @@ class PolynomialFeatureInteractionKernel(AggregationKernel):
             WassersteinKernel,
             ExactWassersteinKernel,
         ]
-    ]
-    max_degree: int
-    include_self_interactions: bool
-    outputscale_prior: Optional[AnyPrior] = None
+    ] = Field(
+        description="The kernels whose interactions are computed. Typically one per "
+        "kind of input, so that the interactions run across the kinds.",
+    )
+    max_degree: int = Field(
+        description="Highest interaction order computed. 1 keeps the kernels "
+        "independent; 2 adds every pairwise interaction, and so on. The number of terms "
+        "grows quickly with this.",
+    )
+    include_self_interactions: bool = Field(
+        description="Whether a kernel may interact with itself, which adds the "
+        "quadratic and higher powers of each kernel alongside the cross terms.",
+    )
+    outputscale_prior: Optional[AnyPrior] = Field(
+        default=None,
+        description="Prior over the output scale applied to each interaction term "
+        "before the terms are summed. If not provided, the surrogate's default is used.",
+    )
 
 
 AdditiveKernel.model_rebuild()
--- a/bofire/data_models/kernels/aggregation.py
+++ b/bofire/data_models/kernels/aggregation.py
@@ -88,8 +88,7 @@ class ScaleKernel(AggregationKernel):
     $k(\mathbf x, \mathbf x') = \theta\, k_{\text{base}}(\mathbf x, \mathbf x')$.
 
     The base kernel sets the shape of the covariance and this sets its magnitude, which
-    is the variance of the response. Most kernels are wrapped in one before use, since
-    the base kernels are normalized.
+    is the variance of the response.
     """
 
     type: Literal["ScaleKernel"] = "ScaleKernel"
@@ -120,9 +119,8 @@ class ScaleKernel(AggregationKernel):
     )
     outputscale_constraint: Optional[AnyPriorConstraint] = Field(
         default=None,
-        description="Bounds the output scale is reparametrized into, making values "
-        "outside them unreachable during fitting, unlike a prior which only shifts the "
-        "fitted value.",
+        description="Bounds the output scale $\\theta$ is restricted to during "
+        "fitting.",
     )
 
 
@@ -181,13 +179,11 @@ class PolynomialFeatureInteractionKernel(AggregationKernel):
             ExactWassersteinKernel,
         ]
     ] = Field(
-        description="The kernels whose interactions are computed. Typically one per "
-        "kind of input, so that the interactions run across the kinds.",
+        description="The kernels whose interactions are computed.",
     )
     max_degree: int = Field(
         description="Highest interaction order computed. 1 keeps the kernels "
-        "independent; 2 adds every pairwise interaction, and so on. The number of terms "
-        "grows quickly with this.",
+        "independent; 2 adds every pairwise interaction, and so on.",
     )
     include_self_interactions: bool = Field(
         description="Whether a kernel may interact with itself, which adds the "
--- a/bofire/data_models/kernels/aggregation.py
+++ b/bofire/data_models/kernels/aggregation.py
@@ -25,10 +25,10 @@ from bofire.data_models.priors.api import AnyPrior, AnyPriorConstraint
 
 
 class AdditiveKernel(AggregationKernel):
-    """Sum of several kernels.
+    r"""Sum of several kernels, $k(\mathbf x, \mathbf x') = \sum_i k_i(\mathbf x, \mathbf x')$.
 
-    Two candidates count as similar if they are similar under any one of the summed
-    kernels, which suits a response made of separable contributions.
+    A sum models a response made of independent additive contributions, and the
+    covariance stays high if any one of the summed kernels is high.
     """
 
     type: Literal["AdditiveKernel"] = "AdditiveKernel"
@@ -54,10 +54,11 @@ class AdditiveKernel(AggregationKernel):
 
 
 class MultiplicativeKernel(AggregationKernel):
-    """Product of several kernels.
+    r"""Product of several kernels, $k(\mathbf x, \mathbf x') = \prod_i k_i(\mathbf x, \mathbf x')$.
 
-    Two candidates count as similar only if they are similar under every one of the
-    multiplied kernels, which is how an interaction between input types is expressed.
+    A product models an interaction: the covariance is high only where every one of the
+    multiplied kernels is high, so this is how a dependence across input types is
+    expressed.
     """
 
     type: Literal["MultiplicativeKernel"] = "MultiplicativeKernel"
@@ -83,10 +84,12 @@ class MultiplicativeKernel(AggregationKernel):
 
 
 class ScaleKernel(AggregationKernel):
-    """Wraps another kernel with a fitted output scale.
+    r"""Wraps another kernel with a fitted output scale,
+    $k(\mathbf x, \mathbf x') = \theta\, k_{\text{base}}(\mathbf x, \mathbf x')$.
 
-    The base kernel decides which candidates are similar; this decides how much the
-    response actually varies, which is why most kernels are wrapped in one before use.
+    The base kernel sets the shape of the covariance and this sets its magnitude, which
+    is the variance of the response. Most kernels are wrapped in one before use, since
+    the base kernels are normalized.
     """
 
     type: Literal["ScaleKernel"] = "ScaleKernel"
@@ -111,13 +114,15 @@ class ScaleKernel(AggregationKernel):
     # dimensionality-scaled priors are supported here.
     outputscale_prior: Optional[AnyPrior] = Field(
         default=None,
-        description="Prior over the output scale, which sets how far the response is "
-        "expected to vary. If not provided, the surrogate's default is used.",
+        description="Prior over the output scale $\\theta$, which sets the variance of "
+        "the response. If not provided, no prior is placed on it and it is fitted from "
+        "the data alone.",
     )
     outputscale_constraint: Optional[AnyPriorConstraint] = Field(
         default=None,
-        description="Hard bounds on the output scale, enforced during fitting rather "
-        "than merely preferred as a prior is.",
+        description="Bounds the output scale is reparametrized into, making values "
+        "outside them unreachable during fitting, unlike a prior which only shifts the "
+        "fitted value.",
     )
 
 
@@ -191,7 +196,8 @@ class PolynomialFeatureInteractionKernel(AggregationKernel):
     outputscale_prior: Optional[AnyPrior] = Field(
         default=None,
         description="Prior over the output scale applied to each interaction term "
-        "before the terms are summed. If not provided, the surrogate's default is used.",
+        "before the terms are summed. If not provided, no prior is placed on them and "
+        "they are fitted from the data alone.",
     )
 
 
--- a/bofire/data_models/kernels/categorical.py
+++ b/bofire/data_models/kernels/categorical.py
@@ -18,11 +18,16 @@ class CategoricalKernel(FeatureSpecificKernel):
 
 
 class HammingDistanceKernel(ARDKernel, LengthscaleKernel, CategoricalKernel):
-    """Kernel measuring similarity as the number of categories two candidates share.
+    r"""Kernel over categorical inputs based on the Hamming distance.
 
-    Treats the categories as unordered and equally distant from one another, which is
-    the right default unless there is structure among them that `IndexKernel` could
-    learn instead.
+    $$
+    k(\mathbf x, \mathbf x') = \exp\left(-\frac{d_H(\mathbf x, \mathbf x')}{\ell}\right)
+    $$
+
+    where $d_H$ counts the positions at which the two inputs hold different categories.
+    Every pair of distinct categories is therefore equally far apart, which is the right
+    assumption unless the categories have structure that `IndexKernel` could learn
+    instead.
     """
 
     type: Literal["HammingDistanceKernel"] = "HammingDistanceKernel"
@@ -43,7 +48,7 @@ class IndexKernel(CategoricalKernel):
     where $B$ is a low-rank matrix, and $\mathbf v$ is a  non-negative vector.
 
     Unlike `HammingDistanceKernel`, which holds all categories equally distant, this
-    learns how similar the categories are from the data.
+    learns from the data how similar the categories are.
     """
 
     type: Literal["IndexKernel"] = "IndexKernel"
@@ -59,13 +64,14 @@ class IndexKernel(CategoricalKernel):
     )
     prior: Optional[AnyPrior] = Field(
         default=None,
-        description="Prior over the entries of the low-rank matrix. If not provided, "
-        "the surrogate's default is used.",
+        description="Prior over the entries of $B$. If not provided, no prior is "
+        "placed on them and they are fitted from the data alone.",
     )
     var_constraint: Optional[AnyPriorConstraint] = Field(
         default=Positive(),
-        description="Constraint on the per-category variance added on the diagonal, "
-        "which keeps it non-negative.",
+        description="Bounds the diagonal entries $\\mathbf v$ are reparametrized into. "
+        "Unlike `prior`, which only shifts the fitted value, this makes values outside "
+        "the bounds unreachable; the default keeps the variances positive.",
     )
 
     @model_validator(mode="after")
@@ -77,7 +83,7 @@ class IndexKernel(CategoricalKernel):
 
 class PositiveIndexKernel(CategoricalKernel):
     r"""
-    Many a times the IndexKernel is not positive definite. This kernel addresses this
+    The IndexKernel is often not positive definite. This kernel addresses that
     by using Cholesky decomposition with positive elements only. So, off diagonal
     elements are always positive and the diagonal elements are normalized to 1 for a
     target task. Mathematically, the kernel is defined as:
@@ -106,13 +112,14 @@ class PositiveIndexKernel(CategoricalKernel):
     )
     prior: Optional[AnyPrior] = Field(
         default=None,
-        description="Prior over the entries of the triangular factor. If not provided, "
-        "the surrogate's default is used.",
+        description="Prior over the entries of $L$. If not provided, no prior is "
+        "placed on them and they are fitted from the data alone.",
     )
     var_constraint: Optional[AnyPriorConstraint] = Field(
         default=Positive(),
-        description="Constraint on the per-category variance added on the diagonal, "
-        "which keeps it non-negative.",
+        description="Bounds the diagonal entries $\\mathbf v$ are reparametrized into. "
+        "Unlike `prior`, which only shifts the fitted value, this makes values outside "
+        "the bounds unreachable; the default keeps the variances positive.",
     )
     task_prior: Optional[AnyPrior] = Field(
         default=None,
--- a/bofire/data_models/kernels/categorical.py
+++ b/bofire/data_models/kernels/categorical.py
@@ -18,16 +18,16 @@ class CategoricalKernel(FeatureSpecificKernel):
 
 
 class HammingDistanceKernel(ARDKernel, LengthscaleKernel, CategoricalKernel):
-    r"""Kernel over categorical inputs based on the Hamming distance.
+    r"""Kernel over categorical inputs, based on the Hamming distance.
 
     $$
-    k(\mathbf x, \mathbf x') = \exp\left(-\frac{d_H(\mathbf x, \mathbf x')}{\ell}\right)
+    k(\mathbf x, \mathbf x') = \exp\left(-\frac{d(\mathbf x, \mathbf x')}{\ell}\right)
     $$
 
-    where $d_H$ counts the positions at which the two inputs hold different categories.
-    Every pair of distinct categories is therefore equally far apart, which is the right
-    assumption unless the categories have structure that `IndexKernel` could learn
-    instead.
+    where $d$ is zero where two inputs hold the same category and one where they differ,
+    averaged over the categorical dimensions. With `ard` there is one lengthscale per
+    categorical feature, so a change of category can matter more in some features than
+    in others. The kernel is not differentiable with respect to its inputs.
     """
 
     type: Literal["HammingDistanceKernel"] = "HammingDistanceKernel"
@@ -69,9 +69,8 @@ class IndexKernel(CategoricalKernel):
     )
     var_constraint: Optional[AnyPriorConstraint] = Field(
         default=Positive(),
-        description="Bounds the diagonal entries $\\mathbf v$ are reparametrized into. "
-        "Unlike `prior`, which only shifts the fitted value, this makes values outside "
-        "the bounds unreachable; the default keeps the variances positive.",
+        description="Bounds the diagonal entries $\\mathbf v$ are restricted to during "
+        "fitting. The default keeps them positive.",
     )
 
     @model_validator(mode="after")
@@ -117,9 +116,8 @@ class PositiveIndexKernel(CategoricalKernel):
     )
     var_constraint: Optional[AnyPriorConstraint] = Field(
         default=Positive(),
-        description="Bounds the diagonal entries $\\mathbf v$ are reparametrized into. "
-        "Unlike `prior`, which only shifts the fitted value, this makes values outside "
-        "the bounds unreachable; the default keeps the variances positive.",
+        description="Bounds the diagonal entries $\\mathbf v$ are restricted to during "
+        "fitting. The default keeps them positive.",
     )
     task_prior: Optional[AnyPrior] = Field(
         default=None,
@@ -138,9 +136,9 @@ class PositiveIndexKernel(CategoricalKernel):
     )
     target_task_index: Annotated[int, Field(ge=0)] = Field(
         default=0,
-        description="Index into the categories of the one treated as the target, whose "
-        "diagonal entry is normalized to 1 so the others are expressed relative to it. "
-        "Must be below `num_categories` - 1.",
+        description="Position of the target category, the one the covariance matrix is "
+        "normalized against: $(LL^\\top)_{t,t}$ becomes the denominator, so the target "
+        "has unit variance and every other category is expressed relative to it.",
     )
     unit_scale_for_target: bool = Field(
         default=True,
--- a/bofire/data_models/kernels/categorical.py
+++ b/bofire/data_models/kernels/categorical.py
@@ -2,20 +2,30 @@ from typing import Annotated, Literal, Optional
 
 from pydantic import Field, model_validator
 
-from bofire.data_models.kernels.kernel import FeatureSpecificKernel
+from bofire.data_models.kernels.kernel import (
+    ARDKernel,
+    FeatureSpecificKernel,
+    LengthscaleKernel,
+)
 from bofire.data_models.priors.api import AnyPrior, AnyPriorConstraint
 from bofire.data_models.priors.constraint import Positive
 
 
 class CategoricalKernel(FeatureSpecificKernel):
+    """Kernel acting on categorical inputs."""
+
     pass
 
 
-class HammingDistanceKernel(CategoricalKernel):
+class HammingDistanceKernel(ARDKernel, LengthscaleKernel, CategoricalKernel):
+    """Kernel measuring similarity as the number of categories two candidates share.
+
+    Treats the categories as unordered and equally distant from one another, which is
+    the right default unless there is structure among them that `IndexKernel` could
+    learn instead.
+    """
+
     type: Literal["HammingDistanceKernel"] = "HammingDistanceKernel"
-    ard: bool = True
-    lengthscale_prior: Optional[AnyPrior] = None
-    lengthscale_constraint: Optional[AnyPriorConstraint] = None
 
 
 class IndexKernel(CategoricalKernel):
@@ -32,22 +42,31 @@ class IndexKernel(CategoricalKernel):
 
     where $B$ is a low-rank matrix, and $\mathbf v$ is a  non-negative vector.
 
-    Attributes:
-        num_categories (int): Number of categorical values, must be ≥ 2.
-        rank (int): Rank of the kernel approximation, must be ≥ 1 and ≤ num_categories.
-                    Lower rank provides more regularization.
-        prior (Optional[AnyPrior]): Prior distribution over kernel hyperparameters.
-        var_constraint (Optional[AnyPriorConstraint]): Constraint on variance parameter,
-                                                         defaults to Positive().
-    Raises:
-        ValueError: If rank > num_categories.
+    Unlike `HammingDistanceKernel`, which holds all categories equally distant, this
+    learns how similar the categories are from the data.
     """
 
     type: Literal["IndexKernel"] = "IndexKernel"
-    num_categories: Annotated[int, Field(ge=2)]
-    rank: Annotated[int, Field(ge=1)] = 1
-    prior: Optional[AnyPrior] = None
-    var_constraint: Optional[AnyPriorConstraint] = Positive()
+    num_categories: Annotated[int, Field(ge=2)] = Field(
+        description="Number of categories the kernel covers.",
+    )
+    rank: Annotated[int, Field(ge=1)] = Field(
+        default=1,
+        description="Rank of the learned similarity matrix, at most `num_categories`. "
+        "A low rank forces the categories onto a few shared dimensions, which "
+        "regularizes the fit; raising it allows a richer relationship at the cost of "
+        "more parameters.",
+    )
+    prior: Optional[AnyPrior] = Field(
+        default=None,
+        description="Prior over the entries of the low-rank matrix. If not provided, "
+        "the surrogate's default is used.",
+    )
+    var_constraint: Optional[AnyPriorConstraint] = Field(
+        default=Positive(),
+        description="Constraint on the per-category variance added on the diagonal, "
+        "which keeps it non-negative.",
+    )
 
     @model_validator(mode="after")
     def validate_rank_vs_categories(self):
@@ -72,34 +91,55 @@ class PositiveIndexKernel(CategoricalKernel):
 
     NOTE: This kernel should only be used when the correlation between different categories
     is expected to be positive.
-
-    Attributes:
-        num_categories (int): Number of categorical values, must be ≥ 2.
-        rank (int): Rank of the kernel approximation, must be ≥ 1 and ≤ num_categories.
-                    Lower rank provides more regularization.
-        prior (Optional[AnyPrior]): Prior distribution over kernel hyperparameters.
-        var_constraint (Optional[AnyPriorConstraint]): Constraint on variance parameter,
-                                                         defaults to Positive().
-        task_prior (Optional[AnyPrior]): Prior distribution over task-specific parameters.
-        diag_prior (Optional[AnyPrior]): Prior distribution over diagonal elements.
-        normalize_covar_matrix (bool): Whether to normalize the covariance matrix.
-        target_task_index (int): Index of the target task for normalization to value 1,
-                                default to 0 (first category).
-        unit_scale_for_target (bool): Whether to use unit scale for the target task.
-    Raises:
-        ValueError: If rank > num_categories.
     """
 
     type: Literal["PositiveIndexKernel"] = "PositiveIndexKernel"
-    num_categories: Annotated[int, Field(ge=2)]
-    rank: Annotated[int, Field(ge=1)] = 1
-    prior: Optional[AnyPrior] = None
-    var_constraint: Optional[AnyPriorConstraint] = Positive()
-    task_prior: Optional[AnyPrior] = None
-    diag_prior: Optional[AnyPrior] = None
-    normalize_covar_matrix: bool = False
-    target_task_index: Annotated[int, Field(ge=0)] = 0
-    unit_scale_for_target: bool = True
+    num_categories: Annotated[int, Field(ge=2)] = Field(
+        description="Number of categories the kernel covers.",
+    )
+    rank: Annotated[int, Field(ge=1)] = Field(
+        default=1,
+        description="Rank of the learned similarity matrix, at most `num_categories`. "
+        "A low rank forces the categories onto a few shared dimensions, which "
+        "regularizes the fit; raising it allows a richer relationship at the cost of "
+        "more parameters.",
+    )
+    prior: Optional[AnyPrior] = Field(
+        default=None,
+        description="Prior over the entries of the triangular factor. If not provided, "
+        "the surrogate's default is used.",
+    )
+    var_constraint: Optional[AnyPriorConstraint] = Field(
+        default=Positive(),
+        description="Constraint on the per-category variance added on the diagonal, "
+        "which keeps it non-negative.",
+    )
+    task_prior: Optional[AnyPrior] = Field(
+        default=None,
+        description="Prior over the off-diagonal entries, which govern how strongly "
+        "the categories are correlated with one another.",
+    )
+    diag_prior: Optional[AnyPrior] = Field(
+        default=None,
+        description="Prior over the diagonal entries, which govern how much variance "
+        "each category has of its own.",
+    )
+    normalize_covar_matrix: bool = Field(
+        default=False,
+        description="Whether to rescale the whole matrix so that the correlations are "
+        "read on a common scale.",
+    )
+    target_task_index: Annotated[int, Field(ge=0)] = Field(
+        default=0,
+        description="Index into the categories of the one treated as the target, whose "
+        "diagonal entry is normalized to 1 so the others are expressed relative to it. "
+        "Must be below `num_categories` - 1.",
+    )
+    unit_scale_for_target: bool = Field(
+        default=True,
+        description="Whether the target category is held at unit variance rather than "
+        "having its own fitted.",
+    )
 
     @model_validator(mode="after")
     def validate_rank_vs_categories(self):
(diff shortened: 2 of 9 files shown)


bofire/data_models/kernels/aggregation.py
```diff
@@ -23,6 +25,12 @@
 
 
 class AdditiveKernel(AggregationKernel):
+    """Sum of several kernels.
+
+    Two candidates count as similar if they are similar under any one of the summed
```
jduerholt: What are candidates in context of kernels. I have no clue why this is here and what it should serve ...

bofire/data_models/kernels/aggregation.py
```diff
@@ -42,11 +50,16 @@ class AdditiveKernel(AggregationKernel):
...
         ]
-    ]
-    type: Literal["AdditiveKernel"] = "AdditiveKernel"
+    ] = Field(description="The kernels to sum.")
 
 
 class MultiplicativeKernel(AggregationKernel):
+    """Product of several kernels.
+
+    Two candidates count as similar only if they are similar under every one of the
```
jduerholt: See comment above on AdditiveKernel

bofire/data_models/kernels/aggregation.py
```diff
@@ -66,10 +79,16 @@ class MultiplicativeKernel(AggregationKernel):
...
             "ScaleKernel",
         ]
-    ]
+    ] = Field(description="The kernels to multiply.")
 
 
 class ScaleKernel(AggregationKernel):
+    """Wraps another kernel with a fitted output scale.
+
+    The base kernel decides which candidates are similar; this decides how much the
```
jduerholt: All this candidates syntax ready crazy ..

bofire/data_models/kernels/aggregation.py
```diff
@@ -153,10 +175,24 @@ class PolynomialFeatureInteractionKernel(AggregationKernel):
...
+        "grows quickly with this.",
+    )
+    include_self_interactions: bool = Field(
+        description="Whether a kernel may interact with itself, which adds the "
+        "quadratic and higher powers of each kernel alongside the cross terms.",
+    )
+    outputscale_prior: Optional[AnyPrior] = Field(
+        default=None,
+        description="Prior over the output scale applied to each interaction term "
+        "before the terms are summed. If not provided, the surrogate's default is used.",
```
jduerholt: What is the surrogates default?

bofire/data_models/kernels/categorical.py
```diff
@@ -32,22 +42,31 @@ class IndexKernel(CategoricalKernel):
...
+    Unlike `HammingDistanceKernel`, which holds all categories equally distant, this
+    learns how similar the categories are from the data.
     """
 
     type: Literal["IndexKernel"] = "IndexKernel"
-    num_categories: Annotated[int, Field(ge=2)]
-    rank: Annotated[int, Field(ge=1)] = 1
-    prior: Optional[AnyPrior] = None
-    var_constraint: Optional[AnyPriorConstraint] = Positive()
+    num_categories: Annotated[int, Field(ge=2)] = Field(
```
jduerholt, line 55: Is this needed? Can it not just reads it from the features?

bofire/data_models/kernels/categorical.py
```diff
@@ -32,22 +42,31 @@ class IndexKernel(CategoricalKernel):
...
+        default=1,
+        description="Rank of the learned similarity matrix, at most `num_categories`. "
+        "A low rank forces the categories onto a few shared dimensions, which "
+        "regularizes the fit; raising it allows a richer relationship at the cost of "
+        "more parameters.",
+    )
+    prior: Optional[AnyPrior] = Field(
+        default=None,
+        description="Prior over the entries of the low-rank matrix. If not provided, "
+        "the surrogate's default is used.",
```
jduerholt: What is surrogates default?

bofire/data_models/kernels/categorical.py
```diff
@@ -32,22 +42,31 @@ class IndexKernel(CategoricalKernel):
...
+    )
+    var_constraint: Optional[AnyPriorConstraint] = Field(
```
jduerholt, line 70: Would be a prior constraint something differnt? Or how does it relate?

bofire/data_models/kernels/categorical.py
jduerholt: Many a times? Sounds wrong

bofire/data_models/kernels/continuous.py
```diff
@@ -32,41 +50,74 @@ def validate_nu(cls, nu):
...
+    Has no lengthscale, so unlike the RBF and Matern kernels it extrapolates rather than
+    reverting to the mean away from the data.
+    """
+
     type: Literal["LinearKernel"] = "LinearKernel"
-    variance_prior: Optional[AnyPrior] = None
+    variance_prior: Optional[AnyPrior] = Field(
+        default=None,
+        description="Prior over the slope's variance, which sets how steep the fitted "
+        "linear response may be. If not provided, the surrogate's default is used.",
```
jduerholt: What do you mean by surrogates default?

bofire/data_models/kernels/continuous.py
```diff
@@ -32,41 +50,74 @@ def validate_nu(cls, nu):
...
+
+    Has no lengthscale, so unlike the RBF and Matern kernels it extrapolates rather than
+    reverting to the mean away from the data.
+    """
+
     type: Literal["LinearKernel"] = "LinearKernel"
-    variance_prior: Optional[AnyPrior] = None
+    variance_prior: Optional[AnyPrior] = Field(
+        default=None,
+        description="Prior over the slope's variance, which sets how steep the fitted "
```
jduerholt: And is this the correct interpretation of the variance prior? Is it not more an outpuscale prior? Not sure here ...

bofire/data_models/kernels/continuous.py
```diff
@@ -32,41 +50,74 @@ def validate_nu(cls, nu):
...
+        "linear response may be. If not provided, the surrogate's default is used.",
+    )
 
 
 class PolynomialKernel(ContinuousKernel):
+    """Polynomial kernel, fitting a response of a fixed degree in the inputs."""
```
jduerholt: Maybe it also makes sense to show formulas at least where it is easy todo, we can orient orieselfs on gpytorch fort this.

bofire/data_models/kernels/kernel.py
```diff
@@ -1,16 +1,65 @@
...
 
+from pydantic import Field
+
 from bofire.data_models.base import BaseModel
+from bofire.data_models.priors.api import AnyPrior, AnyPriorConstraint
 from bofire.data_models.types import NonRestrictedFeatureKeys
 
 
 class Kernel(BaseModel):
+    """Defines how similar the surrogate expects two candidates to behave.
```
jduerholt: Why always two candidates, this reads not very scienitifc.

bofire/data_models/kernels/kernel.py
```diff
@@ -1,16 +1,65 @@
...
     pass
 
 
 class FeatureSpecificKernel(Kernel):
-    features: Optional[NonRestrictedFeatureKeys] = None
+    """Kernel that can be restricted to a subset of the inputs."""
+
+    features: Optional[NonRestrictedFeatureKeys] = Field(
+        default=None,
+        description="Keys of the features this kernel acts on. Naming an engineered "
```
jduerholt: Totally unclear what "naming an engineered feature" means ...

bofire/data_models/kernels/shape.py
```diff
@@ -15,66 +16,75 @@ class WassersteinKernel(FeatureSpecificKernel):
...
-    append_y: List[float] = []
-    normalize_y: float = 1.0
-    normalize_x: bool = True
-    order: Literal[1, 2] = 1
+    squared: bool = Field(
+        default=False,
+        description="Whether to use the squared exponential form rather than the "
+        "absolute exponential one.",
+    )
+    idx_x: List[int] = Field(
```
jduerholt, line 48: @Jimbo994: why is this of type int, why are we not using feature names here, which would be the usual way in BoFire ...

bofire/data_models/kernels/shape.py
```diff
@@ -15,66 +16,75 @@ class WassersteinKernel(FeatureSpecificKernel):
...
+        default=1.0,
+        description="Divisor applied to the y-coordinates before the distance is "
+        "computed, for bringing curves onto a comparable scale.",
+    )
+    normalize_x: bool = Field(
+        default=True,
+        description="Whether to rescale each curve's x-coordinates by its own maximum, "
+        "so that only the shape of the curve is compared and not its extent.",
+    )
+    order: Literal[1, 2] = Field(
```
jduerholt, line 86: Is this not somehwo the same as the `squared` argument from above?

tests/bofire/data_models/test_documentation.py
```diff
@@ -535,3 +448,29 @@ def test_class_allowlist_has_no_stale_entries():
         f"{sorted(stale)} are documented (or no longer exist) but still listed in "
         "UNDOCUMENTED_CLASSES. Delete them: the allowlist may only shrink."
     )
+
+
+def test_deprecation_pointers_name_a_real_replacement():
```
jduerholt: Please remove this test.

jduerholt (commented):
I reviewed it, a few general ponts:
- the candidate similarity slang reads strange
- maybe we can add some formulas as gpytorch and maybe orient ourselfs a bit in this direction.

jduerholt:
Pushed as 2ad5d3a4. Both general points taken, and two of your questions turned up factual errors.

## The "candidate" framing is gone

You were right that it read strangely, and it was wrong on its own terms: in BoFire a *candidate* is what `ask()` returns, whereas a kernel is a covariance function evaluated at two points of the input space. Borrowing the word for kernel arguments was a category error, not just awkward phrasing. Every docstring now uses proper terms, and `Kernel` opens by saying what it is:

> A kernel $k(\mathbf x, \mathbf x')$ gives the prior covariance between the function values at two points of the input space. It encodes the assumptions the surrogate makes about the response: how smooth it is, over what distance observations carry information, and which inputs matter at all.

## Formulas

Added wherever GPyTorch states one — RBF, Matern, Linear, Polynomial, Hamming, Tanimoto, Downsampling, and the additive/multiplicative/scale aggregations. `IndexKernel` and `PositiveIndexKernel` already had theirs. Using `$$` to match those two, which renders through the existing mathjax setup.

The descriptions now reference the symbols, so `offset_prior` on `PolynomialKernel` reads "Prior over the offset $c$. Expanding the bracket shows that $c$ weights the lower-order terms against the highest one, so a large offset makes the kernel behave more like a linear one."

## "What is the surrogate's default?" — there isn't one

Your asking three times was the tell. **The claim was false**, in six places. `bofire/kernels/mapper.py:92-96` passes `None` straight through to GPyTorch:

```python
lengthscale_prior=(
    priors.map(data_model.lengthscale_prior, d=len(active_dims))
    if data_model.lengthscale_prior is not None
    else None
),
```

So no prior is applied at all and the parameter is fitted from the data alone. That is what it says now. (A surrogate's *default kernel* does carry priors — but that is the surrogate choosing a kernel, not a default filled in for a kernel you constructed yourself.)

## `LinearKernel.variance_prior` — you were right to doubt it

It maps to GPyTorch's `variance_prior` in $k(\mathbf x, \mathbf x') = v\,\mathbf x^\top \mathbf x'$, so $v$ scales the entire kernel and sets the magnitude of the linear response. It is much closer to an output scale than to anything about steepness, which is what I had written.

## `prior` versus `var_constraint`

Now stated wherever both appear: a prior shifts the fitted value, a constraint is the interval the parameter is reparametrized into, so values outside it cannot be reached at all. The default `Positive()` on `var_constraint` keeps the diagonal variances positive.

Worth noting the type name does not help here: `AnyPriorConstraint` constrains the *parameter*, not the prior. Adding it to the rename candidates.

## `num_categories` — cannot currently be inferred

Good question, and the answer is no, not without a signature change. The kernel mapper receives only `active_dims`, the column indices, via `features_to_idx_mapper` (`bofire/kernels/mapper.py:66-77`); it never sees the `Inputs` object, so it has nothing to count categories from. Worth its own issue if you want it removed.

## Also

- "Many a times" → "often"
- the `features` description no longer says "naming an engineered feature"; it explains that an engineered feature is referred to by its own key and contributes every dimension it expands to
- the `log_variant` guard test is removed

Skipped the two `ExactWassersteinKernel` comments, as those are for @Jimbo994.

## Verification

- `pytest tests/bofire/data_models` — 1888 passed, 4 skipped
- `pytest tests/bofire -k "kernel or surrogate"` — 647 passed, 2 skipped
- ruff 0.6.5 clean; `ty==0.0.16` at 158, matching the merged-main baseline

bofire/data_models/kernels/aggregation.py
```diff
@@ -66,10 +80,18 @@ class MultiplicativeKernel(AggregationKernel):
...
-    ]
+    ] = Field(description="The kernels to multiply.")
 
 
 class ScaleKernel(AggregationKernel):
+    r"""Wraps another kernel with a fitted output scale,
+    $k(\mathbf x, \mathbf x') = \theta\, k_{\text{base}}(\mathbf x, \mathbf x')$.
+
+    The base kernel sets the shape of the covariance and this sets its magnitude, which
+    is the variance of the response. Most kernels are wrapped in one before use, since
```
jduerholt: Remove the last sentence. It is a bit verbose.

bofire/data_models/kernels/aggregation.py
```diff
@@ -87,11 +109,21 @@ class ScaleKernel(AggregationKernel):
...
-    outputscale_constraint: Optional[AnyPriorConstraint] = None
+    outputscale_prior: Optional[AnyPrior] = Field(
+        default=None,
+        description="Prior over the output scale $\\theta$, which sets the variance of "
+        "the response. If not provided, no prior is placed on it and it is fitted from "
+        "the data alone.",
+    )
+    outputscale_constraint: Optional[AnyPriorConstraint] = Field(
+        default=None,
+        description="Bounds the output scale is reparametrized into, making values "
```
jduerholt: This reads a bit cumbersome.

bofire/data_models/kernels/aggregation.py
```diff
@@ -153,10 +180,25 @@ class PolynomialFeatureInteractionKernel(AggregationKernel):
             WassersteinKernel,
             ExactWassersteinKernel,
         ]
-    ]
-    max_degree: int
-    include_self_interactions: bool
-    outputscale_prior: Optional[AnyPrior] = None
+    ] = Field(
+        description="The kernels whose interactions are computed. Typically one per "
```
jduerholt: Second sentence is not needed.

bofire/data_models/kernels/aggregation.py
```diff
@@ -153,10 +180,25 @@ class PolynomialFeatureInteractionKernel(AggregationKernel):
...
+        "kind of input, so that the interactions run across the kinds.",
+    )
+    max_degree: int = Field(
+        description="Highest interaction order computed. 1 keeps the kernels "
+        "independent; 2 adds every pairwise interaction, and so on. The number of terms "
```
jduerholt: Last sentence not needed, too verbose.

bofire/data_models/kernels/categorical.py
```diff
@@ -2,20 +2,35 @@
...
 
-class HammingDistanceKernel(CategoricalKernel):
+class HammingDistanceKernel(ARDKernel, LengthscaleKernel, CategoricalKernel):
+    r"""Kernel over categorical inputs based on the Hamming distance.
+
+    $$
+    k(\mathbf x, \mathbf x') = \exp\left(-\frac{d_H(\mathbf x, \mathbf x')}{\ell}\right)
+    $$
+
+    where $d_H$ counts the positions at which the two inputs hold different categories.
```
jduerholt: Orient on the docstirng of botorch for it: https://archive.botorch.org/api/_modules/botorch/models/kernels/categorical.html Also mention that via ard, their can be a lengthscale for every categorical feature.

bofire/data_models/kernels/categorical.py
```diff
@@ -32,22 +47,32 @@ class IndexKernel(CategoricalKernel):
...
+    )
+    prior: Optional[AnyPrior] = Field(
+        default=None,
+        description="Prior over the entries of $B$. If not provided, no prior is "
+        "placed on them and they are fitted from the data alone.",
+    )
+    var_constraint: Optional[AnyPriorConstraint] = Field(
+        default=Positive(),
+        description="Bounds the diagonal entries $\\mathbf v$ are reparametrized into. "
+        "Unlike `prior`, which only shifts the fitted value, this makes values outside "
```
jduerholt: Reomve the sentence with the prior relation.

bofire/data_models/kernels/categorical.py
```diff
@@ -72,34 +97,56 @@ class PositiveIndexKernel(CategoricalKernel):
...
+    )
+    prior: Optional[AnyPrior] = Field(
+        default=None,
+        description="Prior over the entries of $L$. If not provided, no prior is "
+        "placed on them and they are fitted from the data alone.",
+    )
+    var_constraint: Optional[AnyPriorConstraint] = Field(
+        default=Positive(),
+        description="Bounds the diagonal entries $\\mathbf v$ are reparametrized into. "
+        "Unlike `prior`, which only shifts the fitted value, this makes values outside "
```
jduerholt: Remove the sentence as stated above.

bofire/data_models/kernels/categorical.py
```diff
@@ -72,34 +97,56 @@ class PositiveIndexKernel(CategoricalKernel):
...
+        "each category has of its own.",
+    )
+    normalize_covar_matrix: bool = Field(
+        default=False,
+        description="Whether to rescale the whole matrix so that the correlations are "
+        "read on a common scale.",
+    )
+    target_task_index: Annotated[int, Field(ge=0)] = Field(
+        default=0,
+        description="Index into the categories of the one treated as the target, whose "
```
jduerholt: Very unclear

bofire/data_models/kernels/continuous.py
```diff
@@ -32,41 +63,85 @@ def validate_nu(cls, nu):
...
-            are learned for each input dimension. Defaults to True.
-        lengthscale_prior: Optional prior distribution for the lengthscale parameter(s).
-        lengthscale_constraint: Optional constraint on the lengthscale parameter(s).
-        bounds: Bounds for the input features. Can be a single tuple for all dimensions
-            or a list of tuples for per-dimension bounds. Defaults to (0.0, 1.0).
-    Raises:
-        ValueError: If ard is False and bounds is not a list with length equal to input dimension.
+
+    This kernel projects the inputs onto a unit sphere and computes the linear kernel in
+    this space, so it responds to the direction of an input vector rather than its
```
jduerholt, line 132: Remove the last interpreations.

jduerholt (commented):
Ok, somme comments regaring wording and verbosity.

jduerholt:
Pushed as 37145a7f. All nine applied; the BoTorch reference caught a mistake in my formula.

## `HammingDistanceKernel` — the formula was wrong

Following your link: BoTorch's `CategoricalKernel` computes `exp(-dist(x1, x2) / lengthscale)` where `dist` is 0 or 1 per dimension **and the mean is taken across dimensions**. I had written $d_H$ as a Hamming *count*, i.e. a sum. It is a mean. Corrected, and the docstring now also carries the two things you asked for and one more from the same source:

> where $d$ is zero where two inputs hold the same category and one where they differ, averaged over the categorical dimensions. With `ard` there is one lengthscale per categorical feature, so a change of category can matter more in some features than in others. The kernel is not differentiable with respect to its inputs.

## `target_task_index` — "very unclear", agreed

It described the validation bound but never the purpose. Now:

> Position of the target category, the one the covariance matrix is normalized against: $(LL^\top)_{t,t}$ becomes the denominator, so the target has unit variance and every other category is expressed relative to it.

## The prior/constraint contrast

You asked to remove it from both `var_constraint` fields, and called the `outputscale_constraint` one cumbersome. The underlying problem was that I had written it out on **five** separate constraint fields. It now appears **once**, on the `LengthscaleKernel` mixin that seven kernels inherit, and the other four simply say the bounds restrict the parameter during fitting — e.g. *"Bounds the diagonal entries $\mathbf v$ are restricted to during fitting. The default keeps them positive."*

That keeps an answer to your earlier "how does it relate?" question available without repeating it. Say the word if you would rather it went entirely.

## Trimmed

- `ScaleKernel` — closing sentence gone
- `PolynomialFeatureInteractionKernel.kernels` — second sentence gone
- `max_degree` — last sentence gone
- `SphericalLinearKernel` — closing interpretation gone

## One thing worth a separate look

`PositiveIndexKernel.validate_target_task_index` rejects `target_task_index >= num_categories - 1`, so with three categories only 0 and 1 are accepted and the **last category can never be the target**. Given the field selects which category is normalized to unit variance, excluding the last one looks like an off-by-one rather than an intent. I have not touched it, and the description deliberately does not restate the bound.

## Verification

- `pytest tests/bofire/data_models` — 1888 passed, 4 skipped
- `pytest tests/bofire -k "kernel or surrogate"` — 647 passed, 2 skipped
- ruff 0.6.5 clean; `ty==0.0.16` at 158, matching the merged-main baseline

bofire/data_models/kernels/categorical.py
```diff
@@ -72,34 +96,55 @@ class PositiveIndexKernel(CategoricalKernel):
...
-    num_categories: Annotated[int, Field(ge=2)]
-    rank: Annotated[int, Field(ge=1)] = 1
-    prior: Optional[AnyPrior] = None
-    var_constraint: Optional[AnyPriorConstraint] = Positive()
-    task_prior: Optional[AnyPrior] = None
-    diag_prior: Optional[AnyPrior] = None
-    normalize_covar_matrix: bool = False
-    target_task_index: Annotated[int, Field(ge=0)] = 0
-    unit_scale_for_target: bool = True
+    num_categories: Annotated[int, Field(ge=2)] = Field(
```
jduerholt, line 100: Add some point this should be changed so that it can read from the data model directly.

bofire/data_models/kernels/continuous.py
```diff
@@ -1,28 +1,59 @@
...
+    the response may be. Reach for this when an RBF fit looks implausibly smooth
+    between observations.
+    """
+
     type: Literal["MaternKernel"] = "MaternKernel"
-    ard: bool = True
-    nu: float = 2.5
-    lengthscale_prior: Optional[AnyPrior] = None
-    lengthscale_constraint: Optional[AnyPriorConstraint] = None
+    nu: float = Field(
```
jduerholt: We should make `nu` a `Literal` and throw away the validator. It is then obsolete.

bofire/data_models/kernels/continuous.py
```diff
@@ -32,41 +63,84 @@ def validate_nu(cls, nu):
 
 
 class LinearKernel(ContinuousKernel):
+    r"""Linear kernel, equivalent to Bayesian linear regression on the inputs.
+
+    $$
+    k(\mathbf x, \mathbf x') = v\,\mathbf x^{\top} \mathbf x'
+    $$
+
+    Having no lengthscale, it does not revert to the prior mean away from the data, so
```
jduerholt: Remove this sentence.

bofire/data_models/kernels/continuous.py
```diff
@@ -32,41 +63,84 @@ def validate_nu(cls, nu):
...
     """
 
     type: Literal["SphericalLinearKernel"] = "SphericalLinearKernel"
-    ard: bool = True
-    lengthscale_prior: Optional[AnyPrior] = None
-    lengthscale_constraint: Optional[AnyPriorConstraint] = None
-    bounds: Union[tuple[float, float], List[tuple[float, float]]] = (0.0, 1.0)
+    bounds: Union[tuple[float, float], List[tuple[float, float]]] = Field(
+        default=(0.0, 1.0),
+        description="Range the inputs are rescaled from before being projected onto "
```
jduerholt: This should also be resolved, or? So that it can be read from the data, and I think that also a dictionariy is favorable over a list, or?

bofire/data_models/kernels/molecular.py
```diff
@@ -1,12 +1,25 @@
...
 
 
 class MolecularKernel(FeatureSpecificKernel):
+    """Kernel acting on molecular descriptors."""
+
     pass
 
 
-class TanimotoKernel(MolecularKernel):
+class TanimotoKernel(ARDKernel, MolecularKernel):
```
jduerholt, line 12: Why is this not also a lenghtscale kernel? It should be possible to apply it it to different molecular inputs and then also have lenghtscales, or?

bofire/data_models/kernels/continuous.py
```diff
@@ -32,41 +63,84 @@ def validate_nu(cls, nu):
...
 
 class InfiniteWidthBNNKernel(ContinuousKernel):
-    features: Optional[List[str]] = None
+    """Kernel equivalent to a Bayesian neural network of infinite width.
+
+    Captures the kind of hierarchical, non-stationary structure a deep network would,
+    rather than assuming one lengthscale applies across the whole space.
+    """
+
+    features: Optional[List[str]] = Field(
```
jduerholt: This should not be redefined, remove it.

jduerholt (commented):
I provided additonal comments.

jduerholt:
Pushed as 837f2faa. Fixed the documentation and the two cheap structural things here; everything else is now an issue so this can merge.

## Fixed in this PR

**`nu` is a `Literal`** (`continuous.py:50`). `nu: Literal[0.5, 1.5, 2.5]` and `validate_nu` is gone — `CLAUDE.md` already asks for this, since a validator that only restates what the annotation could express is unreachable prose. The accepted values are unchanged, and the schema improves from a bare number to an enum, which is what an agent needs to pick a value:

```
before: {'default': 2.5, 'title': 'Nu', 'type': 'number'}
after:  {'default': 2.5, 'enum': [0.5, 1.5, 2.5], 'type': 'number'}
```

**`InfiniteWidthBNNKernel.features`** is validated like every other kernel's. It redeclared the field as plain `Optional[List[str]]` where `FeatureSpecificKernel` declares `Optional[NonRestrictedFeatureKeys]`, silently dropping the uniqueness check. That dates to #476, the commit that introduced the base class — every other kernel took the shared type, so it reads as an oversight. Dropping the redeclaration restores both the check and the inherited description. One line under **Fixed** in the CHANGELOG, since a kernel naming the same feature twice is now rejected.

**`LinearKernel`** loses the sentence about extrapolating, as requested.

## Filed as issues

- **#812 · bug ·** Shape kernels: lengthscale handling. Three linked problems — a dimensionality-scaled prior shifts 4x between a 2-node and a 32-node curve while the distance is *exactly* node-invariant; `lengthscale.mean()` silently hides redundant lengthscales, which is what kept the first problem invisible; and `WassersteinKernel` has no `ard` where per-dimension lengthscales would be meaningful.
- **#813 · bug ·** `TanimotoKernel.ard` is inert — `has_lengthscale = False`, so `ard_num_dims` shapes nothing. Answers your question about the lengthscale: the kernel returns the Tanimoto similarity directly as the covariance, so there is no distance to divide and adding one changes its functional form.
- **#814 · bug ·** `PositiveIndexKernel.target_task_index` rejects `>= num_categories - 1`, so the last category can never be the target. Looks like an off-by-one.
- **#815 · enhancement ·** Kernels should read configuration from the domain: the index kernels' `num_categories`/`rank`, and `SphericalLinearKernel.bounds` — which should also be a dict keyed by feature rather than a positional list.
- **#816 · enhancement ·** `ExactWassersteinKernel` should address its curve by feature key.

The `x_keys`/`y_keys` refactor is dropped from this branch and recorded as #816 instead, so this PR stays documentation plus the two tidy-ups.

## Verification

- `pytest tests/bofire/data_models` — 1888 passed, 4 skipped
- `pytest tests/bofire -k "kernel or surrogate"` — 647 passed, 2 skipped
- ruff 0.6.5 clean; `ty==0.0.16` at 158, matching the merged-main baseline

jduerholt:
@TobyBoyne: Can you review this PR, for now it focuses just on documentation, I find a few other issues for which I created issues to fix them later.

bofire/data_models/kernels/aggregation.py
```diff
@@ -23,6 +25,12 @@
 
 
 class AdditiveKernel(AggregationKernel):
+    r"""Sum of several kernels, $k(\mathbf x, \mathbf x') = \sum_i k_i(\mathbf x, \mathbf x')$.
+
+    A sum models a response made of independent additive contributions, and the
+    covariance stays high if any one of the summed kernels is high.
```
TobyBoyne: (Tiny nit) This is fine, but maybe a tiny bit misleading. If I have 10 kernels, and only one of them is high, then the covariance of the summed kernels is high. But this is really pedantic of me so you can ignore!

bofire/data_models/kernels/aggregation.py
```diff
@@ -66,10 +80,17 @@ class MultiplicativeKernel(AggregationKernel):
             "MultiplicativeKernel",
             "ScaleKernel",
         ]
-    ]
+    ] = Field(description="The kernels to multiply.")
 
 
 class ScaleKernel(AggregationKernel):
+    r"""Wraps another kernel with a fitted output scale,
+    $k(\mathbf x, \mathbf x') = \theta\, k_{\text{base}}(\mathbf x, \mathbf x')$.
```
TobyBoyne, line 79: Just to confirm - does the latex math render in the docs? I can't find any examples of it, If not, things like `\text` only make this harder to read. But I imagine it does (or at least should)?

bofire/data_models/kernels/aggregation.py
```diff
@@ -66,10 +80,17 @@ class MultiplicativeKernel(AggregationKernel):
...
+
+    The base kernel sets the shape of the covariance and this sets its magnitude, which
+    is the variance of the response.
```
TobyBoyne: Hm maybe this should specify that it is the variance of the underlying, noiseless signal. An objective function that is noisy but flat will have a low kernel variance but the "variance of the response" will be high.

bofire/data_models/kernels/categorical.py
```diff
@@ -58,7 +82,7 @@ def validate_rank_vs_categories(self):
 
 class PositiveIndexKernel(CategoricalKernel):
```
TobyBoyne, line 81: I would expect this to inherit from `IndexKernel`, or at least have more in common with it. For example, why does this have `normalize_cover_matrix`, but `IndexKernel` doesn't?

bofire/data_models/kernels/categorical.py
```diff
@@ -32,22 +47,31 @@ class IndexKernel(CategoricalKernel):
...
+    prior: Optional[AnyPrior] = Field(
+        default=None,
+        description="Prior over the entries of $B$. If not provided, no prior is "
+        "placed on them and they are fitted from the data alone.",
+    )
+    var_constraint: Optional[AnyPriorConstraint] = Field(
+        default=Positive(),
+        description="Bounds the diagonal entries $\\mathbf v$ are restricted to during "
+        "fitting. The default keeps them positive.",
+    )
```
TobyBoyne, line 72: If the default is to restrict them to be positive, doesn't this do the same thing as `PositiveIndexKernel`?

Personally, I would merge the two kernels together. It might not be clear to a user, and I would add a very explicit field called something like `allow_tasks_to_be_anticorrelated`, with a default of `False`. Otherwise a user would likely just use `IndexKernel` without ever realising that `PositiveIndexKernel` is a thing.

bofire/data_models/kernels/conditional.py
```diff
@@ -50,18 +50,39 @@ class ConditionalEmbeddingKernel(Kernel):
...
+        "settings are ignored, so leave them unset and configure the lengthscale on "
+        "this kernel instead.",
+    )
 
-    conditions: Sequence[tuple[str, str, Condition]]
+    conditions: Sequence[tuple[str, str, Condition]] = Field(
+        description="Which feature is active under which circumstances, as triples of "
+        "the dependent feature, the feature it depends on, and the condition that must "
+        "hold. A feature may depend on itself, which expresses that it is only "
+        "meaningful at certain of its own values.",
```
TobyBoyne: "meaningful at certain of its own values" needs to be reworded. I would say "which expresses that it is only relevant to the model under some conditions, e.g., if it is positive."

bofire/data_models/kernels/conditional.py
```diff
@@ -10,8 +10,8 @@
     RBFKernel,
     SphericalLinearKernel,
 )
-from bofire.data_models.kernels.kernel import Kernel
-from bofire.data_models.priors.api import AnyPrior, AnyPriorConstraint
+from bofire.data_models.kernels.kernel import ARDKernel, Kernel, LengthscaleKernel
+from bofire.data_models.priors.api import AnyPrior
 
 
 class ConditionalEmbeddingKernel(Kernel):
```
TobyBoyne, line 17: Should this kernel have an `Examples` block? I think it might be a little unclear currently. Taken from `conditional_features_bo.qmd` below, you might also want a condition that depends on a different feature.

```python
# the catalyst concentration only has an impact if it is non-zero
# otherwise, it is dropped by the model
conditions = [
    (
        "catalyst_concentration",
        "catalyst_concentration",
        NonZeroCondition(),
    )
]

kernel_data_model = ConditionalEmbeddingKernel(
    base_kernel=RBFKernel(),
    conditions=conditions,
)


```

bofire/data_models/kernels/conditional.py
```diff
@@ -50,18 +50,39 @@ class ConditionalEmbeddingKernel(Kernel):
...
+    angle_prior: Optional[AnyPrior] = Field(
+        default=None,
+        description="Prior over the wedge's opening angle, which sets how far apart "
+        "the embeddings of an active and an inactive point are placed.",
+    )
+    radius_prior: Optional[AnyPrior] = Field(
+        default=None,
+        description="Prior over the wedge's radius, which sets how much the value of "
+        "an active feature contributes relative to the fact of it being active.",
+    )
```
TobyBoyne, line 86: I think these descriptions are not quite right.

`angle_prior`: "which sets the distance between the lower and upper bound of the feature (when active) in the embedded space."

`radius_prior`: "which sets the distance between the inactive point and the active points in the embedded space."

bofire/data_models/kernels/continuous.py
```diff
@@ -1,72 +1,131 @@
...
+    the response may be. Reach for this when an RBF fit looks implausibly smooth
+    between observations.
+    """
+
+    type: Literal["MaternKernel"] = "MaternKernel"
+    nu: Literal[0.5, 1.5, 2.5] = Field(
+        default=2.5,
+        description="Smoothness parameter. 0.5 gives a nowhere-differentiable "
+        "response, 1.5 a once-differentiable one and 2.5 a twice-differentiable one. "
+        "In the limit of large nu the kernel becomes the RBF kernel.",
```
TobyBoyne: True but not really relevant, given `nu` can only take a value of 2.5

bofire/data_models/kernels/continuous.py
```diff
@@ -1,72 +1,131 @@
...
+    r"""Linear kernel, equivalent to Bayesian linear regression on the inputs.
+
+    $$
+    k(\mathbf x, \mathbf x') = v\,\mathbf x^{\top} \mathbf x'
+    $$
+    """
+
     type: Literal["LinearKernel"] = "LinearKernel"
-    variance_prior: Optional[AnyPrior] = None
+    variance_prior: Optional[AnyPrior] = Field(
```
TobyBoyne, line 66: Why does this have a variance prior? Shouldn't it just be wrapped in a scale kernel?

bofire/data_models/kernels/continuous.py
```diff
@@ -1,72 +1,131 @@
...
+        default=None,
+        description="Prior over the offset $c$. Expanding the bracket shows that $c$ "
+        "weights the lower-order terms against the highest one, so a large offset makes "
+        "the kernel behave more like a linear one. If not provided, no prior is placed "
+        "on it and it is fitted from the data alone.",
+    )
+    power: int = Field(
+        default=2,
+        description="Degree $p$ of the polynomial. 2 captures pairwise interactions and "
+        "curvature; higher degrees fit more shapes but extrapolate increasingly badly.",
```
TobyBoyne: I don't like "fit more shapes". Maybe:

> Degree $p$ of the polynomial. $p=2$ captures pairwise interactions, and quadratic curvature; higher degrees fit more complex functions, but lead to worse extrapolation and overfitting.

bofire/data_models/kernels/fidelity.py
```diff
@@ -21,11 +21,38 @@ class FidelityKernel(FeatureSpecificKernel):
...
     type: Literal["DownsamplingKernel"] = "DownsamplingKernel"
-    offset_prior: AnyPrior | None = None
-    offset_constraint: AnyPriorConstraint | None = None
-    power_prior: AnyPrior | None = None
-    power_constraint: AnyPriorConstraint | None = None
+    offset_prior: AnyPrior | None = Field(
+        default=None,
+        description="Prior over the offset $c$, which sets how much a low-fidelity "
+        "observation still says about the target fidelity. If not provided, no prior is "
+        "placed on it and it is fitted from the data alone.",
```
TobyBoyne: The phrase "it is fitted from the data alone" is used for all of the priors here. I have a slight issue with it, since it muddies the two purposes of priors:
1. Provide external knowledge/past observations, which is the typical use of a "Bayesian prior"
2. Provide regularization in the model fit

Wrt (1), then yes it is fair to say that it is fitted from the data alone, without external information. But I don't think priors are used here in that sense at all. For (2), I think the regularization impact is really important (eg the log-dim scaled priors), and the model is still fit from the data.

The sentence isn't exactly *false*, but I think it does not do a good job of explaining what the prior is for. If I was a user unfamiliar with the literature, I would see this, think "I don't have any other information that I want to use to fit my model, I want to fit with my data alone" and get rid of all of my priors.

Maybe I'm being too picky here. Curious what you think! I would sooner get rid of this, and write a docstring under `AnyPrior` explaining the role of priors as regularizers.

(Also, I think it's a bit clunky to have the "If not provided, no prior is placed on it and it is fitted from the data alone" repeated so often, but it's not a big deal).

bofire/data_models/kernels/fidelity.py
```diff
@@ -21,11 +21,38 @@ class FidelityKernel(FeatureSpecificKernel):
...
+    power_prior: AnyPrior | None = Field(
+        default=None,
+        description="Prior over the power $\\delta$, which sets how quickly the "
+        "approximation improves as the fidelity rises. If not provided, no prior is "
+        "placed on it and it is fitted from the data alone.",
+    )
+    power_constraint: AnyPriorConstraint | None = Field(
+        default=None,
+        description="Bounds the power $\\delta$ is restricted to during fitting.",
+    )
```
TobyBoyne, line 53: I think maybe the power should be positive by default?

bofire/data_models/kernels/kernel.py
```diff
@@ -1,16 +1,69 @@
...
+
+class ARDKernel(BaseModel):
+    r"""Mixin for a kernel supporting automatic relevance determination."""
+
+    ard: bool = Field(
+        default=True,
+        description="Whether to fit a separate lengthscale per input dimension rather "
+        "than one shared lengthscale. Separate lengthscales let the model down-weight "
+        "dimensions the response does not depend on, at the cost of one more parameter "
+        "per dimension, so a single lengthscale can fit better when data is scarce.",
```
TobyBoyne: "so a single lengthscale can fit better when data is scarce"

I wouldn't necessarily agree. If the underlying function truly is isotropic, then yes a single lengthscale will fit better when data is scarce. But if it isn't (which it rarely is), then I wouldn't have thought this is true as long as you have well behaved priors.

TobyBoyne (commented):
Looks good! Had a few bits of feedback, interested to hear your thoughts :)

bofire/data_models/kernels/aggregation.py
jduerholt: I will remove the whole sentence. Everybody should now what a sum is.

jduerholt: And also remove the corresponding sentence for the ProductKernel

bofire/data_models/kernels/aggregation.py
jduerholt: Having proper doc rendering for the data models will come later, but there are solutions out which can handle it.

bofire/data_models/kernels/aggregation.py
jduerholt: True, we should add this. Good point.

bofire/data_models/kernels/categorical.py
jduerholt: These two kernels were added by @KislayaRavi (if I remember correct), I will file an issue to resolve this.

bofire/data_models/kernels/categorical.py
jduerholt: Yep, I agree. CC @KislayaRavi. But should be a seperate PR.

bofire/data_models/kernels/conditional.py
jduerholt: Thx, this is why I asked you to review, I am quite far away from this kernel.

bofire/data_models/kernels/continuous.py
jduerholt: Hmm, why? Also 0.5 and 1.5 is allowed ...

TobyBoyne: Sorry, I meant nu can take a maximum value of 2.5. In the limit of large nu, let's say as nu >> 20, then this kernel approaches RBF. But since its maximum value is 2.5, I'm not sure how relevant it is. I suppose it maybe gives some understanding into the relationship between the kernels - feel free to leave it

bofire/data_models/kernels/continuous.py
jduerholt: This is an inconsistency that we inherit from gpytorch ...

bofire/data_models/kernels/continuous.py
jduerholt: Ah ok, now I get it.

bofire/data_models/kernels/continuous.py
jduerholt: Nice addition.

bofire/data_models/kernels/categorical.py
KislayaRavi: Your intuition makes a lot of sense.
My preferred re-design would have shared internal base:
```python
class IndexKernelBase(CategoricalKernel):
    num_categories: Annotated[int, Field(ge=2)]
    rank: Annotated[int, Field(ge=1)] = 1
    var_constraint: Optional[AnyPriorConstraint] = Positive()

    def validate_rank_vs_categories(self):
        if self.rank > self.num_categories:
            raise ValueError("rank must be less than or equal to num_categories")
        return self


class IndexKernel(IndexKernelBase):
    type: Literal["IndexKernel"] = "IndexKernel"
    prior: Optional[AnyPrior] = None


class PositiveIndexKernel(IndexKernelBase):
    type: Literal["PositiveIndexKernel"] = "PositiveIndexKernel"
    task_prior: Optional[AnyPrior] = None
    diag_prior: Optional[AnyPrior] = None
    ...
```
However, it will just add another Base class. Its looks good from software design perspective but do we really need a new abstraction?
Regarding the question on `normalize_covar_matrix` -> This one is because the implementation of BoTorch requires this particular variable.
`IndexKernel` is not present in BoTorch but only in GPyTorch
`PositiveIndexKernel` is used from BoTorch
The variables choices and differences arise from the difference in source.

bofire/data_models/kernels/fidelity.py
jduerholt: It is kind of a meta prior, the actual prior is the covariance function itself which is a prior over functions. Maybe We should get rid of it at all.

bofire/data_models/kernels/fidelity.py
jduerholt: Good catch.

bofire/data_models/kernels/kernel.py
jduerholt: Yep, agree. We should remove it. Ofc. it has some kind of regularatory effect and prevents overfitting, classical bias-variance tradeoff ... But we should be in general less verbose ...

bofire/data_models/kernels/categorical.py
KislayaRavi: I don't think both are same. We should not merge them. The basic structure is same:
```math
$K = BB^{T} + diag(v)$
```
Index Kernel : Elements of B are unconstrained, only v is constrained to be positive. So, the correlation between the data points can be both positive and negative.
Positive Index Kernel: Both elements of B and v are constrained to be positive. So, the correlation is always positive.

jduerholt:
Thanks @TobyBoyne — pushed as cad024a9. Mean description length is down from 135 to 105 characters, and two of your corrections were things I had got wrong rather than merely written clumsily.

## Your corrections, adopted

**`angle_prior` and `radius_prior`** were wrong, not just unclear. They set distances in the embedded space — between the feature's bounds when active, and between the inactive point and the active points — and now say so in your words. These were the two claims I flagged in the PR body as my weakest inferences, so the flag earned its keep.

**`ScaleKernel`** now says the variance of the *noiseless signal*, in both the class docstring and `outputscale_prior`. Your point that a noisy but flat objective has low kernel variance and high response variance is exactly the distinction I was eliding.

**`PolynomialKernel.power`** and the self-dependency wording in `ConditionalEmbeddingKernel.conditions` use your phrasing verbatim.

## The prior phrasing

You were right, and it was worse than clunky. "If not provided, no prior is placed on it and it is fitted from the data alone" appeared in **eight** places, and as you say it conflates a prior as external knowledge with a prior as regularizer — a reader unfamiliar with the literature would take it as an invitation to delete their priors. Removed everywhere.

Your suggestion of a docstring under `AnyPrior` explaining priors as regularizers is a good one and belongs with the priors package rather than here; worth its own issue if you want to raise it.

## Also dropped, per your comments

- **`ard`** — "so a single lengthscale can fit better when data is scarce". Only holds if the response is isotropic, which as you say it rarely is.
- **`nu`** — the sentence about large nu approaching the RBF kernel. `nu` caps at 2.5, so the limit is unreachable.
- **`AdditiveKernel` / `MultiplicativeKernel`** — the sentences explaining a sum and a product, per @jduerholt.

## The `Examples` block

It already existed, but was malformed — `Example:` rather than Google's `Examples:`, `>>>` on continuation lines where doctest wants `...`, and `ConditionalEmebeddingKernel` misspelled. Fixed in place rather than adding a second example, since it already shows both a cross-feature and a self-dependent condition, which was your point.

## Filed rather than fixed here

- **#819 ·** `DownsamplingKernel.power` should be constrained positive by default. Your catch; it changes the fitted model, so it wants confirmation and a CHANGELOG entry rather than riding along in a docs PR.

**Not changed:** `LinearKernel.variance_prior`, since per @jduerholt that inconsistency is inherited from gpytorch; and the LaTeX, since data model doc rendering is coming later. The `IndexKernel`/`PositiveIndexKernel` merge and the `IndexKernelBase` proposal from @KislayaRavi are @jduerholt's issue, and I have trimmed #815 down to `SphericalLinearKernel.bounds` so it does not overlap.

## Verification

- `pytest tests/bofire/data_models` — 1888 passed, 4 skipped
- `pytest tests/bofire -k "kernel or surrogate"` — 647 passed, 2 skipped
- ruff 0.6.5 clean; `ty==0.0.16` at 158, matching the merged-main baseline

jduerholt:
Hi @TobyBoyne, can you have a look again?

bofire/data_models/kernels/categorical.py
TobyBoyne: My suggestion would be to merge them on the BoFire side, and then dispatch them based on a new feature like `allow_tasks_to_be_anticorrelated`. I'm not sure a BoFire user should have to worry about what is implemented in BoTorch vs GPyTorch. But happy with whatever you both settle on.

And in reply to your other comment: I see the difference, thanks for explaining!

TobyBoyne (approved):
All good on my end :)

PR #811, opened 2026-09-11
feature: Added shap feature for categorical ouput

KislayaRavi:
## Motivation

Extend SHAP feature importance to support `CategoricalOutput`. The implementation explains every category probability and supports multiclass and mixed continuous/categorical output predictors while preserving category names when combining explanations.

yes

yes

## Test Plan

Added tests in `tests/bofire/surrogates/test_feature_importance.py` to:
- test shap value calculation for a three-category categorical output and,
- test mixed  continuous/categorical output.


bofire/surrogates/feature_importance.py
```diff
@@ -118,6 +118,23 @@ def predict(X: np.ndarray, output_key: str) -> np.ndarray:
...
+    def predict_categorical(
+        X: np.ndarray, output_key: str, categories: Sequence[str]
+    ) -> np.ndarray:
+        """Predict function for the categorical output of the surrogate."""
+        frame = pd.DataFrame(X, columns=predictor.inputs.get_keys())
+        predictions = predictor.predict(frame)
+        probability_columns = [
+            f"{output_key}_{category}_prob" for category in categories
+        ]
+        missing_columns = set(probability_columns) - set(predictions.columns)
```
jduerholt: The missing_columns tests are from my point of view not necessary, this is an internal function that it is not exposed via any API and both `output_key` and `categories` are populated based on the `CategoricalOutput` feature. So from my perspective, it can be dropped.

tests/bofire/surrogates/test_feature_importance.py
```diff
@@ -73,6 +81,164 @@ def test_shap_importance_for_surrogate():
...
+    combined = combine_shap_importances([importance, importance])["f_1"]
+    assert combined.output_names == outputs[0].categories
+    assert combined.values.shape == (
+        2 * len(candidates),
+        len(inputs),
+        len(outputs[0].categories),
+    )
+
+
+def test_shap_importance_for_continuous_and_categorical_outputs():
```
jduerholt: I think that this test can be dropped as we are currently not really supporting multi output models, it is more an artefact of the shap method that this is written in the specific manner that it allows it.

jduerholt (changes requested):
Hi Ravi,

thank you very much. I only have a few small requests, else it is fine!

Best,

Johannes

tests/bofire/surrogates/test_feature_importance.py
KislayaRavi: I dropped it.

bofire/surrogates/feature_importance.py
KislayaRavi: I agree with you. I removed the `missing_columns` check.

jduerholt (approved):
Thanks, looks good to me!

Issue #820, opened 2026-09-17
Surrogate presets are built three different ways; the mixed GP should be a kernel


jduerholt:
Found while documenting the `surrogates` data models.

## The inconsistency

Four of the eighteen surrogates have no functional class of their own — they are presets over `SingleTaskGPSurrogate`, and each uses a different mechanism:

| surrogate | mechanism |
|---|---|
| `LinearSurrogate` | sibling class, narrows `kernel` to `LinearKernel`, maps to `SingleTaskGPSurrogate`'s functional class |
| `PolynomialSurrogate` | same |
| `SingleTaskIBNNSurrogate` | **subclasses** `SingleTaskGPSurrogate` and narrows `kernel` |
| `MixedSingleTaskGPSurrogate` | data-model-to-data-model translation in `bofire/surrogates/mapper.py:29-71` |

For contrast, `TanimotoGPSurrogate` is *not* a preset — it has real specialisation (`tanimoto_calculation_mode`, the precomputed similarity matrix).

## Proposal

**A preset belongs in the package that owns the thing being preset**, and is a class when it adds *structure*, a function when it only adds *values*. BoFire already uses both halves of that split: `PolynomialFeatureInteractionKernel` is a class standing for a composition, and `THREESIX_LENGTHSCALE_PRIOR = partial(GammaPrior, concentration=3.0, rate=6.0)` is a function standing for a set of values.

The mixed GP's entire content is a kernel composition — `map_MixedSingleTaskGPSurrogate` builds

```python
AdditiveKernel(kernels=[
    ScaleKernel(base_kernel=AdditiveKernel(kernels=[k_cont, ScaleKernel(base_kernel=k_cat)])),
    ScaleKernel(base_kernel=MultiplicativeKernel(kernels=[k_cont, k_cat])),
])
```

so it should be a `MixedKernel` in `kernels`:

```python
class MixedKernel(AggregationKernel):
    type: Literal["MixedKernel"] = "MixedKernel"
    continuous_kernel: AnyContinuousKernel
    categorical_kernel: AnyCategoricalKernel
```

[...]

PR #606, opened 2025-06-27
Additions for Fully Bayesian Models

jduerholt:
## Motivation

This PR adds the following stuff:
- FullyBayesianLinearSingleTaskGP
- FullyBayesianSignleTaskGP
- Input warping for fully bayesian GPs (there is a bug in botorch regarding the warpinng which is addressed here https://github.com/pytorch/botorch/pull/2894 and leads to the failure of a test in this PR, in addition the warping feature is not yet in a release of botorch, so addional tests will fail)
- MAP version of the SAASFullyBayesianSingleTaskGP
- SaasFullyBayesianSingleTaskGPSurrogate is rename to FullyBayesianSignleTaskGPSurrogate and by the `model_type` keyword one can switch between a linear kernel, a matern kernel with saas priors and a matern kernel with Hvafner priors.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/experimental-design/bofire/blob/main/CONTRIBUTING.md#pull-requests)?

Yes.

## Test Plan

Unit tests.

PR #821, opened 2026-09-17
Document the surrogate data models (part 1: base classes and the GP family)

jduerholt:
Sixth package batch of the migration started in #802, covering the base classes and the GP family — including the GP presets. 71 field descriptions and 18 class docstrings, taking the ratchet from **328 to 242** remaining entries.

Files: `surrogate.py`, `botorch.py`, `trainable.py`, `trainable_botorch.py`, `scaler.py`, `single_task_gp.py`, `mixed_single_task_gp.py`, `multi_task_gp.py`, `robust_single_task_gp.py`, `fully_bayesian.py`, `map_saas.py`, `tanimoto_gp.py`, `pairwise_gp.py`, `linear.py`, `polynomial.py`, `bnn.py`.

Part 2 is the non-GP surrogates and the container: `mlp.py`, `random_forest.py`, `deterministic.py`, `empirical.py`, `botorch_surrogates.py`. Splitting on "is it a GP" keeps each diff reviewable and lets part 2 inherit the base-class descriptions settled here.

## No new mixins

The previous packages set a different expectation, so worth saying explicitly: unlike `kernels` and `acquisition_functions` there is little to hoist here.

- `kernel` is declared on eleven classes, but each narrows the type deliberately (`AnyKernel`, `AnyContinuousKernel`, `LinearKernel`, `PolynomialKernel`, `InfiniteWidthBNNKernel`, `Union[ScaleKernel, RBFKernel, MaternKernel]`, …).
- `noise_prior` / `noise_constraint` share a type across nine classes but differ in default, and pydantic cannot override only a default — redeclaring the field drops the inherited `description`.

The repeated prose therefore lives in module constants referenced from each redeclaration: `NOISE_PRIOR_DESCRIPTION`, `NOISE_CONSTRAINT_DESCRIPTION`, `HYPERCONFIG_DESCRIPTION` in `trainable_botorch.py`; `HYPERSTRATEGY_DESCRIPTION`, `HYPERCONFIG_INPUTS_DESCRIPTION`, `TARGET_METRIC_DESCRIPTION` in `trainable.py`.

`SingleTaskIBNNSurrogate` is the clearest case of why this matters: it redeclared `kernel` and `hyperconfig` as bare annotations, silently dropping both inherited descriptions.

## Two small structural changes

**Schema-visible defaults.** `BotorchSurrogate.categorical_encodings`, `engineered_features` and `MultiTaskGPSurrogate.task_prior` used `default_factory`, which pydantic omits from `model_json_schema()` — so a caller reading the schema could not see what they default to. They now use literal defaults, which pydantic deep-copies. Verified that two surrogates built from the defaults do not share the objects.

**Dead code.** `map_saas.TestSurrogate`, an empty class added in #606, is removed. Never referenced, never exported from the api.

## Claims taken from BoTorch rather than read out of this repo

Flagging these explicitly, since that caught real errors in #807 and #809:

- The SAAS prior concentrates the inverse lengthscales so only a few inputs retain influence.
- `AdditiveMapSaasSingleTaskGP` sums Matern-5/2 terms at taus drawn from a HalfCauchy; `EnsembleMapSaasSingleTaskGP` keeps them as a batch and mixes their posteriors, so the spread between them enters the predicted uncertainty.
- `PairwiseGP`'s `probit` and `logit` likelihoods correspond to Gaussian comparison noise and to Bradley-Terry respectively.
- `FullyBayesianSingleTaskGPSurrogate`'s `warmup_steps` / `num_samples` / `thinning` are NUTS parameters, and the fitted model retains `num_samples / thinning` hyperparameter sets.
- `RobustSingleTaskGPSurrogate` fits a per-point noise level (Ament et al. 2024).

Everything else — the tanimoto precompute/on-the-fly trade-off, what `Scaler.features` means when empty, what each hyperconfig searches over — is read out of `bofire/surrogates/` and `bofire/surrogates/utils.py`.

## Design findings filed separately

The investigation that preceded this turned up several inconsistencies. Per the agreed scoping they are filed as #820 rather than acted on here: surrogate presets are built three different ways, the mixed GP would be better expressed as a `MixedKernel`, and `MixedSingleTaskGPSurrogate.validate_categoricals` mutates the caller's kernel in place.

## Verification

- `pytest tests/bofire/data_models` — 1888 passed, 4 skipped
- `pytest tests/bofire -k "surrogate or gp or mlp or forest or linear or polynomial or bnn"` — 549 passed, 3 skipped
- `ruff check` / `ruff format --check` clean
- `ty check bofire` — 158 diagnostics, unchanged from the `main` baseline
- Spot-checked that the base-class descriptions reach `SingleTaskGPSurrogate.model_json_schema()` by inheritance
- Descriptions do not affect `model_dump()`, so the serialization specs need no changes


bofire/data_models/surrogates/fully_bayesian.py
```diff
@@ -8,16 +8,45 @@
...
+        "inputs matter.",
+    )
+    warmup_steps: Annotated[int, Field(ge=1)] = Field(
+        default=256,
+        description="MCMC steps discarded before sampling starts, allowing the chain "
+        "to reach its stationary distribution.",
+    )
+    num_samples: Annotated[int, Field(ge=1)] = Field(
+        default=128,
+        description="MCMC steps kept after warm-up, before thinning.",
```
jduerholt: Is `kept` the correct word, it is in principle the number of steps after warmup that are run.

bofire/data_models/surrogates/map_saas.py
```diff
@@ -1,30 +1,37 @@
...
-    """Additive MAP SAAS single-task GP
-
-    Maximum-a-posteriori (MAP) version of the sparse axis-aligned subspace
-    `FullyBayesianSingleTaskGPSurrogate` with `model_type` equals to "saas".
-
-    Attributes:
-        n_taus (PositiveInt): Number of sub-kernels to use in the SAAS model.
+    """GP whose kernel sums several sparse Matern terms, at different sparsity levels.
+
+    The sparse axis-aligned subspace (SAAS) prior on each term pulls most
```
jduerholt: Shorten it, just say that it is a map version of the original fully bayesian implementation, keep the sentence when to pick.

bofire/data_models/surrogates/map_saas.py
```diff
@@ -1,30 +1,37 @@
...
+        >>> surrogate = AdditiveMapSaasSingleTaskGPSurrogate(
+        ...     inputs=inputs, outputs=outputs, n_taus=8
+        ... )
     """
 
     type: Literal["AdditiveMapSaasSingleTaskGPSurrogate"] = (
         "AdditiveMapSaasSingleTaskGPSurrogate"
     )
-    n_taus: PositiveInt = 4
+    n_taus: PositiveInt = Field(
```
jduerholt: In principle number of sparse matern kernels beeing summed up

bofire/data_models/surrogates/map_saas.py
```diff
@@ -38,20 +45,27 @@ def is_output_implemented(cls, my_type: Type[AnyOutput]) -> bool:
 
 
 class EnsembleMapSaasSingleTaskGPSurrogate(TrainableBotorchSurrogate):
-    """Ensemble MAP SAAS single-task GP
-
-    Batched ensemble of ``SingleTaskGP``s with the Matern-5/2 kernel and a SAAS prior.
+    """Ensemble of sparse GPs, each assuming a different number of inputs matters.
```
jduerholt: Just write that it is as the one above another map apporximation of the fullybayesian saas model using a different mechanism. And that this one is the preferred one.

bofire/data_models/surrogates/map_saas.py
```diff
@@ -38,20 +45,27 @@ def is_output_implemented(cls, my_type: Type[AnyOutput]) -> bool:
...
+    separate models and averaged rather than summed, so the spread between them
+    feeds into the predicted uncertainty.
     """
 
     type: Literal["EnsembleMapSaasSingleTaskGPSurrogate"] = (
         "EnsembleMapSaasSingleTaskGPSurrogate"
     )
-    n_taus: PositiveInt = 4
-    output_scaler: ScalerEnum = ScalerEnum.STANDARDIZE
+    n_taus: PositiveInt = Field(
```
jduerholt: see above

bofire/data_models/surrogates/mixed_single_task_gp.py
```diff
@@ -102,25 +123,43 @@ def _update_hyperparameters(
 
 
 class MixedSingleTaskGPSurrogate(TrainableBotorchSurrogate):
+    """Gaussian process over a mix of continuous and categorical inputs.
+
+    Combines a kernel over the continuous inputs with one over the categoricals, both
```
jduerholt: just add the actual kernel structure, then we need less words.

bofire/data_models/surrogates/mixed_single_task_gp.py
```diff
@@ -102,25 +123,43 @@ def _update_hyperparameters(
...
+
     type: Literal["MixedSingleTaskGPSurrogate"] = "MixedSingleTaskGPSurrogate"
     continuous_kernel: AnyContinuousKernel = Field(
         default_factory=lambda: RBFKernel(
             ard=True,
             lengthscale_prior=HVARFNER_LENGTHSCALE_PRIOR(),
             lengthscale_constraint=GreaterThan(lower_bound=2.500e-02),
-        )
+        ),
+        description="Covariance function over the continuous inputs. Its `features` "
```
jduerholt: Say that they are set from then when they are empty.

bofire/data_models/surrogates/mixed_single_task_gp.py
```diff
@@ -102,25 +123,43 @@ def _update_hyperparameters(
...
+        "are filled in from the inputs that are not ordinal-encoded categoricals, so "
+        "they need not be set.",
     )
     categorical_kernel: AnyCategoricalKernel = Field(
         default_factory=lambda: HammingDistanceKernel(
             ard=True, lengthscale_constraint=GreaterThan(lower_bound=1.000e-06)
         ),
+        description="Covariance function over the categorical inputs. Its `features` "
```
jduerholt: say, that this happens when they are not specified.

bofire/data_models/surrogates/pairwise_gp.py
```diff
@@ -16,18 +16,12 @@
 
 
 class PairwiseGPSurrogate(BotorchSurrogate, TrainableSurrogate):
```
jduerholt: Why not inherit from TrainableBotorchSurrogate? Can one restucure maybe a bit?

bofire/data_models/surrogates/robust_single_task_gp.py
```diff
@@ -43,22 +47,42 @@ class RobustSingleTaskGPSurrogate(TrainableBotorchSurrogate):
...
             lengthscale_constraint=ROBUSTGP_LENGTHSCALE_CONSTRAINT(),
             outputscale_constraint=ROBUSTGP_OUTPUTSCALE_CONSTRAINT(),
         ),
+        description=HYPERCONFIG_DESCRIPTION,
     )
 
-    prior_mean_of_support: Optional[int] = Field(default=None)
-    convex_parametrization: bool = Field(default=True)
-    cache_model_trace: bool = Field(default=False)
+    prior_mean_of_support: Optional[int] = Field(
```
jduerholt: use the descriptoion that also botorch uses here.

bofire/data_models/surrogates/scaler.py
```diff
@@ -18,15 +23,38 @@ class ScalerEnum(str, Enum):
...
+    experiments instead.
+    """
+
     type: Literal["Normalize"] = "Normalize"
 
 
 class Standardize(Scaler):
+    """Rescales each feature to zero mean and unit variance, using the observed data.
+
+    Unlike `Normalize`, which uses the declared bounds, this follows the experiments, so
```
jduerholt: Remove this sentence, it is is obvious

bofire/data_models/surrogates/single_task_gp.py
```diff
@@ -136,20 +165,42 @@ def matern_15(
 
 
 class SingleTaskGPSurrogate(TrainableBotorchSurrogate):
+    """Gaussian process over one output, the usual starting point.
```
jduerholt: remove "usual staring point"

bofire/data_models/surrogates/single_task_gp.py
```diff
@@ -136,20 +165,42 @@ def matern_15(
...
+
+    Predicts a mean and an uncertainty everywhere, which is what lets a Bayesian
```
jduerholt: Remove this sentence

bofire/data_models/surrogates/single_task_gp.py
```diff
@@ -136,20 +165,42 @@ def matern_15(
...
+    optimization strategy trade exploration against exploitation.
+
+    Examples:
```
jduerholt: Remove this example block

bofire/data_models/surrogates/single_task_gp.py
```diff
@@ -136,20 +165,42 @@ def matern_15(
...
+
     type: Literal["SingleTaskGPSurrogate"] = "SingleTaskGPSurrogate"
 
     kernel: AnyKernel = Field(
         default_factory=lambda: RBFKernel(
             ard=True,
             lengthscale_prior=HVARFNER_LENGTHSCALE_PRIOR(),
-        )
+        ),
+        description="Covariance function, encoding what the model assumes about the "
```
jduerholt: make this everywhere more technical in the direction of. "The module computing the covariance (kernel) matrix which encodes the similarity between inputs.

bofire/data_models/surrogates/surrogate.py
```diff
@@ -11,10 +11,26 @@
 
 
 class Surrogate(BaseModel):
+    """Model of how the outputs respond to the inputs, fitted to past experiments.
```
jduerholt: it does not needs to be fitted see determiinsitic models. It just models the relation between inputs and outputs.

bofire/data_models/surrogates/surrogate.py
```diff
@@ -11,10 +11,26 @@
...
+
+    A strategy uses a surrogate to predict what an untried candidate would give, so the
```
jduerholt: Remove this sentence.

bofire/data_models/surrogates/surrogate.py
```diff
@@ -11,10 +11,26 @@
...
+    choice of surrogate is the choice of what the optimizer believes about the response.
+    """
+
     type: Any
-    inputs: Inputs
-    outputs: Outputs
-    dump: Optional[str] = None
+    inputs: Inputs = Field(
+        description="Input features the surrogate is fitted on. Need not be the whole "
```
jduerholt: This reads strange, especially witht he domain, one has to relate it to the strategy. if the surrogate is part of a strategy than not every feature has to be in, it can also act on a subset ...

bofire/data_models/surrogates/tanimoto_gp.py
```diff
@@ -14,10 +14,19 @@
...
 )
-from bofire.data_models.surrogates.trainable_botorch import TrainableBotorchSurrogate
+from bofire.data_models.surrogates.trainable_botorch import (
+    NOISE_CONSTRAINT_DESCRIPTION,
+    NOISE_PRIOR_DESCRIPTION,
+    TrainableBotorchSurrogate,
+)
 
 
 class TanimotoGPSurrogate(TrainableBotorchSurrogate):
```
jduerholt: maybe one should strip down the complete surrogate to just molecular inputs, in the current setup this is what makes sense. Then we also do not let to set a kernel and just set the priors.

bofire/data_models/surrogates/trainable.py
```diff
@@ -27,14 +27,45 @@
...
+    """
+
     type: Any
     hyperstrategy: Literal[
         "RandomStrategy", "FractionalFactorialStrategy", "SoboStrategy"
-    ]
-    inputs: Inputs
-    n_iterations: Optional[Annotated[int, Field(ge=1)]] = None
-    target_metric: RegressionMetricsEnum = RegressionMetricsEnum.MAE
+    ] = Field(description=HYPERSTRATEGY_DESCRIPTION)
```
jduerholt: In a lot of hypercofigs, these things like target metric are defined again?

bofire/data_models/surrogates/trainable.py
```diff
@@ -75,7 +106,17 @@ def _update_hyperparameters(surrogate_data, hyperparameters: pd.Series):
...
-    hyperconfig: Optional[Hyperconfig] = None
+    """Surrogate whose parameters are learned from the experiments.
+
+    Such a surrogate can also have its hyperparameters — the settings that are not
+    learned by the fit itself — searched over before fitting.
+    """
+
+    hyperconfig: Optional[Hyperconfig] = Field(
+        default=None,
+        description="Search over the surrogate's own hyperparameters, run before "
```
jduerholt: run before fitting reads strange, (also above), it is more that these configs can be used to run a hyperopt upfront

bofire/data_models/surrogates/trainable_botorch.py
```diff
@@ -5,9 +5,40 @@
...
 
 
+# reused by the GP surrogates, which must redeclare these fields because their defaults
+# differ and pydantic cannot override a default without redeclaring
+NOISE_PRIOR_DESCRIPTION = (
+    "Prior over the observation noise, which sets how much of the spread in the data "
+    "the model attributes to measurement error rather than to the response."
+)
+NOISE_CONSTRAINT_DESCRIPTION = (
+    "Bounds the observation noise is restricted to during fitting. A positive lower "
```
jduerholt: Reads strange ...

jduerholt: Also remove the last sentence.

bofire/data_models/surrogates/trainable_botorch.py
```diff
@@ -5,9 +5,40 @@
...
+    "bound keeps the fit numerically stable."
+)
+HYPERCONFIG_DESCRIPTION = (
+    "Search over this surrogate's own hyperparameters, run before fitting. Set to null "
```
jduerholt: again the run before fitting is not what it is. Also set to null is not what it is. it can be used for optionally run hyperopts as standalones or within a strategy. but it just defines the confoig.

bofire/data_models/surrogates/trainable_botorch.py
```diff
@@ -5,9 +5,40 @@
...
+    "to use the hyperparameters as configured."
+)
+
+
 class TrainableBotorchSurrogate(BotorchSurrogate, TrainableSurrogate):
-    scaler: AnyScaler = Field(default_factory=Normalize)
-    output_scaler: ScalerEnum = ScalerEnum.STANDARDIZE
+    """BoTorch surrogate fitted to the experiments, with rescaling on both sides.
```
jduerholt:
```suggestion
    """BoTorch based surrogate fitted to the experiments.
```

bofire/data_models/surrogates/trainable_botorch.py
```diff
@@ -5,9 +5,40 @@
...
+
+    Fitting is scale-sensitive, so the inputs and the output are rescaled before the
```
jduerholt:
```suggestion
    Fitting is often scale-sensitive, so the inputs and the output are rescaled by default before the
```

bofire/data_models/surrogates/trainable_botorch.py
```diff
@@ -5,9 +5,40 @@
...
+
+    scaler: AnyScaler = Field(
+        default_factory=Normalize,
+        description="How the inputs are rescaled before fitting. Set to null to leave "
+        "them as they are.",
+    )
+    output_scaler: ScalerEnum = Field(
+        default=ScalerEnum.STANDARDIZE,
+        description="How the outputs are rescaled before fitting, and undone when "
+        "predicting. Standardizing lets one set of priors suit outputs of any "
```
jduerholt: Remove thhe sentence regarding priors

bofire/data_models/surrogates/bnn.py
jduerholt: I would remove it at least as class, remove completely or just have a function with the same name which takes `Inputs` and `Outputs` as arguments and spits out a SingletaskGP with the kernel, but honestly, I do not see too much value in it as it is kind of special and one can very easy construcht it from a singletaskgp.

bofire/data_models/surrogates/linear.py
```diff
@@ -11,16 +11,35 @@
...
 )
-from bofire.data_models.surrogates.trainable_botorch import TrainableBotorchSurrogate
+from bofire.data_models.surrogates.trainable_botorch import (
+    NOISE_CONSTRAINT_DESCRIPTION,
+    NOISE_PRIOR_DESCRIPTION,
+    TrainableBotorchSurrogate,
+)
 
 
 class LinearSurrogate(TrainableBotorchSurrogate):
```
jduerholt: Do we need a class for it? A function preset which just takes the arguments and prouces a singletaskgp should be fine or?

jduerholt: kernely hyperparams like priors would then also be arguments with defaults.

bofire/data_models/surrogates/polynomial.py
```diff
@@ -11,20 +11,54 @@
...
 )
-from bofire.data_models.surrogates.trainable_botorch import TrainableBotorchSurrogate
+from bofire.data_models.surrogates.trainable_botorch import (
+    NOISE_CONSTRAINT_DESCRIPTION,
+    NOISE_PRIOR_DESCRIPTION,
+    TrainableBotorchSurrogate,
+)
 
 
 class PolynomialSurrogate(TrainableBotorchSurrogate):
```
jduerholt: I do not know, from my point of view this can also go into a function preset as linear one.

jduerholt:
All 32 comments addressed. Two commits: `e5a197b4` for the doc-only ones, `19b5be96` for the structural ones.

## Doc comments

Most were direct. The ones where I changed more than the words:

**"run before fitting" was wrong twice over.** You flagged it on `trainable.py:117` and `trainable_botorch.py:19`, and checking `bofire/runners/hyperoptimize.py` you're right on both counts — the hyperconfig is *only* a config, and nothing runs it at fit time. `hyperoptimize()` runs it explicitly, or a strategy does via `frequency_hyperopt`. Reworded on `Hyperconfig`, `TrainableSurrogate`, `HYPERCONFIG_DESCRIPTION` and the IBNN override, all of which said some version of the same wrong thing.

**Kernel wording.** `KERNEL_DESCRIPTION` in `botorch.py` now carries your sentence — "The module computing the covariance matrix, which encodes the similarity between inputs" — and opens every kernel field, with the narrowing clause appended where the type is fixed.

**`target_metric` (your `trainable.py:57` question).** Yes — `SingleTaskGPHyperconfig`, `MixedSingleTaskGPHyperconfig` and `MultiTaskGPHyperconfig` each redeclared it with the base class's own type *and* default, so the redeclaration did nothing but drop the inherited description. All three deleted. `hyperstrategy` genuinely does differ (the subclasses add a default), so it stays.

**`prior_mean_of_support`** now uses BoTorch's own wording: mean of the default exponential prior over the support size.

## Structural comments

Per your steer — IBNN removed, linear/polynomial to function presets, pairwise fixed, Tanimoto to the issue.

**`LinearSurrogate` / `PolynomialSurrogate` → functions**, `SingleTaskIBNNSurrogate` removed. No production code referenced any of them; only tests and the registration plumbing. Call sites are unchanged, which is why the two functions keep their CapWords names. `from_power` becomes a `power=` argument.

One thing worth flagging, because it would have been a silent regression: the classes inherited `hyperconfig = None` from `TrainableSurrogate`, but `SingleTaskGPSurrogate` defaults to a hyperconfig that searches over RBF and Matern — which would have *discarded the preset kernel*. Both functions set `hyperconfig=None`, and there's a test pinning it.

**PairwiseGP.** The blocker was `output_scaler`: `bofire/surrogates/pairwise_gp.py:33` reads only `self.scaler`, and a latent utility has no observed values to rescale, so inheriting `TrainableBotorchSurrogate` would have added a field that is structurally meaningless. Split the base instead — `InputScaledBotorchSurrogate` carries `scaler` and its validator, `TrainableBotorchSurrogate` adds `output_scaler` on top. No other surrogate changes, and PairwiseGP's serialized field set is identical. It does gain the `validate_scaler_features` check it silently skipped before.

**Tanimoto** → recorded on #820 rather than done. You called it complex and I agree, but there's a second reason: it interacts with the `MixedKernel` proposal already in that issue. `MixedKernel(continuous_kernel=RBFKernel(), categorical_kernel=TanimotoKernel())` is the composition a molecular-plus-continuous GP wants, which argues for keeping Tanimoto expressible *as a kernel* rather than hard-wiring it into a surrogate that forbids a kernel field. Worth settling together.

#820 updated with what's now done and what's left.

## Verification

- Full suite: `3217 passed, 66 skipped, 1 failed`. The failure is `tests/bofire/strategies/doe/test_utils.py::test_n_zero_eigvals_{constrained,unconstrained}` — **pre-existing**, verified failing on clean `HEAD` with these changes stashed. Unrelated to surrogates.
- `ruff check` / `format --check` clean; `ty check bofire` 158 diagnostics, unchanged from the `main` baseline.
- Ratchet at 242.

bofire/data_models/surrogates/botorch.py
```diff
@@ -9,25 +9,35 @@
...
 
-    Attributes:
-        categorical_encodings: A dictionary specifying how
-            categorical features are to be encoded **within** the botorch based surrogate.
-            Keys are the feature keys and values are the encoding types. If a feature is
-            not specified, a default is chosen from the descriptor *data* the feature
-            carries: a feature with a structure column (e.g. ``smiles``) is fingerprint
-            encoded, one with numeric descriptor columns is descriptor encoded, and a
-            plain categorical falls back to the surrogate-specific default (one-hot here).
+    Everything the model sees is numeric, so this is the level at which categoricals
```
jduerholt: Remove this sentence.

bofire/data_models/surrogates/fully_bayesian.py
```diff
@@ -8,16 +8,46 @@
 
 
 class FullyBayesianSingleTaskGPSurrogate(TrainableBotorchSurrogate):
+    """Gaussian process whose hyperparameters are sampled rather than point-fitted.
+
+    Averaging predictions over the sampled hyperparameters gives more honest uncertainty
```
jduerholt:
```suggestion
    Averaging predictions over the sampled hyperparameters gives improved uncertainty
```

